PR #6, opened 2022-02-25
fix(grids): Fix grid component to properly update columns when they are changed via the `columns` input

Blackbaud-TrevorBurch:
The grid component was only updating the displayed columns array when columns were updated via template changes. This fix updated the displayed column array and the widths of all columns no matter how the available columns are updated.

### DIFF
--- a/.skyuxdev.json
+++ b/.skyuxdev.json
@@ -1,5 +1,5 @@
 {
-  "allowedSemverRange": ">5.5.0-beta <5.5.0-rc",
+  "allowedSemverRange": "^5.0.0",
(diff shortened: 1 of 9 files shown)


Blackbaud-SteveBrush (dismissed):

Blackbaud-SteveBrush (commented):
See title, "properly updated"?

Blackbaud-TrevorBurch:
@Blackbaud-SteveBrush - good catch. Fixed

Blackbaud-SteveBrush (approved):

PR #13, opened 2022-02-28
docs: Add code example for box component

Blackbaud-AlexKingman:


apps/code-examples/src/app/app.component.html
```diff
@@ -7,6 +7,12 @@
       <li><a routerLink="action-bars/tab">Tab</a></li>
     </ul>
   </li>
+  <li>
+    Layout
+    <ul>
+      <li><a routerLink="box/basic">Basic</a></li>
```
Blackbaud-SteveBrush: See comment below, route would be `layout/box/basic`.

apps/code-examples/src/app/features/box.module.ts
```diff
@@ -0,0 +1,23 @@
...
+@NgModule({
+  imports: [RouterModule.forChild(routes)],
+  exports: [RouterModule],
+})
+export class BoxRoutingModule {}
+
+@NgModule({
+  imports: [BasicModule, BoxRoutingModule],
+})
+export class BoxModule {}
```
Blackbaud-SteveBrush: I figured we could create modules for each library (not necessarily each component). So this would be the `layout.module.ts`

(The `action-bars.module.ts` was probably a bad component to start with since its library and component are named the same thing.)

Blackbaud-SteveBrush: We'd follow this same structure with the playground and integration SPAs as well.

Blackbaud-SteveBrush (approved):

PR #35, opened 2022-03-03
ci: vulnerabilities remediation

johnhwhite:


package.json
```diff
@@ -96,9 +96,9 @@
     "fs-extra": "10.0.0",
     "inquirer": "8.2.0",
     "jasmine": "3.6.4",
```
Blackbaud-SteveBrush: There's a `3.8.0` version of `jasmine`, too, if we want to keep the two packages aligned.

Blackbaud-SteveBrush (approved):

PR #49, opened 2022-03-08
fix(components/datetime): emit value change event only once when setting the control value of a datepicker

Blackbaud-TrevorBurch:


apps/playground/src/app/components/datetime/date-range-picker/date-range-picker.component.html
```diff
@@ -0,0 +1,99 @@
...
+    </tr>
+  </table>
+
+  <div *ngIf="pickerFormControl.errors" class="sky-error-label">
+    {{ pickerFormControl.errors | json }}
+  </div>
+
+  <h3>&lt;sky-date-range-picker&gt; value</h3>
+
+  <!-- <pre><code style="display:block;">calculatorId: {{ reactiveForm.value.lastDonation?.calculatorId }}
```
Blackbaud-SteveBrush: This comment block still needed?

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
```diff
@@ -273,27 +275,33 @@ export class SkyDateRangePickerComponent
...
-      // This is needed to address a bug in Angular 4.
-      // When a control value is set intially, its value is not represented on the view.
-      // See: https://github.com/angular/angular/issues/13792
-      /* istanbul ignore else */
-      if (this.control) {
-        this.control.setValue(this.value, {
-          emitEvent: false,
-        });
-      }
+      this.ngZone.onStable.pipe(first()).subscribe(() => {
```
Blackbaud-SteveBrush, line 281: Why do we need to use `ngZone`? Mind adding a comment above for future reference?

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
```diff
@@ -476,44 +484,28 @@ export class SkyDateRangePickerComponent
     // For example, if we change the value of the form group internally, we don't want the event
     // listeners to be triggered, as those are reserved for user interactions.
```
Blackbaud-SteveBrush: Does this comment still apply now that we're removing `emitEvent: false`?

libs/components/datetime/src/lib/modules/datepicker/datepicker-input-fuzzy.directive.ts
```diff
@@ -213,82 +213,13 @@ export class SkyFuzzyDatepickerInputDirective
...
-
-      if (this.isFirstChange && this._value) {
-        this.isFirstChange = false;
-      }
-
-      this.datepickerComponent.selectedDate = dateValue;
-    }
-
-    this.setInputElementValue(formattedDate || '');
+    this.updateValue(value);
```
Blackbaud-SteveBrush, line 216: Good change.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input-fuzzy.directive.ts
```diff
@@ -604,4 +533,73 @@ export class SkyFuzzyDatepickerInputDirective
   private onTouched = () => {};
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   private onValidatorChange = () => {};
+
+  private updateValue(value: any, isProgramatic = false): void {
```
Blackbaud-SteveBrush: Spelling: `isProgrammatic`.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input-fuzzy.directive.ts
```diff
@@ -604,4 +533,73 @@ export class SkyFuzzyDatepickerInputDirective
...
+      if (isProgramatic) {
+        this.control?.setValue(this._value, { emitEvent: false });
+      } else {
+        this.onChange(this._value);
+      }
+
+      this.datepickerComponent.selectedDate = dateValue;
+    }
+
+    this.setInputElementValue(formattedDate || '');
```
Blackbaud-SteveBrush, line 608: ☝️ Are all these changes a simple copy/paste from the original setter? Any changes that I need to focus on?

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -519,17 +489,11 @@ export class SkyDatepickerInputDirective
...
-
-    // Do not mark the field as "dirty"
-    // if the field has been initialized with a value.
-    if (this.isFirstChange && this.control) {
-      this.control.markAsPristine();
-    }
-
-    if (this.isFirstChange && this._value) {
-      this.isFirstChange = false;
+  private notifyUpdatedValue(isProgramatic = false): void {
```
Blackbaud-SteveBrush: See my comment RE: `isProgrammatic`, above.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -540,4 +504,39 @@ export class SkyDatepickerInputDirective
       this.adapter.setPlaceholder(this.elementRef, this.dateFormat);
     }
   }
+
+  private updateValue(value: any, isProgramatic = false): void {
```
Blackbaud-SteveBrush: Ditto.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input-fuzzy.directive.ts
Blackbaud-SteveBrush: Also, what is meant by "programmatic"? As in, the value change is done _internally_ to the control, and not as a result of a user? If there's not a better word to describe the behavior we should put a JSDoc comment above this method explaining what the `isProgrammatic` argument does.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -540,4 +504,39 @@ export class SkyDatepickerInputDirective
...
+
+    if (dateValue && isValidDateString) {
+      const formattedDate = this.dateFormatter.format(
+        dateValue,
+        this.dateFormat
+      );
+      this.setInputElementValue(formattedDate);
+    } else {
+      this.setInputElementValue(value || '');
+    }
```
Blackbaud-SteveBrush, line 547: ☝️ Are all these changes a simple copy/paste from the original setter? Any changes that I need to focus on?

Blackbaud-SteveBrush (commented):
This is a much cleaner implementation. Nice work. 

For the commit message, we can simplify it to something like:
```
fix(components/datepicker): emit value change event only once when setting the control value
```
Since the message will be output beneath a "Bug fixes" heading, we don't necessarily have to include the word "fix" in the commit's title.

```
### Bug fixes

**components/datepicker:** emit value change event only once when setting the control value
```

apps/playground/src/app/components/datetime/date-range-picker/date-range-picker.component.html
Blackbaud-TrevorBurch: I fixed this. The date pipe can't take in invalid strings and it was causing an issue. So I had commented it out. I changed this to only show the values when it is valid. Not too worried about it since it is just the playground.

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
Blackbaud-TrevorBurch: Added a comment. Using `ngZone` as it is more reliable than `setTimeout`. Alex and I recently used a similar techinique in Flyout based on what Angular CDK does under the hood for focus stuff.

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
Blackbaud-TrevorBurch: Good catch - removed

libs/components/datetime/src/lib/modules/datepicker/datepicker-input-fuzzy.directive.ts
Blackbaud-TrevorBurch: Thanks :)

libs/components/datetime/src/lib/modules/datepicker/datepicker-input-fuzzy.directive.ts
Blackbaud-TrevorBurch: The `isProgrammatic` denotes that it was done via a `setValue` call to the form control or internally. If this is the case - we don't want to fire an `onChange` event as that triggers `valueChange` and `statusChange` and extra time.

Comment added

libs/components/datetime/src/lib/modules/datepicker/datepicker-input-fuzzy.directive.ts
Blackbaud-TrevorBurch: Just a copy and paste of the original setter other than the new if statement for `isProgrammatic` where we set the controls value without using `onChange`.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
Blackbaud-TrevorBurch: Updated

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
Blackbaud-TrevorBurch: Updated

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
Blackbaud-TrevorBurch: Just a copy and paste of the original setter other than the new if statement inside `notifyUpdatedValue` for `isProgrammatic` where we set the controls value without using onChange and then the pass through of the variable to that function.

Blackbaud-SteveBrush (approved):

PR #65, opened 2022-03-17
feat(components/layout): implement scrollable host service in Back To Top component

Blackbaud-TrevorBurch:


apps/integration/src/app/integrations/vertical-tabset-back-to-top/vertical-tabset-back-to-top.component.html
```diff
@@ -0,0 +1,272 @@
+<h2>Vertical tabset</h2>
+<div id="screenshot-vertical-tabset" [@.disabled]="true">
```
Blackbaud-SteveBrush: Instead of `[@.disabled]`, let's use `NoopAnimationsModule` so that animations are disabled for every integration test.

apps/playground/src/app/components/tabs/vertical-tabset/vertical-tabset.component.html
```diff
@@ -0,0 +1,94 @@
+<h2>Vertical tabset</h2>
+<div id="screenshot-vertical-tabset" [@.disabled]="true">
```
Blackbaud-SteveBrush: We probably don't want animations disabled for the playground SPA, yes?

apps/playground/src/app/components/tabs/vertical-tabset/vertical-tabset.component.html
```diff
@@ -0,0 +1,94 @@
...
+      </sky-vertical-tab>
+      <sky-vertical-tab tabHeading="Group 3 Tab 2">
+        Group 3 Tab 2 content
+      </sky-vertical-tab>
+    </sky-vertical-tabset-group>
+  </sky-vertical-tabset>
+</div>
+
+<h2>Vertical tabset with no tab groups</h2>
+<div id="screenshot-vertical-tabs-no-groups" [@.disabled]="true">
```
Blackbaud-SteveBrush: Ditto.

apps/playground/src/app/components/tabs/vertical-tabset/vertical-tabset.component.html
```diff
@@ -0,0 +1,94 @@
...
+    <sky-vertical-tab tabHeading="tab 1"> Tab 1 content </sky-vertical-tab>
+    <sky-vertical-tab tabHeading="tab 2" active="true">
+      Tab 2 content
+    </sky-vertical-tab>
+    <sky-vertical-tab tabHeading="tab 3"> Tab 3 content </sky-vertical-tab>
+  </sky-vertical-tabset>
+</div>
+
+<h2>Vertical tabset dynamically generated by an ngFor</h2>
+<div id="screenshot-vertical-dynamic-tabs" [@.disabled]="true">
```
Blackbaud-SteveBrush: Ditto.

libs/components/core/src/lib/modules/dock/dock.service.ts
```diff
@@ -15,7 +15,7 @@ import { sortByStackOrder } from './sort-by-stack-order';
  * This service docks components to specific areas on the page.
  */
 @Injectable({
-  providedIn: 'root',
+  providedIn: 'any',
```
Blackbaud-SteveBrush, line 21: Context for this change? This could have downstream ramifications for our consumers.

libs/components/core/src/lib/modules/scrollable-host/fixtures/scrollable-host.component.fixture.ts
```diff
@@ -40,16 +48,16 @@ export class ScrollableHostFixtureComponent {
   public watchScrollableHost(
     alternativeTarget?: ElementRef
   ): Observable<HTMLElement | Window> {
-    return this.scrollableHostService.watchScrollableHost(
-      alternativeTarget || this.target
-    );
+    return this.scrollableHostService
+      .watchScrollableHost(alternativeTarget || this.target)
+      .pipe(takeUntil(this.ngUnsubscribe));
```
Blackbaud-SteveBrush, line 58: Good change. 👍

libs/components/core/src/lib/modules/viewkeeper/viewkeeper.directive.spec.ts
```diff
@@ -146,12 +146,12 @@ describe('Viewkeeper directive', () => {
 
     validateViewkeepersCreated(fixture);
 
-    // Disconnect is called three times from the scrollable host service when we watch for scrollable parents.
-    expect(mockMutationObserver.disconnect).toHaveBeenCalledTimes(3);
+    // Disconnect is called four times from the scrollable host service when we watch for scrollable parents.
+    expect(mockMutationObserver.disconnect).toHaveBeenCalledTimes(4);
```
Blackbaud-SteveBrush, line 150: Curious about the bump in the number of times `disconnect` is called?

libs/components/layout/src/lib/modules/back-to-top/back-top-top.component.spec.ts
```diff
@@ -62,26 +62,29 @@ function scrollElement(
...
   });
 
   describe('when parent is window', () => {
-    it('should show when backToTopTarget is defined and the target element is scrolled out of view', fakeAsync(() => {
+    beforeEach(() => {
+      TestBed.configureTestingModule({
+        imports: [SkyBackToTopFixturesModule],
+      });
+
+      fixture = TestBed.createComponent(SkyBackToTopFixtureComponent);
```
Blackbaud-SteveBrush: Instead of copying the `TestBed` configuration in multiple places, could we remove the `fixture.detectChanges()` from the first `beforeEach` and populate it to the `it` blocks that require it?

libs/components/layout/src/lib/modules/back-to-top/back-to-top.directive.ts
```diff
@@ -62,16 +66,17 @@ export class SkyBackToTopDirective implements AfterViewInit, OnDestroy {
   constructor(
     private dockService: SkyDockService,
     private domAdapter: SkyBackToTopDomAdapterService,
-    private element: ElementRef
+    private element: ElementRef,
+    private scrollableHostService: SkyScrollableHostService
```
Blackbaud-SteveBrush: Both the `SkyBackToTopDirective` and `SkyBackToTopDomAdapterService` inject the `SkyScrollableHostService`. And it looks like you're passing the `scrollableHost` (which was derived from the scrollable host service) to the adapter (which also has the scrollable host service). Any way to refactor so that the implementation is simplified?

libs/components/core/src/lib/modules/scrollable-host/scrollable-host.service.ts
```diff
@@ -41,29 +41,61 @@ export class SkyScrollableHostService {
...
+
+        documentHiddenElementMutationObserver.observe(
+          document.documentElement,
+          {
+            attributes: true,
+            attributeFilter: ['style.display', 'hidden'],
+            childList: true,
+            subtree: true,
+          }
+        );
```
Blackbaud-SteveBrush: Could we put this block of code into a separate method (much like `observeForScrollableHostChanges`) that describes what the method is observing/listening for?

apps/integration/src/app/integrations/vertical-tabset-back-to-top/vertical-tabset-back-to-top.component.html
Blackbaud-TrevorBurch: Ya. That was a product of our old setup where we needed to be able to control them for visual tests but have an easy way to turn them back on when locally testing. I've removed all these.

apps/playground/src/app/components/tabs/vertical-tabset/vertical-tabset.component.html
Blackbaud-TrevorBurch: Removed

apps/playground/src/app/components/tabs/vertical-tabset/vertical-tabset.component.html
Blackbaud-TrevorBurch: Removed

apps/playground/src/app/components/tabs/vertical-tabset/vertical-tabset.component.html
Blackbaud-TrevorBurch: Removed

libs/components/core/src/lib/modules/dock/dock.service.ts
Blackbaud-TrevorBurch: This is the same issue that we have had to solve for Modals and others where the "providedIn: `root`" causes internationalization problems with dynamically generated components. This is the same fix that we did for modals in order to ensure that all the resource labels are output correctly.

libs/components/core/src/lib/modules/scrollable-host/fixtures/scrollable-host.component.fixture.ts
Blackbaud-TrevorBurch: Thanks

libs/components/core/src/lib/modules/viewkeeper/viewkeeper.directive.spec.ts
Blackbaud-TrevorBurch: We added the new mutation observer for the hidden elements. We then disconnect from it and so we call `disconnect` one more time than we used to.

libs/components/layout/src/lib/modules/back-to-top/back-top-top.component.spec.ts
Blackbaud-TrevorBurch: I was able to modify things to only run an extra `beforeEach` with a few properties in one of the `describe` blocks and moved the other `beforeEach` to outside the main `describe`.

libs/components/core/src/lib/modules/scrollable-host/scrollable-host.service.ts
Blackbaud-TrevorBurch: Yup. Done.

libs/components/layout/src/lib/modules/back-to-top/back-to-top.directive.ts
Blackbaud-TrevorBurch: I've modified things. I was attempting to work within the bounds of the existing public method on the adapter service. However, I went ahead and modified it to only take in an elementRef and then do the service call internally. I think we are ok there since the adapter isn't publicly exported through our index file.

libs/components/core/src/lib/modules/scrollable-host/scrollable-host.service.ts
```diff
@@ -173,6 +201,17 @@ export class SkyScrollableHostService {
     return parent;
   }
 
+  private observeDocumentHiddenElemntChanges(
```
Blackbaud-SteveBrush: "Elemnt"

libs/components/core/src/lib/modules/scrollable-host/scrollable-host.service.ts
```diff
@@ -181,14 +220,14 @@ export class SkyScrollableHostService {
...
-        attributeFilter: ['class', 'style.overflow', 'style.overflow-y'],
+        attributeFilter: ['class', 'style'],
         childList: true,
         subtree: true,
       });
     } else {
       mutationObserver.observe(document.documentElement, {
         attributes: true,
-        attributeFilter: ['class', 'style.overflow', 'style.overflow-y'],
+        attributeFilter: ['class', 'style'],
```
Blackbaud-SteveBrush, line 230: Curious about this change. Won't it be more performant to be more specific?

libs/components/core/src/lib/modules/dock/dock.service.ts
Blackbaud-SteveBrush: In this case, though, couldn't we be in a situation where we have multiple dock host elements? One (at least) for the root injector, and another dock host for each lazy-loaded module that gets spun up. Which component/service is affected by the dock service using `providedIn: 'root'`?

It might be worth looking into using `static` properties for the `dockRef` and `items` instead of `providedIn: 'any'`, but I'm unsure of all the repercussions of that change.

libs/components/core/src/lib/modules/scrollable-host/scrollable-host.service.ts
Blackbaud-TrevorBurch: I actually found that the more specific ones weren't working and I was just missing the test coverage to catch it. I added some tests and switched it back to just `style` as the mutation observer wasn't actually catching anything for `style.overflow` like I thought it would.

libs/components/core/src/lib/modules/scrollable-host/scrollable-host.service.ts
Blackbaud-TrevorBurch: Fixed.

Blackbaud-SteveBrush (approved):

PR #67, opened 2022-03-17
chore(dev-infra): add Nx 13 migrate script

Blackbaud-SteveBrush:


libs/sdk/testing/src/lib/test-utility/test-utility.spec.ts
```diff
@@ -134,9 +134,7 @@ describe('Test utility', () => {
 
     expect(imageUrl).toBe('https://example.com/bg/');
 
-    imageUrl = SkyAppTestUtility.getBackgroundImageUrl(
-      new DebugElement(bgEl, document.body, undefined)
-    );
+    imageUrl = SkyAppTestUtility.getBackgroundImageUrl(new DebugElement(bgEl));
```
Blackbaud-SteveBrush, line 137: The second and third parameters are optional in Angular 12 (and removed in Angular 13).

libs/components/packages/src/schematics/ng-add/ng-add.schematic.spec.ts
```diff
@@ -58,38 +58,13 @@ describe('ng-add.schematic', () => {
...
-      'karma-jasmine-html-reporter': '~1.7.0',
-      'ng-packagr': '^12.1.1',
-      typescript: '~4.3.5',
-    });
+    expect(packageJson.dependencies).toEqual(
+      expect.objectContaining({
+        '@skyux/layout': '5.0.0',
+        '@skyux/core': 'LATEST',
+        '@skyux/i18n': 'LATEST',
+      })
```
Blackbaud-SteveBrush, line 66: No need to check all dependencies; only the SKY UX dependencies are affected by the test.

libs/components/router/src/lib/modules/link/link-external.directive.spec.ts
```diff
@@ -115,7 +115,7 @@ describe('SkyAppLinkExternal Directive', () => {
       By.directive(SkyAppLinkExternalDirective)
     );
     expect(directive.attributes['target']).toEqual('_top');
-    expect(directive.properties['href']).toEqual('testUrl/test');
+    expect(directive.properties['href'].endsWith('testUrl/test')).toBeTrue();
```
Blackbaud-SteveBrush, line 118: Angular 13 inserts the localhost URL into the `href` string. Changing the test here to accommodate both Angular 12 and 13.

scripts/dev-migrate.ts
```diff
@@ -0,0 +1,104 @@
...
+      await runCommand('npm', ['uninstall', '@nrwl/cli', '@nrwl/tao']);
+
+      await runCommand('npx', ['nx', 'migrate', 'latest']);
+
+      await removeUnwantedMigrations();
+
+      await runCommand('npx', ['nx', 'migrate', '--run-migrations']);
+
+      // Update Nx CLI global install.
+      await runCommand('npm', ['install', '--global', 'nx@latest']);
```
johnhwhite: This global install failed on my machine when creating the symlink — not sure why, but also wouldn't necessarily expect this step. Can this be removed?

johnhwhite (commented):
One thing. Obviously lots to follow up on after this gets run.

scripts/dev-migrate.ts
Blackbaud-SteveBrush: Yeah, that's fine. They recommend `npx` anyway. 👍

johnhwhite (approved):

PR #69, opened 2022-03-18
fix(components/lookup): fix tostring call on undefined value

johnhwhite:


Blackbaud-SteveBrush (commented):
Probably need a unit test, yes?

Blackbaud-SteveBrush (approved):

PR #70, opened 2022-03-21
feat(components/core): create a resize observer service as a media query service for modals

johnhwhite:


libs/components/modals/src/lib/modules/modal/modal.component.ts
```diff
@@ -34,7 +37,22 @@ let skyModalUniqueIdentifier = 0;
   templateUrl: './modal.component.html',
   styleUrls: ['./modal.component.scss'],
   animations: [skyAnimationModalState],
-  providers: [SkyModalComponentAdapterService, SkyDockService],
+  providers: [
+    SkyModalComponentAdapterService,
+    SkyDockService,
+    SkyResizeObserverMediaQueryService,
+    SkyResizeObserverService,
```
Blackbaud-SteveBrush: We're providing this in the 'root'. Curious why we need to also provide it in the component?

libs/components/core/src/lib/modules/resize-observer/fixtures/resize-obserable-mock.ts
```diff
@@ -0,0 +1,34 @@
+const originalResizeObserver = global.ResizeObserver;
```
Blackbaud-SteveBrush: The file name has a misspelling "obserable".

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
```diff
@@ -0,0 +1,95 @@
...
+   * Returns the current breakpoint.
+   */
+  public get current(): SkyMediaBreakpoints {
+    return this._current;
+  }
+
+  private currentSubject = new ReplaySubject<SkyMediaBreakpoints>(1);
+
+  private _current: SkyMediaBreakpoints;
+  private _element: Element;
```
Blackbaud-SteveBrush: I'm seeing "'_element' is declared but its value is never read".

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  constructor(private resizeObserverService: SkyResizeObserverService) {}
+
+  public ngOnDestroy(): void {
+    this.removeListeners();
+    this.stopListening.complete();
+    this.currentSubject.complete();
+  }
+
+  public observe(element: Element) {
```
Blackbaud-SteveBrush: Should this method accept an `ElementRef` parameter (instead of an `Element`) on the off chance that we need to abstract away the DOM in our services?

Blackbaud-SteveBrush: Also, needs a return type.

libs/components/core/src/lib/modules/resize-observer/resize-observer.service.ts
```diff
@@ -0,0 +1,50 @@
+import { Injectable, NgZone, OnDestroy } from '@angular/core';
+
+import { Observable, Subject } from 'rxjs';
+import { finalize } from 'rxjs/operators';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SkyResizeObserverService implements OnDestroy {
```
Blackbaud-SteveBrush, line 18: Needs JSDocs with a description.

libs/components/core/src/lib/modules/resize-observer/resize-observer.service.ts
```diff
@@ -0,0 +1,50 @@
...
+        this.callback(entries[0]);
+      }
+    );
+  }
+
+  public ngOnDestroy(): void {
+    this.resizeObserver.disconnect();
+  }
+
+  public observe(element: Element): Observable<ResizeObserverEntry> {
```
Blackbaud-SteveBrush: Needs JSDocs with a description.

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
```diff
@@ -0,0 +1,95 @@
...
+    return this._observable.subscribe((entry) => {
+      this.checkBreakpointChange(entry.contentRect.width);
+    });
+  }
+
+  /**
+   * Suscribes to screen size changes.
+   * @param listener Specifies a function that is called when breakpoints change.
+   */
+  public subscribe(listener: SkyMediaQueryListener): Subscription {
```
Blackbaud-SteveBrush, line 123: The description says "...screen size changes". Is this service designed for screens, elements, or both? Just confirming for my own sake.

libs/components/core/src/lib/modules/resize-observer/resize-observer.service.ts
Blackbaud-SteveBrush: Ditto (from above) on using `ElementRef` instead of `Element`, here.

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+import { SkyMediaBreakpoints } from '../media-query/media-breakpoints';
+import { SkyMediaQueryListener } from '../media-query/media-query-listener';
+
+import { SkyResizeObserverService } from './resize-observer.service';
+
+/**
+ * Acts like the media query service for a container element.
+ */
+@Injectable()
```
Blackbaud-SteveBrush: I'm assuming we want this to always be provided on the component level, hence the lack of `providedIn: 'root'`, here?

libs/components/modals/src/lib/modules/modal/modal.component.ts
```diff
@@ -34,7 +37,22 @@ let skyModalUniqueIdentifier = 0;
...
+    {
+      provide: SkyMediaQueryService,
+      useFactory: /* istanbul ignore next */ (
+        resizeObserverService: SkyResizeObserverService
+      ) => {
+        /* istanbul ignore next */
```
Blackbaud-SteveBrush: Code coverage never reaches this factory method?

libs/components/modals/src/lib/modules/modal/modal.component.ts
```diff
@@ -117,6 +135,7 @@ export class SkyModalComponent implements AfterViewInit {
     private windowRef: SkyAppWindowRef,
     private componentAdapter: SkyModalComponentAdapterService,
     private coreAdapter: SkyCoreAdapterService,
+    private mediaQueryService: SkyResizeObserverMediaQueryService,
```
Blackbaud-SteveBrush: Since we're removing `SkyMediaQueryService` completely from the constructor, what's the point of overwriting its provider in the `providers` array?

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
Blackbaud-SteveBrush: Also, needs JSDocs description.

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  public ngOnDestroy(): void {
+    this.removeListeners();
+    this.stopListening.complete();
+    this.currentSubject.complete();
+  }
+
+  /**
+   * Sets the container element to watch. The `SkyResizeObserverMediaQueryService` will only observe one element at a
+   * time. Any previous subscriptions will be unsubscribed when a new element is observed.
```
Blackbaud-SteveBrush, line 85: Good info in this description. 👍

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
```diff
@@ -0,0 +1,97 @@
...
+  private currentSubject = new ReplaySubject<SkyMediaBreakpoints>(1);
+
+  private _current: SkyMediaBreakpoints;
+  private _observable: Observable<ResizeObserverEntry>;
+  private _breakpoints: {
+    check: (width: number) => boolean;
+    name: SkyMediaBreakpoints;
+  }[] = [
+    {
+      check: (width: number) => width <= 767,
```
Blackbaud-TrevorBurch, line 34: Not necessarily for right now - but I wonder if we shouldn't at some point clean things up so that we don't have these magic numbers scattered around. I see 22 instances of `767` in just typescript files alone.

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
```diff
@@ -0,0 +1,97 @@
...
+    this.currentSubject.complete();
+  }
+
+  /**
+   * Sets the container element to watch. The `SkyResizeObserverMediaQueryService` will only observe one element at a
+   * time. Any previous subscriptions will be unsubscribed when a new element is observed.
+   * @param element
+   */
+  public observe(element: ElementRef): void {
+    this.stopListening.next();
```
Blackbaud-TrevorBurch: Should we reset `current` whenever we observe a new element? We call `checkBreakPointChange` but that won't emit if a second element is observed that had the same breakpoint as a previously observed element. That is maybe ok? I just wanted to clarify.

libs/components/core/src/lib/modules/scrollable-host/scrollable-host.service.ts
```diff
@@ -138,6 +138,7 @@ export class SkyScrollableHostService {
           }
           scrollableHost = newScrollableHost;
           newScrollableHostObservable = new Subject();
+          /* istanbul ignore else */
           if (newScrollableHost) {
```
Blackbaud-TrevorBurch: Was this causing you issues? I was surprised to see this in this PR

libs/components/modals/src/lib/modules/modal/modal.component.spec.ts
```diff
@@ -78,6 +79,7 @@ describe('Modal component', () => {
   beforeEach(() => {
     TestBed.configureTestingModule({
       imports: [SkyModalFixturesModule],
+      providers: [SkyResizeObserverMediaQueryService],
```
Blackbaud-TrevorBurch: Should we just place this provider in the `SkyModalFixturesModule`? That is where all the other setup lives.

libs/components/modals/src/lib/modules/modal/modal.component.ts
```diff
@@ -34,7 +37,22 @@ let skyModalUniqueIdentifier = 0;
...
   animations: [skyAnimationModalState],
-  providers: [SkyModalComponentAdapterService, SkyDockService],
+  providers: [
+    SkyModalComponentAdapterService,
+    SkyDockService,
+    SkyResizeObserverMediaQueryService,
+    SkyResizeObserverService,
+    {
+      provide: SkyMediaQueryService,
+      useFactory: /* istanbul ignore next */ (
```
Blackbaud-TrevorBurch: What is the reasoning for using `useFactory` over just `useClass`?

libs/components/modals/src/lib/modules/modal/modal.component.ts
Blackbaud-TrevorBurch: I could be wrong - but I believe that is being done to make sure all children components of the modal are looking at the `SkyResizeObserverMediaQueryService` instead of the app one for their own size breakpoints.

Blackbaud-TrevorBurch (changes requested):
@johnhwhite - just a few questions on this

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
johnhwhite: Right, there is an additional `observe` call that makes this not a drop-in replacement for the media query service, so pulling it in needs to be an explicit choice.

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
johnhwhite: Good call. I'm trying to limit the scope of this PR, but there is an opportunity to replace a lot of the media-query-ish services with this service.

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
johnhwhite: Good catch.

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
johnhwhite: Good catch, should have been "element size changes"

libs/components/modals/src/lib/modules/modal/modal.component.ts
```diff
@@ -110,13 +123,16 @@ export class SkyModalComponent implements AfterViewInit {
   @ViewChild('modalContentWrapper', { read: ElementRef })
   private modalContentWrapperElement: ElementRef;
 
+  private ngDestroy = new Subject<void>();
```
Blackbaud-SteveBrush: Mind if we rename this to `ngUnsubscribe` like we do in our other components? I have a feeling [we'll need to refactor these at some point](https://itnext.io/rxjs-dont-takeuntil-ce76a01e977), and it would be helpful if they all shared the same name so they are easy to find in the source code.

Blackbaud-SteveBrush: Mind if we rename this to `ngUnsubscribe` like we do in our other components? I have a feeling [we'll need to refactor these at some point](https://itnext.io/rxjs-dont-takeuntil-ce76a01e977), and it would be helpful if they all shared the same name so they are easy to find in the source code.

libs/components/modals/src/lib/modules/modal/modal.component.ts
```diff
@@ -110,13 +123,16 @@ export class SkyModalComponent implements AfterViewInit {
...
+
   constructor(
     private hostService: SkyModalHostService,
     private config: SkyModalConfiguration,
     private elRef: ElementRef,
     private windowRef: SkyAppWindowRef,
     private componentAdapter: SkyModalComponentAdapterService,
     private coreAdapter: SkyCoreAdapterService,
+    private mediaQueryService: SkyMediaQueryService,
```
Blackbaud-SteveBrush: New contructor arguments should be added at the end, and marked as `@Optional` so that existing mocks in consumer unit tests don't break.

Blackbaud-SteveBrush: New contructor arguments should be added at the end, and marked as `@Optional` so that existing mocks in consumer unit tests don't break.

libs/components/modals/src/lib/modules/modal/modal.component.ts
```diff
@@ -193,6 +209,27 @@ export class SkyModalComponent implements AfterViewInit {
       referenceEl: this.modalContentWrapperElement.nativeElement,
       zIndex: 5,
     });
+
+    /* istanbul ignore else */
+    if (this.mediaQueryService instanceof SkyResizeObserverMediaQueryService) {
```
Blackbaud-SteveBrush: Since this service is getting provided in this component's decorator, won't this always be an instance of `SkyResizeObserverMediaQueryService`? If you're seeing compiler errors, maybe you could cast the type:
```
(this.mediaQueryService as SkyResizeObserverMediaQueryService).observe()
```

Blackbaud-SteveBrush: Since this service is getting provided in this component's decorator, won't this always be an instance of `SkyResizeObserverMediaQueryService`?

libs/components/modals/src/lib/modules/modal/modal.component.ts
Blackbaud-SteveBrush: Also, shouldn't we prefix this injector with `@Host()` so that we always use the resize observer?

libs/components/modals/src/lib/modules/modal/modal.module.ts
```diff
@@ -36,5 +41,13 @@ import { SkyModalComponent } from './modal.component';
     SkyModalHeaderComponent,
   ],
   entryComponents: [SkyModalHostComponent],
+  providers: [
+    SkyResizeObserverMediaQueryService,
+    SkyResizeObserverService,
+    {
+      provide: SkyMediaQueryService,
+      useExisting: SkyResizeObserverMediaQueryService,
+    },
```
Blackbaud-SteveBrush: Why do we need to provide the media query service both in the module _and_ in the modal component?

Blackbaud-SteveBrush: Why do we need to provide the media query service both in the module _and_ in the modal component?

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -51,7 +54,8 @@ export class SkyModalHostComponent {
     private adapter: SkyModalAdapterService,
     private injector: Injector,
     private router: Router,
-    private changeDetector: ChangeDetectorRef
+    private changeDetector: ChangeDetectorRef,
+    @Optional() private ngModuleRef: NgModuleRef<SkyModalModule>
```
Blackbaud-SteveBrush: Needs a `?` after `ngModuleRef`.

Blackbaud-SteveBrush: Needs a `?` after `ngModuleRef`.

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
```diff
@@ -0,0 +1,127 @@
...
+import { takeUntil } from 'rxjs/operators';
+
+import { SkyMediaBreakpoints } from '../media-query/media-breakpoints';
+import { SkyMediaQueryListener } from '../media-query/media-query-listener';
+import { SkyMediaQueryService } from '../media-query/media-query.service';
+
+import { SkyResizeObserverService } from './resize-observer.service';
+
+/**
+ * Acts like `SkyMediaQueryService` for a container element, emitting the same responsive breakpoints.
```
Blackbaud-SteveBrush, line 13: @blackbaud-johnly Would appreciate your feedback on all the JSDocs comments being added in this PR.

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
```diff
@@ -0,0 +1,127 @@
...
+    // do not add listeners in the constructor
+  }
+
+  protected notifyBreakpointChange(breakpoint: SkyMediaBreakpoints) {
+    this._currentBreakpoint = breakpoint;
+    this.currentBreakpoint.next(breakpoint);
+  }
+
+  protected removeListeners(): void {
+    this.stopListening.next();
```
Blackbaud-SteveBrush: You might need to call `this.stopListening.complete()` as well.

libs/components/modals/src/lib/modules/modal/modal.component.ts
```diff
@@ -193,6 +209,27 @@ export class SkyModalComponent implements AfterViewInit {
...
+      this.mediaQueryService.observe(this.modalContentWrapperElement);
+      this.hostService.close
+        ?.asObservable()
+        .pipe(takeUntil(this.ngDestroy))
+        .subscribe(() => {
+          /* istanbul ignore else */
+          if (
+            this.mediaQueryService instanceof SkyResizeObserverMediaQueryService
```
Blackbaud-SteveBrush: Same here.

Blackbaud-SteveBrush: Same here.

libs/components/core/src/lib/modules/resize-observer/resize-observer.service.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  public ngOnDestroy(): void {
+    this.resizeObserver.disconnect();
+  }
+
+  /**
+   * Create rxjs observable to get size changes for an element ref.
+   * @param element
+   * @return Observable<ResizeObserverEntry>
```
Blackbaud-SteveBrush: `@param` and `@return` aren't needed since those types are derived from TypeScript metadata.

Blackbaud-SteveBrush: `@param` and `@return` aren't needed since those types are derived from TypeScript metadata.

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
```diff
@@ -0,0 +1,127 @@
...
+  implements OnDestroy
+{
+  /**
+   * Returns the current breakpoint.
+   */
+  public get current(): SkyMediaBreakpoints | undefined {
+    return this._currentBreakpoint;
+  }
+
+  private currentBreakpoint = new ReplaySubject<SkyMediaBreakpoints>(1);
```
Blackbaud-SteveBrush: Should be `ReplaySubject<SkyMediaBreakpoints | undefined>` since we're emitting `undefined` in the constructor.

Blackbaud-SteveBrush: Should be `ReplaySubject<SkyMediaBreakpoints | undefined>` since we're emitting `undefined` in the constructor.

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
```diff
@@ -0,0 +1,127 @@
...
+    super(zone);
+    this.stopListening.subscribe(() => {
+      this.notifyBreakpointChange(undefined);
+    });
+  }
+
+  public ngOnDestroy(): void {
+    this.removeListeners();
+    this.stopListening.complete();
+    this.currentBreakpoint.complete();
```
Blackbaud-SteveBrush: Why not put the `complete` calls in the `removeListeners()` method?

Blackbaud-SteveBrush: Why not put the `complete` calls in the `removeListeners()` method?

Blackbaud-SteveBrush (commented):
This is kind of a big ask, but I think it would be super helpful if you created a demo in the "playground" app with a fluid grid in a modal where we could test this functionality manually. What do you think?

libs/components/modals/src/lib/modules/modal/modal.component.ts
johnhwhite: Fixed the provider. This was necessary for the section form within a modal to pick up `SkyResizeObserverMediaQueryService` as its `SkyMediaQueryService`.

libs/components/core/src/lib/modules/resize-observer/resize-observer-media-query.service.ts
johnhwhite: Switched to `ElementRef`, added return type, added JSDocs description.

libs/components/modals/src/lib/modules/modal/modal.module.ts
johnhwhite: Because there are [two injector hierarchies](https://angular.io/guide/hierarchical-dependency-injection#two-injector-hierarchies). With the component providing `SkyMediaQueryService` with `useExisting: SkyResizeObserverMediaQueryService` it will reuse the same instance but redirect to `SkyResizeObserverMediaQueryService` earlier in the injection chain.

libs/components/core/src/lib/modules/scrollable-host/scrollable-host.service.ts
johnhwhite: I took the change out of this PR. I was getting errors on less than 100% coverage on core.

libs/components/core/src/lib/modules/resize-observer/resize-observer.service.ts
johnhwhite: Switched to `ElementRef`, added JSDocs description.

libs/components/modals/src/lib/modules/modal/modal.component.spec.ts
johnhwhite: No longer applies.

libs/components/modals/src/lib/modules/modal/modal.component.ts
johnhwhite: Switched it. No longer applies.

libs/components/modals/src/lib/modules/modal/modal.component.ts
johnhwhite: Ended up not needing this.

libs/components/modals/src/lib/modules/modal/modal.component.ts
johnhwhite: Good catch.

libs/components/modals/src/lib/modules/modal/modal.component.ts
johnhwhite: Good call. Now casting the type.

Blackbaud-SteveBrush (approved):

Blackbaud-TrevorBurch (approved):

PR #71, opened 2022-03-22
fix(components/router): fix skyhref resolution with onpush change detection

johnhwhite:


johnhwhite:
[AB#2096650](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2096650)

libs/components/router/src/lib/modules/href/href.directive.spec.ts
```diff
@@ -76,6 +76,7 @@ describe('SkyHref Directive', () => {
     expect(links.filter((e: HTMLElement) => !!e.offsetParent).length).toEqual(
       6
     );
+    flush();
```
Blackbaud-TrevorBurch: Instead of doing these `flush` calls at the end of all these tests is there harm in just adding them to an `afterEach` so that we don't have to remember that line on any new test?

Blackbaud-TrevorBurch (changes requested):
@johnhwhite one question

libs/components/router/src/lib/modules/href/href.directive.ts
```diff
@@ -55,6 +57,8 @@ export class SkyHrefDirective {
     private router: Router,
     private renderer: Renderer2,
     private element: ElementRef,
+    private applicationRef: ApplicationRef,
+    private changeDetectorRef: ChangeDetectorRef,
```
Blackbaud-SteveBrush: Since it's possible that folks are mocking this directive in their unit tests, any new parameters need to be `@Optional()`.

The only way around this is to provide a test fixture/harness so they don't need to mock it out themselves.

Blackbaud-SteveBrush (approved):
LGTM

libs/components/router/src/lib/modules/href/href.directive.spec.ts
johnhwhite: I was able to make the timer conditional and eliminate the need for all of these calls.

Blackbaud-TrevorBurch (approved):

PR #75, opened 2022-03-23
fix(components/flyout): navigating when a flyout is open does not throw an error

Blackbaud-TrevorBurch:
I'm removing that `applicationRef` `tick` call here as it was causing consumers to receive "Application ref was called recursively" errors. I did some manual testing using the playground page that I added and found that it wasn't necessary. I'm honestly not 100% sure why I felt that the `tick` was needed (and even put a comment in) but could not reproduce the scenario when testing for this change.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -2274,7 +2274,6 @@
             ],
             "styles": [
               "libs/components/theme/src/lib/styles/sky.scss",
-              "libs/components/theme/src/lib/styles/_mixins.scss",
               "libs/components/theme/src/lib/styles/themes/modern/styles.scss",
               "libs/components/ag-grid/src/lib/styles/ag-grid-styles.scss",
               "apps/playground/src/styles.scss"
--- a/angular.json
+++ b/angular.json
@@ -2274,6 +2274,7 @@
             ],
             "styles": [
               "libs/components/theme/src/lib/styles/sky.scss",
+              "libs/components/theme/src/lib/styles/_mixins.scss",
               "libs/components/theme/src/lib/styles/themes/modern/styles.scss",
               "libs/components/ag-grid/src/lib/styles/ag-grid-styles.scss",
               "apps/playground/src/styles.scss"
--- a/libs/components/flyout/src/lib/modules/flyout/flyout.service.ts
+++ b/libs/components/flyout/src/lib/modules/flyout/flyout.service.ts
@@ -1,8 +1,10 @@
 import {
+  ApplicationRef,
   ComponentRef,
   Injectable,
   NgZone,
   OnDestroy,
+  Optional,
   Type,
 } from '@angular/core';
 import { NavigationStart, Router } from '@angular/router';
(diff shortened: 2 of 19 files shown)


angular.json
```diff
@@ -2274,6 +2274,7 @@
             ],
             "styles": [
               "libs/components/theme/src/lib/styles/sky.scss",
+              "libs/components/theme/src/lib/styles/_mixins.scss",
```
Blackbaud-SteveBrush: Is this file actually needed? This is usually imported by a component's SCSS file.

libs/components/flyout/src/lib/modules/flyout/flyout.service.ts
```diff
@@ -42,8 +41,7 @@ export class SkyFlyoutService implements OnDestroy {
     private windowRef: SkyAppWindowRef,
     private dynamicComponentService: SkyDynamicComponentService,
     private router: Router,
-    private readonly _ngZone: NgZone,
-    private readonly applicationRef: ApplicationRef
```
Blackbaud-SteveBrush, line 46: Removing this argument will cause problems for consumers that mock the flyout service in their tests: https://dev.azure.com/blackbaud/_search?action=contents&type=code&text=provide%3A%20SkyFlyoutService&result=DefaultCollection/Products/skyux-spa-reportview/GBmaster//src/app/components/dashboard-view/dashboard-view.component.spec.ts

libs/components/flyout/src/lib/modules/flyout/flyout.service.ts
```diff
@@ -97,8 +94,6 @@ export class SkyFlyoutService implements OnDestroy {
             this._ngZone.onStable.pipe(take(1)).subscribe(() => {
               if (this.host) {
                 this.removeHostComponent();
-                // Without this tick - the host does not actually get removed on initial navigation in this case.
-                this.applicationRef.tick();
```
Blackbaud-SteveBrush, line 101: Does the original PR/story that introduced this change give any insight as to _why_ we added this to begin with?

libs/components/flyout/src/lib/modules/flyout/flyout.service.ts
Blackbaud-TrevorBurch: Added back as `Optional`

angular.json
Blackbaud-TrevorBurch: Removed

libs/components/flyout/src/lib/modules/flyout/flyout.service.ts
Blackbaud-TrevorBurch: No. I did this recently. The issue I was seeing at the time was that change detection wasn't actually removing the host when things were called. However, when I tested during this - I was not seeing that same issue. Also, this block of code was for a sanity check for a state that consumers should never end up in and we only added it because an old bug was getting people into this state.

Blackbaud-SteveBrush (approved):

PR #77, opened 2022-03-23
fix(components/tabs): fix disappearing URL params

bradenbiz:


apps/code-examples/src/app/code-examples/flyout/flyout/basic/flyout-demo.module.ts
```diff
@@ -11,4 +11,4 @@ import { FlyoutDemoComponent } from './flyout-demo.component';
   entryComponents: [FlyoutDemoFlyoutComponent],
   exports: [FlyoutDemoComponent],
 })
-export class AutocompleteDemoModule {}
+export class FlyoutDemoModule {}
```
Blackbaud-SteveBrush, line 13: LOL

apps/integration/src/app/integrations/flyout-with-tabs/basic/flyout-demo.component.ts
```diff
@@ -0,0 +1,59 @@
+import { Component } from '@angular/core';
```
Blackbaud-SteveBrush: I'm not seeing any integration tests/spec files? Was this supposed to go in the playground SPA?

apps/code-examples/src/app/code-examples/flyout/flyout/basic/flyout-demo.module.ts
bradenbiz: goteem!

apps/integration/src/app/integrations/flyout-with-tabs/basic/flyout-demo.component.ts
bradenbiz: No, I need to add those! Thanks for the reminder.

apps/integration/src/app/home/home.component.html
```diff
@@ -4,4 +4,9 @@
       >Vertical Tabset with Back To Top</a
     >
   </li>
+  <li>
```
Blackbaud-TrevorBurch: Nitpicky - could you potentially place this new one first so that they are in alphabetical order? That may make it easier to find things quickly once we add more things.

apps/integration/src/app/integrations/flyout-tabs/flyout-tabs.component.ts
```diff
@@ -0,0 +1,59 @@
...
+  SkyFlyoutConfig,
+  SkyFlyoutInstance,
+  SkyFlyoutService,
+} from '@skyux/flyout';
+
+import { FlyoutTabsContentComponent } from './flyout-tabs-content.component';
+
+screenY;
+@Component({
+  selector: 'app-flyout-demo',
```
Blackbaud-TrevorBurch: Nitpicky - would you mind changing this selector to `app-flyout-tabs` so that it matches the file name and the component name?

apps/integration/src/app/integrations/flyout-tabs/flyout-tabs.module.ts
```diff
@@ -0,0 +1,15 @@
...
+import { SkyFlyoutModule } from '@skyux/flyout';
+import { SkyTabsModule } from '@skyux/tabs';
+
+import { FlyoutTabsContentComponent } from './flyout-tabs-content.component';
+import { FlyoutTabsComponent } from './flyout-tabs.component';
+
+@NgModule({
+  imports: [CommonModule, SkyFlyoutModule, SkyTabsModule],
+  declarations: [FlyoutTabsComponent, FlyoutTabsContentComponent],
+  entryComponents: [FlyoutTabsContentComponent],
```
Blackbaud-TrevorBurch: You shouldn't need `entryComponents` in Angular 12

Blackbaud-TrevorBurch (dismissed):
@bradenbiz - just a couple nitpicky things - along with the missing integration tests.

apps/integration/src/app/integrations/flyout-tabs/flyout-tabs.module.ts
bradenbiz: The `code-examples` SPA has tons of `entryComponents` set in them. Should I remove all of those while I'm at it?

Blackbaud-TrevorBurch: I would leave them for now and do that separately. We can file a bug for it.

bradenbiz: I'll do that.

bradenbiz: https://blackbaud.visualstudio.com/Products/_workitems/edit/2105675

Done!^

apps/integration/tsconfig.json
```diff
@@ -12,5 +12,16 @@
...
+    "forceConsistentCasingInFileNames": true,
+    "strict": false,
+    "noImplicitReturns": false,
+    "noFallthroughCasesInSwitch": true
+  },
+  "angularCompilerOptions": {
+    "strictInjectionParameters": false,
+    "strictInputAccessModifiers": false,
+    "strictTemplates": false
+  }
```
Blackbaud-SteveBrush: Strict mode is disabled by default, so you can likely just remove these altogether? This isn't related to the tabset fix?

libs/components/tabs/src/lib/modules/tabs/tabset.component.spec.ts
```diff
@@ -1574,4 +1593,52 @@ describe('Tabset component', () => {
...
+        'example-path',
+        'example-child-path',
+      ]);
+
+      fixture.detectChanges();
+      tick();
+      fixture.detectChanges();
+      tick();
+      fixture.detectChanges();
+      tick();
```
Blackbaud-SteveBrush: I know this was probably a copy/paste of another test, but I was able to get the test to pass with only one set of:
```
fixture.detectChanges();
tick();
```
(Same for the two blocks, below.)

libs/components/tabs/src/lib/modules/tabs/tabset.component.spec.ts
```diff
@@ -1574,4 +1593,52 @@ describe('Tabset component', () => {
       validateTabSelected(fixture.nativeElement, 0);
     }));
   });
+
+  describe('Permalinks with non-base URLs', () => {
```
Blackbaud-SteveBrush: What's meant by "non-base URL"? 

Does this spec need its own `describe`? Could it instead be added after the `should not attempt to remove the query param if permalinkId is not set` test?

apps/integration/src/app/app.module.ts
```diff
@@ -7,7 +7,7 @@ import { AppComponent } from './app.component';
 
 @NgModule({
   declarations: [AppComponent],
-  imports: [BrowserModule, BrowserAnimationsModule, AppRoutingModule],
+  imports: [BrowserAnimationsModule, BrowserModule, AppRoutingModule],
```
Blackbaud-SteveBrush: Since we're cherry-picking PRs between multiple versions now, it'd probably be good to avoid any changes that aren't directly related to the fix or feature of the current PR (so that we avoid merge conflicts between the branches).

apps/code-examples/src/app/code-examples/flyout/flyout/grid/flyout-demo.module.ts
```diff
@@ -24,4 +24,4 @@ import { FlyoutDemoComponent } from './flyout-demo.component';
   declarations: [FlyoutDemoComponent, FlyoutDemoFlyoutComponent],
   exports: [FlyoutDemoComponent],
 })
-export class AutocompleteDemoModule {}
+export class FlyoutDemoModule {}
```
Blackbaud-SteveBrush, line 27: ☝️ These are good changes!

apps/integration/tsconfig.json
bradenbiz: Nope, it isn't. The `integration` SPA did need updates to the `tsconfig.json`, but these here are just remnants of me playing around to fix the issues I was running into. I'll remove them.

libs/components/tabs/src/lib/modules/tabs/tabset.component.spec.ts
bradenbiz: Good point. Thanks! Will remove the extras.

libs/components/tabs/src/lib/modules/tabs/tabset.component.spec.ts
bradenbiz: "non-base URL" meaning a URL that is not `/`, ie. `/example-path/example-child-path`. If you have a better way of describing this I'll happily make a change.

It does need its own `describe`. It needs to use the SkyTabsetPermalinksFixtureComponent without running the `afterEach()` from the "Permalinks" `describe` block after the test has been run.

apps/integration/src/app/app.module.ts
bradenbiz: Good call. Fixing.

apps/code-examples/src/app/code-examples/flyout/flyout/grid/flyout-demo.module.ts
bradenbiz: Lol thanks! 😁

libs/components/tabs/src/lib/modules/tabs/tabset.component.spec.ts
Blackbaud-SteveBrush: I was saying you could put this test in the "root" `describe` block, outside of the "Permalink" `describe`.

apps/integration/tsconfig.json
```diff
@@ -12,5 +12,11 @@
...
       "path": "./tsconfig.editor.json"
     }
-  ]
+  ],
+  "compilerOptions": {
+    "forceConsistentCasingInFileNames": true,
+    "strict": false,
+    "noImplicitReturns": false,
+    "noFallthroughCasesInSwitch": true
+  }
```
Blackbaud-SteveBrush: This change still needed?

libs/components/tabs/src/lib/modules/tabs/tabset.component.spec.ts
Blackbaud-SteveBrush: I see what you're saying. That `afterEach` was poorly written (by me). You could move the test to the "Permalinks" `describe`, and modify the `afterEach` like so:
```
afterEach(() => {
  const spy = spyOn(location, 'go');

  fixture.destroy();

  if (spy.calls.all().length > 0) {
    expect(
      spy.calls.mostRecent().args[0].indexOf('?foobar-active-tab')
    ).toBe(
      -1,
      'The permalink param should be cleared when the tabset is destroyed.'
    );
  }
});
```
This would prevent the confusion of the new `describe` block, and allow your test to be executed in the "Permalink" `describe`.

libs/components/tabs/src/lib/modules/tabs/tabset.component.spec.ts
```diff
@@ -1574,4 +1593,40 @@ describe('Tabset component', () => {
...
+    let fixture: ComponentFixture<SkyTabsetPermalinksFixtureComponent>;
+    let location: Location;
+
+    beforeEach(() => {
+      location = TestBed.inject(Location);
+      fixture = TestBed.createComponent(SkyTabsetPermalinksFixtureComponent);
+    });
+
+    it('should not affect the base URL when opened or closed', fakeAsync(() => {
```
Blackbaud-SteveBrush: If you choose to move the spec to the "Permalinks" `describe` block, you could rename this to: "should not affect the current route when destroyed".

apps/integration/tsconfig.json
bradenbiz: hm. I guess not. Not sure why I thought I needed that before...

libs/components/tabs/src/lib/modules/tabs/tabset.component.spec.ts
bradenbiz: done

Blackbaud-SteveBrush (approved):

PR #79, opened 2022-03-23
fix(components/action-bars): summary action bar summaries can be added and removed dynamically

Blackbaud-TrevorBurch:


libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.spec.ts
```diff
@@ -10,6 +10,7 @@ import {
 import { By } from '@angular/platform-browser';
 import { expect } from '@skyux-sdk/testing';
 import { SkyMediaBreakpoints, SkyMediaQueryService } from '@skyux/core';
+// eslint-disable-next-line @nrwl/nx/enforce-module-boundaries
```
bradenbiz, line 13: Why does this need to be added here?

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.ts
```diff
@@ -64,7 +64,16 @@ export class SkySummaryActionBarComponent implements AfterViewInit, OnDestroy {
...
+  private set summaryELement(element: ElementRef) {
+    this._summaryElement = element;
+    this.changeDetector.markForCheck();
+  }
+
+  private get summaryElement(): ElementRef {
+    return this._summaryElement;
+  }
+
+  private _summaryElement: ElementRef;
```
bradenbiz, line 76: How does this fix the issue at hand? From what I can see, it seems all you've done is set up getters and setters for this ElementRef.

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.spec.ts
```diff
@@ -246,6 +264,25 @@ describe('Summary Action Bar component', () => {
         expect(cmp.summaryActionBar.isSummaryCollapsible).toBeTruthy();
       });
 
+      it('should recognize when the summary tag when it is toggled externally when on a xs screen', () => {
```
bradenbiz, line 267: Why is a separate test needed for an `xs` screen but not the other viewport sizes?

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.spec.ts
Blackbaud-TrevorBurch: This was just for a warning that the IDE was throwing. It appeared to be because the testing library it is coming from isn’t a true library and the tests are in one. However, the tests aren’t actually built so it is ok.

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.ts
Blackbaud-TrevorBurch: The key here is the `markForCheck`. When the `ngIf` is toggled externally the internal change detection doesn’t fire because the component is `OnPush`. Adding the `markForCheck` allows the internal change detection to run and for the template to update accordingly.

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.spec.ts
Blackbaud-TrevorBurch: I added that as this issue was most prominent in the mobile view during my manual testing.

bradenbiz (approved):

PR #81, opened 2022-03-24
docs: remove entry components from code examples and test files

Blackbaud-TrevorBurch:


libs/components/toast/src/lib/modules/toast/toast.module.ts
```diff
@@ -18,10 +18,5 @@ import { SkyToasterComponent } from './toaster.component';
     SkyToastResourcesModule,
   ],
   exports: [SkyToastComponent],
-  entryComponents: [
```
Blackbaud-SteveBrush: Did we confirm that we can also remove `entryComponents` for library modules? I seem to remember Angular 12's "partial" compilation mode still requires entry components be defined.

Blackbaud-TrevorBurch: Reverted these

Blackbaud-SteveBrush (approved):

PR #87, opened 2022-03-28
fix(components/ag-grid): ag-grid: display validator popover using a delayed hover event

johnhwhite:
[AB#1929223](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/1929223)


johnhwhite:
[AB#1929223](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/1929223)

libs/components/ag-grid/src/lib/modules/ag-grid/cell-validator/ag-grid-cell-validator-tooltip.component.ts
```diff
@@ -50,10 +52,30 @@ export class SkyAgGridCellValidatorTooltipComponent {
       }
     });
 
+    this.cellRendererParams.eGridCell?.addEventListener('mouseover', () => {
+      if (!this._hoverTimeout) {
+        this._hoverTimeout = setTimeout(() => {
```
Blackbaud-SteveBrush: You can call `window.setTimeout`, here, to instruct the TypeScript compiler to reference the DOM's version of `setTimeout`, instead of importing the `Timeout` type from `NodeJS`. We probably want to use the DOM API since this component is used on a webpage?

Blackbaud-SteveBrush: Also, it would be helpful to future maintainers if we provided a comment explaining why a timeout is needed, with a link to a StackOverflow/Github issue, if applicable.

Blackbaud-AdamFunderburk:
Reviewed this with some others, and decided that the danger icon should remain when then popover appears.

Blackbaud-AdamFunderburk:
Possible bug: ESC isn't dropping focus on the editable currency cells in the demo. The other cell types work as expected.

Blackbaud-AdamFunderburk:
It seems unlikely, but is it possible to remove the popover through a longer transition animation when the cell changes to editable? Can we cheat by making it fade out after 500, or even 1000ms? Or is it just going to be blown away regardless once the cell switches modes?

johnhwhite:
> Reviewed this with some others, and decided that the danger icon should remain when then popover appears.

This change has been implemented.

> Possible bug: ESC isn't dropping focus on the editable currency cells in the demo. The other cell types work as expected.

This was an issue with `autoNumeric` and has been addressed.

> It seems unlikely, but is it possible to remove the popover through a longer transition animation when the cell changes to editable? Can we cheat by making it fade out after 500, or even 1000ms? Or is it just going to be blown away regardless once the cell switches modes?

This is not possible in the current implementation.

libs/components/ag-grid/src/lib/modules/ag-grid/cell-validator/ag-grid-cell-validator-tooltip.component.ts
johnhwhite: The timeout is being used to schedule the popover. I've wrapped this logic in `scheduleDelayedPopover` and `cancelDelayedPopover` functions.

libs/components/ag-grid/src/lib/modules/ag-grid/cell-validator/ag-grid-cell-validator-tooltip.component.ts
```diff
@@ -76,36 +83,39 @@ export class SkyAgGridCellValidatorTooltipComponent {
...
+    this.cancelDelayedPopover();
     this.popoverMessageStream.next({ type: SkyPopoverMessageType.Open });
   }
+
+  private scheduleDelayedPopover() {
+    /* istanbul ignore else */
+    if (!this._hoverTimeout) {
+      this._hoverTimeout = window.setTimeout(() => {
+        this.showPopover();
+      }, 300);
```
Blackbaud-SteveBrush, line 111: Is this number precise? If we dropped it to `0` would it fail to work? Will `300` _always_ work?

libs/components/ag-grid/src/lib/modules/ag-grid/cell-validator/ag-grid-cell-validator-tooltip.component.ts
```diff
@@ -50,6 +49,14 @@ export class SkyAgGridCellValidatorTooltipComponent {
       }
     });
 
+    this.cellRendererParams.eGridCell?.addEventListener('mouseover', () => {
```
Blackbaud-SteveBrush: The `mouseover` and `mouseout` events can be expensive events to listen to. Can the feature be resolved with `mouseenter` and `mouseleave`?

apps/code-examples/src/app/code-examples/ag-grid/data-entry-grid/basic/data-entry-grid-docs-demo-edit-modal.component.html
```diff
@@ -1,6 +1,6 @@
 <sky-modal>
   <sky-modal-header> Edit employee information </sky-modal-header>
-  <sky-modal-content>
+  <sky-modal-content (keyup.escape)="onPressEscape($event)">
```
Blackbaud-SteveBrush: Can this behavior be added to our `sky-ag-grid-wrapper` component? It could stop propagation if the form is in edit mode, or let it bubble-up otherwise.

libs/components/ag-grid/src/lib/modules/ag-grid/cell-validator/ag-grid-cell-validator-tooltip.component.ts
johnhwhite: `300` was the feature requirement from UX.

libs/components/ag-grid/src/lib/modules/ag-grid/cell-validator/ag-grid-cell-validator-tooltip.component.ts
johnhwhite: Switched the events.

apps/code-examples/src/app/code-examples/ag-grid/data-entry-grid/basic/data-entry-grid-docs-demo-edit-modal.component.html
johnhwhite: I was able to implement this in `sky-ag-grid-wrapper` but the "edit mode" calculation did not work within this event handler. The user would need to tab outside of the grid to use escape.

Blackbaud-SteveBrush (approved):

PR #90, opened 2022-03-29
feat(component/ag-grid): use modules instead of packages to reduce bundle size

Blackbaud-KevinEnns:
Using the module import instead of the package import allows for smaller bundle sizes.


Blackbaud-KevinEnns:
This is related to ADO User story [1981305](https://dev.azure.com/blackbaud/Products/_workitems/edit/1981305).

libs/components/ag-grid/package.json
```diff
@@ -30,8 +30,8 @@
     "@skyux/lookup": "0.0.0-PLACEHOLDER",
     "@skyux/popovers": "0.0.0-PLACEHOLDER",
     "@skyux/theme": "0.0.0-PLACEHOLDER",
-    "ag-grid-angular": "^26.0.0",
-    "ag-grid-community": "^26.0.0"
+    "@ag-grid-community/angular": "^26.0.0",
+    "@ag-grid-community/core": "^26.0.0"
```
Blackbaud-SteveBrush, line 34: Thanks for the pull request! Our consumers are expecting to see peers of `ag-grid-angular` and `ag-grid-community`, so changing the peer dependencies would be considered a "breaking change", since it would require the consumer to uninstall the `ag-grid-community` package, and install `@ag-grid-community/core`. The consumer would also need to update their import statements to reflect the new package locations.

This pull request _does_ introduce some interesting improvements, however, so I don't want to discard it outright. I'll bring this up to the team to see what options we have to introduce this change to our consumers.

Blackbaud-SteveBrush (changes requested):

libs/components/ag-grid/package.json
Blackbaud-SteveBrush: Update: we created a story to research adding this to SKY UX 6. 👍

Blackbaud-KevinEnns: Thanks!

Blackbaud-TrevorBurch:
Closing this PR for now as it was against the 5.x.x and we will still have it historically if/when we look at this again.

PR #104, opened 2022-04-04
docs: add modal classes content, fix misspellings

blackbaud-paulah:


Blackbaud-SteveBrush (dismissed):

Blackbaud-SteveBrush (dismissed):

libs/components/modals/src/lib/modules/modal/modal-close-args.ts
```diff
@@ -1,4 +1,15 @@
+/**
+ * Contains an object with the data passed from users when
+ * a modal is closed, and the reason that the modal was closed.
```
blackbaud-johnly:
```suggestion
 * a modal is closed and the reason that the modal was closed.
```
I don't think we need a comma here since it's not a compound sentence.

blackbaud-johnly (approved):

PR #113, opened 2022-04-05
feat(components/ag-grid): add text cell maxlength and number cell min/max options

johnhwhite:
[AB#1583254](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/1583254)

### DIFF
--- a/libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-number/cell-editor-number.component.ts
+++ b/libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-number/cell-editor-number.component.ts
@@ -8,6 +8,9 @@ import {
 import { ICellEditorAngularComp } from 'ag-grid-angular';
 import { ICellEditorParams } from 'ag-grid-community';
 
+import { SkyCellEditorDatepickerParams } from '../../types/cell-editor-datepicker-params';
+import { SkyCellEditorNumberParams } from '../../types/cell-editor-number-params';
+
 /**
  * @internal
  */
@@ -25,6 +28,8 @@ export class SkyAgGridCellEditorNumberComponent
   public columnWidth: number;
   public rowHeightWithoutBorders: number;
   public rowNumber: number;
+  public max: number;
+  public min: number;
 
   private params: ICellEditorParams;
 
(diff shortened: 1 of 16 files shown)


libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-text/cell-editor-text.component.ts
```diff
@@ -26,6 +28,7 @@ export class SkyAgGridCellEditorTextComponent
   public columnWidth: number;
   public rowHeightWithoutBorders: number;
   public rowNumber: number;
+  public maxlength: number;
```
Blackbaud-SteveBrush, line 31: Should be `number | undefined`.

Blackbaud-SteveBrush, line 31: Should be `number | undefined`.

libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-number/cell-editor-number.component.ts
```diff
@@ -25,6 +28,8 @@ export class SkyAgGridCellEditorNumberComponent
   public columnWidth: number;
   public rowHeightWithoutBorders: number;
   public rowNumber: number;
+  public max: number;
+  public min: number;
```
Blackbaud-SteveBrush, line 32: ☝️ Both should be `number | undefined`.

Blackbaud-SteveBrush, line 32: ☝️ Both should be `number | undefined`.

libs/components/ag-grid/src/lib/modules/ag-grid/types/text-properties.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * @internal
+ */
+export interface SkyTextProperties {
+  maxlength?: number;
+}
```
Blackbaud-SteveBrush, line 6: ☝️ What's the value of separating the `SkyNumberProperties` and `SkyTextProperties` from their `SkyCellEditorNumberParams` and `SkyCellEditorTextParams` interfaces? Feels like it's adding needless complexity. For example:

```
export interface SkyCellEditorNumberParams extends ICellEditorParams {
  skyComponentProperties?: { max?: number; min?: number; };
}
```

If we need to keep them separated for some reason, we should rename them to be more specific to AgGrid in case we need to export them to the public API in the future (`SkyNumberProperties` is a bit too generic).

Blackbaud-SteveBrush, line 6: ☝️ What's the value of separating the `SkyNumberProperties` and `SkyTextProperties` from their `SkyCellEditorNumberParams` and `SkyCellEditorTextParams` interfaces? Feels like it's adding needless complexity. For example:

```
export interface SkyCellEditorNumberParams extends ICellEditorParams {
  skyComponentProperties?: { max?: number; min?: number; };
}
```

If we need to keep them separated for some reason, we should rename them to be more specific to AgGrid in case we need to export them to the public API in the future (`SkyNumberProperties` is a bit too generic).

Blackbaud-SteveBrush (approved):

PR #129, opened 2022-04-08
feat(components/datetime):  date-picker shortcut functionality

rakesh-garg:
If you are entering a date into a charge (this also happens in single contributions), and you clear the date field and enter "28," you would expect it to be smart enough to render that as 04/28/2021. It does not, it renders it as 01/01/2028.
If you enter "0201" it will render it as 02/01/2021 (which is correct). FE NXT renders the "28" as 04/28/2021, saving users keystrokes.
If you enter a nonsensical date (such as "0332" - clearly there is no March 32nd), it renders it as 01/01/0332. FE NXT renders that as 04/01/2021. This is a bit of an edge case, granted.


Blackbaud-SteveBrush:
Closing in favor of: https://github.com/blackbaud/skyux/pull/179

PR #169, opened 2022-04-18
fix(apps/code-examples): fix type errors for ag-grid basic code example

Blackbaud-SteveBrush:


apps/code-examples/src/app/code-examples/ag-grid/data-entry-grid/basic/data-entry-grid-docs-demo.component.ts
```diff
@@ -160,7 +160,7 @@ export class SkyDataEntryGridDemoComponent {
     });
   }
 
-  public searchApplied(searchText: string | undefined): void {
+  public searchApplied(searchText: string | void): void {
```
Blackbaud-SteveBrush, line 163: Reverting a prior change (it used to be `string | void`). This needs to be a union type because this method is acting as the callback for all three of the Search component's outputs: https://github.com/blackbaud/skyux/blob/main/libs/components/lookup/src/lib/modules/search/search.component.ts#L65-L78

johnhwhite (approved):

PR #174, opened 2022-04-19
deprecation(components/layout): `SkyDefinitionListComponent` is deprecated; use `SkyDescriptionListComponent` instead

Blackbaud-TrevorBurch:


libs/components/layout/src/lib/modules/definition-list/definition-list.component.ts
```diff
@@ -32,5 +33,13 @@ export class SkyDefinitionListComponent {
...
   }
 
-  constructor(public service: SkyDefinitionListService) {}
+  constructor(public service: SkyDefinitionListService, logger: SkyLogService) {
+    logger.deprecated('SkyDefinitionListComponent', {
+      deprecationMajorVersion: 6,
+      moreInfoUrl:
+        'https://developer.blackbaud.com/skyux/components/description-list',
+      replacementRecommendation:
+        'We recommend `SkyDescriptionListComponent` instead.',
```
Blackbaud-SteveBrush: I talked to John Lyons offline about this and he said he liked `Use x instead.` a bit better. I think we should put this sentence in the PR title (and consequently, the changelog), too.

libs/components/layout/src/lib/modules/definition-list/definition-list.module.ts
```diff
@@ -10,6 +10,9 @@ import { SkyDefinitionListLabelComponent } from './definition-list-label.compone
 import { SkyDefinitionListValueComponent } from './definition-list-value.component';
 import { SkyDefinitionListComponent } from './definition-list.component';
 
+/**
+ * @deprecated Use `SkyDescriptionListModule` instead
```
Blackbaud-SteveBrush: Needs a period.

libs/components/layout/src/lib/modules/definition-list/definition-list.component.ts
Blackbaud-TrevorBurch: Updated. The PR title does now have periods - which isn't standard for our modern entries. But it doesn't make sense without them since we have two sentences.

libs/components/layout/src/lib/modules/definition-list/definition-list.module.ts
Blackbaud-TrevorBurch: Done

Blackbaud-SteveBrush (approved):

PR #161, opened 2022-04-15
docs(apps/code-examples): cleanup CSS class usage

Blackbaud-SteveBrush:

PR #177, opened 2022-04-21
fix(components/popovers): popovers placed above or below target should not be assigned a vertical alignment

Blackbaud-SteveBrush:


libs/components/popovers/src/lib/modules/popover/popover.directive.ts
```diff
@@ -220,6 +236,22 @@ export class SkyPopoverDirective implements OnInit, OnDestroy {
...
 
   private sendMessage(messageType: SkyPopoverMessageType): void {
-    this.skyPopoverMessageStream.next({ type: messageType });
+    this.messageStream.next({ type: messageType });
+  }
+
+  private setupMessageStream(value?: Subject<SkyPopoverMessage>): void {
+    this.teardownMessageStream();
+
+    this._messageStream = value ?? new Subject<SkyPopoverMessage>();
```
Blackbaud-PaulCrowder: I generally try to avoid referencing any private fields that back getter/setters in anything other than the getter/setter. That ensures all access (even within the same class) goes through the getter, which sometimes may have logic of its own, like returning default values. I'd move the assignment here back into the setter, then change `_messageStreamSub` to `messageStreamSub` (or better yet, `#messageStreamSub` to make it truly private) since it doesn't back a getter/setter property.

libs/components/popovers/src/lib/modules/popover/popover.directive.ts
```diff
@@ -19,25 +19,40 @@ export class SkyPopoverDirective implements OnInit, OnDestroy {
    * @required
    */
   @Input()
-  public skyPopover: SkyPopoverComponent;
+  public set skyPopover(value: unknown) {
```
Blackbaud-PaulCrowder: Now that you're removing the code example where this is set in the component's class, I'd revert this back to a simple public property of type `SkyPopoverComponent` with no getter/setter.

apps/code-examples/src/app/app.component.html
```diff
@@ -130,6 +130,26 @@
...
   </li>
 
+  <li>
+    Popovers
+    <ul>
+      <li>
+        Popover
+        <ul>
+          <li>
+            <a routerLink="popovers/popover/asynchronous">Asynchronous</a>
```
Blackbaud-PaulCrowder: Should you remove this now that the code example is gone? Or does this point to something else?

apps/code-examples/src/app/code-examples/popovers/popover/basic/popover-demo.component.html
```diff
@@ -1,13 +1,13 @@
 <button
-  class="sky-btn sky-margin-inline-compact"
+  class="sky-btn sky-btn-default sky-margin-inline-compact"
```
Blackbaud-PaulCrowder, line 2: Does `sky-margin-inline-compact` need to change to the "modern" CSS class? Or will you handle that in a subsequent PR?

apps/code-examples/src/app/app.component.html
Blackbaud-SteveBrush: Yep, good catch.

apps/code-examples/src/app/code-examples/popovers/popover/basic/popover-demo.component.html
Blackbaud-SteveBrush: That's been changed in this PR: https://github.com/blackbaud/skyux/pull/161/files#diff-1f41a0be4ff900f211dc873c2ab15d7994809bfec19a2f1d4fd4332ecbd63c6f

Blackbaud-PaulCrowder (approved):

PR #179, opened 2022-04-21
feat(components/datetime): convert a user-entered digit into a date for the datepicker component

Blackbaud-TrevorBurch:


libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -534,7 +576,7 @@ export class SkyDatepickerInputDirective
       this.datepickerComponent.selectedDate = this._value;
     }
 
-    if (dateValue && isValidDateString) {
+    if (dateValue && dateValue instanceof Date && isValidDateString) {
```
Blackbaud-SteveBrush: The variable name `dateValue` was written to imply that the value was an instance of `Date`.  Is there a way to keep the variable of type `Date` to avoid having the `instanceof Date` checks in multiple places?

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -353,16 +353,57 @@ export class SkyDatepickerInputDirective
       return;
     }
 
-    const value: any = control.value;
+    const value: Date | string = control.value;
```
Blackbaud-SteveBrush: Isn't the return value of the control technically an `any`? Number, string, undefined, null, Date, etc.?

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -503,11 +527,29 @@ export class SkyDatepickerInputDirective
...
+      if (Number.isInteger(num)) {
+        const now = new Date();
+        const shortcutDate = new Date(now.getFullYear(), now.getMonth(), num);
+        const daysInMonth = shortcutDate.getDate();
+        if (num > 0 && num <= daysInMonth) {
+          dateValue = shortcutDate;
+        } else {
+          dateValue = value;
+        }
+      } else {
```
Blackbaud-SteveBrush: If we could bake this logic into a self-describing private method, that would help for readability.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
Blackbaud-TrevorBurch: Good catch. Switched this to `unkown`.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
Blackbaud-TrevorBurch: I changed this to `convertedDate` and added a comment.

We can't assume it is always a `Date` anymore as we don't always convert it if it is a string that is not a valid date or one that is numeric but outside the current month's number of days.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
Blackbaud-TrevorBurch: I refactored code to try to make the flow easier to understand and have it be more concise.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -353,16 +353,57 @@ export class SkyDatepickerInputDirective
...
       return;
     }
 
-    const dateValue = this.getDateValue(value);
-    const isDateValid = dateValue && this.dateFormatter.dateIsValid(dateValue);
+    if (value instanceof Date) {
+      const isDateValid = this.dateFormatter.dateIsValid(value);
+
+      /* sanity check */
+      /* istanbul ignore if */
```
Blackbaud-SteveBrush: We're ignoring the `if` statement for coverage?

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -440,22 +461,46 @@ export class SkyDatepickerInputDirective
     this.renderer.setProperty(this.elementRef.nativeElement, 'value', value);
   }
 
-  private getDateValue(value: any): Date {
-    let dateValue: Date;
+  /**
+   * Gets the date value from a value - if possible. Will not convert unconvertable dates or numbers outside of the current month's number of days
+   */
+  private getDateValue(value: unknown): unknown {
```
Blackbaud-SteveBrush: The name of the method implies that I'm getting a `Date` value. What if we made the return type `Date | null` so the value is more explicit and easier to work with. The logic you wrote below will still work with that return type as well.
```
private getDateValue(value: unknown): Date | null {}

private getDateFromShortcut(value: string): Date | null {}

private updateValue(value: unknown, emitEvent = true): void {
  const dateValue = this.getDateValue(value);

  // Your existing `if` checks, here.
}
```

Blackbaud-SteveBrush (commented):
If we could start the commit message with a verb, that would be ideal. Something to the effect of, "convert a user-entered digit into a date for the datepicker component"?

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
Blackbaud-TrevorBurch: Good call - I did that early on. I've added tests to cover this and removed the ignore statement here.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
Blackbaud-TrevorBurch: I've refactored things just a tad to now allow for just `Date | undefined` as the return type from these methods. And then the existing logic handles if `undefined` is returned. Sorry that I had overcomplicated things - it took me a bit after you commented this to even be able to step back and see this path of logic.

Blackbaud-SteveBrush (approved):

PR #181, opened 2022-04-22
fix(components/ag-grid): respect value of deprecated `frameworkComponents`

Blackbaud-SteveBrush:


libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid.service.ts
```diff
@@ -213,17 +212,13 @@ export class SkyAgGridService implements OnDestroy {
     if (mergedGridOptions.stopEditingWhenGridLosesFocus !== undefined) {
```
johnhwhite: This block can be deleted as AG Grid internally handles this in its `checkForDeprecated` [here](https://github.com/ag-grid/ag-grid/blob/973937824caf3666c648d57725b2239386aaf09a/community-modules/core/src/ts/gridOptionsWrapper.ts#L1730).

johnhwhite (approved):

PR #183, opened 2022-04-25
feat(components/pages): add recently accessed service support

Blackbaud-PaulCrowder:


Blackbaud-SteveBrush (dismissed):

libs/components/router/src/lib/modules/recently-accessed/recently-accessed.service.ts
```diff
@@ -0,0 +1,39 @@
+import { Injectable } from '@angular/core';
+
+import { Observable, of } from 'rxjs';
```
johnhwhite: `of` is not used and should not be here.

johnhwhite (commented):
Looks good. One small thing.

libs/components/router/src/lib/modules/recently-accessed/recently-accessed.service.ts
Blackbaud-SteveBrush: Ah, we didn't turn on `noUnusedLocals` in `5.x.x`.

Blackbaud-PaulCrowder: Good catch! Fixed.

Blackbaud-SteveBrush (approved):

johnhwhite (approved):

PR #185, opened 2022-04-26
fix(components/flyout): revert breaking change to `SkyFlyoutService` injectors

Blackbaud-SteveBrush:
We have several consumers mocking out `SkyFlyoutService` and receive the following error in their unit tests:
```
SkyFlyoutService: Expected 5 arguments, but got 6.
```

### DIFF
--- a/libs/components/flyout/src/lib/modules/flyout/flyout.service.ts
+++ b/libs/components/flyout/src/lib/modules/flyout/flyout.service.ts
@@ -1,8 +1,10 @@
 import {
+  ApplicationRef,
   ComponentRef,
   Injectable,
   NgZone,
   OnDestroy,
+  Optional,
   Type,
 } from '@angular/core';
 import { NavigationStart, Router } from '@angular/router';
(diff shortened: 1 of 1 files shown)


libs/components/flyout/src/lib/modules/flyout/flyout.service.ts
```diff
@@ -41,8 +43,16 @@ export class SkyFlyoutService implements OnDestroy {
...
-    private readonly _ngZone: NgZone
-  ) {}
+    private readonly _ngZone: NgZone,
+    // NOTE: This used to be used for an `applicationRef.tick` which has since been removed.
+    // We can not remove this due to it being a breaking change for those manually constructing the service.
+    @Optional() private readonly _applicationRef?: ApplicationRef
+  ) {
+    /*istanbul ignore if*/
+    /*eslint-disable-next-line no-empty*/
+    if (this._applicationRef) {
```
Blackbaud-SteveBrush, line 53: This was needed due to `noUnusedLocals` being set.

Blackbaud-TrevorBurch (approved):

PR #190, opened 2022-04-26
fix(components/a11y): skip link button shows proper localized strings

Blackbaud-SandhyaRajasabeson:
- injects a11y/skip-links at component module level instead of root level
- adds a11y/skip-links to playground


Blackbaud-TrevorBurch (approved):

Blackbaud-TrevorBurch (changes requested):
@Blackbaud-SandhyaRajasabeson can we title this a little closer to what is actually being fixed.  Your current title describes the change but isn't the easiest to understand what is being fixed for the consumer

`fix(components/a11y): skip link buttons show proper localized strings`

Blackbaud-TrevorBurch (approved):

PR #201, opened 2022-04-28
feat(components/validation): create v2 ruleset for URL validation

bradenbiz:


libs/components/validation/package.json
```diff
@@ -21,6 +21,7 @@
     "@angular/forms": "^12.2.16"
   },
   "dependencies": {
-    "tslib": "^2.3.1"
+    "tslib": "^2.3.1",
+    "validator": "^13.7.0"
```
Blackbaud-SteveBrush: Except for `tslib`, all dependencies should be hardened: `"13.7.0"`.

libs/components/validation/src/lib/modules/validation/validation.ts
```diff
@@ -8,8 +12,39 @@ export abstract class SkyValidation {
...
+        allow_underscores: false,
+        host_whitelist: false,
+        host_blacklist: false,
+        allow_trailing_dot: false,
+        allow_protocol_relative_urls: false,
+        allow_fragments: true,
+        allow_query_components: true,
+        disallow_auth: false,
+        validate_length: true,
+      };
```
Blackbaud-SteveBrush: If these are the defaults, we can just leave them out of our implementation (and overwrite them as needed, in the future).

libs/components/validation/src/lib/modules/url-validation/models/url-validation-options.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Specifies options for the URL validator component.
+ */
+export interface SkyURLValidationOptions {
```
Blackbaud-SteveBrush: I'd follow strict pascal casing for the name: `SkyUrlValidationOptions`.

libs/components/validation/src/lib/modules/url-validation/models/url-validation-options.ts
```diff
@@ -0,0 +1,9 @@
...
+  /**
+   * Indicates which ruleset version of URL validation to use.
+   */
+  rulesetVersion: 1 | 2;
```
Blackbaud-SteveBrush: I like the name "ruleset". 👍 

What if we made the ruleset a string union to make it more human-readable/self-explanatory to the consumer:
```
export type SkyValidationRuleSet = 'legacy' | 'common';
```
```
export interface SkyUrlValidationOptions {
  ruleset: SkyValidationRuleSet;
}
```

libs/components/validation/src/lib/modules/url-validation/url-validation.directive.ts
```diff
@@ -21,6 +23,16 @@ const SKY_URL_VALIDATION_VALIDATOR = {
   providers: [SKY_URL_VALIDATION_VALIDATOR],
 })
 export class SkyUrlValidationDirective implements Validator {
+  /**
+   * Specifies configuration options for the URL validation component.
+   */
+  @Input()
+  public set skyUrlValidation(value: SkyURLValidationOptions) {
```
Blackbaud-SteveBrush: I believe for strict mode this input needs to also have an `undefined` value. Try enabling strict mode (in the code-examples SPA) with the following setups to see what happens (they should all  result in a successful build).
```
<input type="text" formControlName="myControl" skyUrlValidation />
<input type="text" formControlName="myControl" [skyUrlValidation] />
<input type="text" formControlName="myControl" [skyUrlValidation]="" />
<input type="text" formControlName="myControl" [skyUrlValidation]="{}" />
<input type="text" formControlName="myControl" [skyUrlValidation]="{ ruleset: 'common' }" />
```

libs/components/validation/src/lib/modules/url-validation/url-validation.directive.ts
```diff
@@ -21,6 +23,16 @@ const SKY_URL_VALIDATION_VALIDATOR = {
...
+    this.skyUrlValidationOptions = value;
+  }
+
+  private skyUrlValidationOptions: SkyURLValidationOptions;
```
Blackbaud-SteveBrush: Properties that are handled by getters/setters should be prefixed with `_`. This is more or less a reminder to maintainers that the property should only be set by the setter, and gotten by the getter.

libs/components/validation/src/lib/modules/url-validation/url-validation.directive.ts
```diff
@@ -38,6 +50,10 @@ export class SkyUrlValidationDirective implements Validator {
   }
 
   public urlIsValid(url: string): boolean {
-    return SkyValidation.isUrl(url);
+    if (this.skyUrlValidationOptions) {
+      return SkyValidation.isUrl(url, this.skyUrlValidationOptions);
+    } else {
+      return SkyValidation.isUrl(url);
```
Blackbaud-SteveBrush: Since the second parameter is optional, can we remove the `if` blocks and just do the following?
```
return SkyValidation.isUrl(url, this.skyUrlValidationOptions);
```

libs/components/validation/src/lib/modules/url-validation/url-validation.directive.ts
Blackbaud-SteveBrush: This input can be undefined, so we should assign its type to: `SkyUrlValidationOptions | undefined`.

libs/components/validation/src/lib/modules/validation/validation.ts
```diff
@@ -8,8 +12,39 @@ export abstract class SkyValidation {
     return regex.test(emailAddress);
   }
 
-  public static isUrl(url: string): boolean {
-    const regex = /^((http|https):\/\/)?([\w-]+\.)+[\w-]+/i;
-    return regex.test(url);
+  public static isUrl(
+    url: string,
+    urlValidationOptions?: SkyURLValidationOptions
```
Blackbaud-SteveBrush: Parameter name can be simplified to `options`.

libs/components/validation/src/lib/modules/validation/validation.ts
```diff
@@ -8,8 +12,39 @@ export abstract class SkyValidation {
...
+  ): boolean {
+    console.log(
+      'skyUrlValidationOptions in Validation.ts function',
```
Blackbaud-SteveBrush: Accidentally left the `console.log` in here?

libs/components/validation/src/lib/modules/validation/validation.ts
```diff
@@ -8,8 +12,39 @@ export abstract class SkyValidation {
     return regex.test(emailAddress);
   }
 
-  public static isUrl(url: string): boolean {
-    const regex = /^((http|https):\/\/)?([\w-]+\.)+[\w-]+/i;
-    return regex.test(url);
+  public static isUrl(
```
Blackbaud-SteveBrush, line 15: Since the core validation logic happens in the `SkyValidation` utility, I would put a few unit tests in the `validation.spec.ts` file to confirm it works in isolation (maybe some of the strings you were given in the ADO story?).

libs/components/validation/src/lib/modules/validation/validation.ts
```diff
@@ -8,8 +12,39 @@ export abstract class SkyValidation {
...
+        host_whitelist: false,
+        host_blacklist: false,
+        allow_trailing_dot: false,
+        allow_protocol_relative_urls: false,
+        allow_fragments: true,
+        allow_query_components: true,
+        disallow_auth: false,
+        validate_length: true,
+      };
+      return isURL(url, ourOptions);
```
Blackbaud-SteveBrush: I got a TypeScript error here after I installed the typings for `validator` (which we should probably do):
```
npm i -ED @types/validator
```

package.json
```diff
@@ -62,6 +62,7 @@
     "normalize-scss": "7.0.1",
     "rxjs": "6.6.7",
     "tslib": "2.3.1",
+    "validator": "13.7.0",
     "zone.js": "0.11.4"
   },
   "devDependencies": {
```
Blackbaud-SteveBrush, line 68: I would install the types for `validator`, `npm i -ED @types/validator`.

libs/components/validation/src/lib/modules/validation/validation.ts
Blackbaud-SteveBrush, line 15: There's also a `url` method on the `SkyValidators` class we should allow options for:
https://github.com/blackbaud/skyux/blob/5.x.x/libs/components/validation/src/lib/modules/validators/validators.ts#L35. 

I'm unsure how you'd add arguments to a custom `FormControl` validator, but I'm sure there's a tutorial out there for it. 😎

libs/components/validation/src/lib/modules/url-validation/models/url-validation-options.ts
Blackbaud-SteveBrush: Also, I noticed you put this in a "models" directory. I'm not sure an options interface would be classified as a model since it doesn't represent a system, such as the shape of an API's response, or the representation of a template's view. We've used a "types" folder historically, but even that isn't perfect since everything in TypeScript is a "type". You're probably safe to just include this file next to the other ones in `/url-validation`.

libs/components/validation/src/lib/modules/validation/validation.ts
```diff
@@ -8,8 +12,39 @@ export abstract class SkyValidation {
...
-    return regex.test(url);
+  public static isUrl(
+    url: string,
+    urlValidationOptions?: SkyURLValidationOptions
+  ): boolean {
+    console.log(
+      'skyUrlValidationOptions in Validation.ts function',
+      urlValidationOptions
+    );
+    if (urlValidationOptions && urlValidationOptions.rulesetVersion === 2) {
```
Blackbaud-SteveBrush: I know we only have two rulesets, but a switch statement would make it easy to add new options in the future.

libs/components/validation/src/lib/modules/url-validation/url-validation.directive.ts
Blackbaud-SteveBrush: Also, come to think of it, does this `Input` need to be a getter/setter? If it does, we should also create a getter.

Blackbaud-SteveBrush (commented):
- The PR title uses `fix`, but I would consider this a `feat`.
- It would be nice if this PR included routes to the code examples so they could be easily interacted with for the review.

libs/components/validation/package.json
bradenbiz: I thought dependencies should be hardened only in the root `package.json`?

libs/components/validation/src/lib/modules/url-validation/models/url-validation-options.ts
bradenbiz: fixed in commit: [be3d5d9](https://github.com/blackbaud/skyux/pull/201/commits/be3d5d99f24810abd668e0b63bfb90a30332c01d)

bradenbiz: fixed in commit: [1f1b312](https://github.com/blackbaud/skyux/pull/201/commits/1f1b312163e69c8899c4a8899529167b22a85a51)

bradenbiz:
> * It would be nice if this PR included routes to the code examples so they could be easily interacted with for the review.

From the docs? Not sure what you mean by this.

libs/components/validation/src/lib/modules/url-validation/models/url-validation-options.ts
bradenbiz: I talked with @Blackbaud-PaulCrowder about this.

He brought up a good point that as we refine this more and more, there is a potentially unlimited set of future rulesets, which lends itself better to versioning than naming. Ie. `export type SkyValidationRuleSet = 'legacy' | 'common' | 'strict' | 'stricter' | 'even-stricter'`

bradenbiz: And thanks, "ruleset" was Paul's idea. 😄

libs/components/validation/package.json
Blackbaud-SteveBrush: Nope, `dependencies` are always hardened (except for Angular's `tslib`). Peer dependencies have ranges, however.

Blackbaud-SteveBrush:
It'd be nice if I could run `nx serve code-examples` and manually interact with the validation code examples.

libs/components/validation/src/lib/modules/url-validation/url-validation.directive.ts
bradenbiz: On making this change and testing it out, there aren't any strict mode errors with the input, but there are a few unrelated to it. Do those need to be fixed in this PR?

libs/components/validation/src/lib/modules/validation/validation.ts
bradenbiz: fixed in commit: [ac166e3](https://github.com/blackbaud/skyux/pull/201/commits/ac166e3bf088260a6d436796ee5e6719e96b63a5)

libs/components/validation/src/lib/modules/url-validation/url-validation.directive.ts
bradenbiz: regardless, fixed the input type in commit: [f437c4f](https://github.com/blackbaud/skyux/pull/201/commits/f437c4f2f03552a0aadd2c2d3dcf5a09e4981aea)

libs/components/validation/package.json
bradenbiz: fixed in commit: [fdb1fcd](https://github.com/blackbaud/skyux/pull/201/commits/fdb1fcdee69d2b7e43f6939a6b0d2da8d7db5f13)

libs/components/validation/src/lib/modules/url-validation/url-validation.directive.ts
bradenbiz: Good call, it does not need a getter & setter. I'll remove it.

bradenbiz: fixed in commit: [0d6947d](https://github.com/blackbaud/skyux/pull/201/commits/0d6947d42e61b53d56d09a70ceb5019924a3cfc2)

libs/components/validation/src/lib/modules/url-validation/url-validation.directive.ts
bradenbiz: Good call. I made it even shorter by removing the helper function completely and just doing that inline. Fixed in commit: [3d3ec05](https://github.com/blackbaud/skyux/pull/201/commits/3d3ec05fb4ed4b2e11730655947aac9e0e4c926b)

libs/components/validation/src/lib/modules/url-validation/url-validation.directive.ts
bradenbiz: fixed in commit: [0d6947d](https://github.com/blackbaud/skyux/pull/201/commits/0d6947d42e61b53d56d09a70ceb5019924a3cfc2)

libs/components/validation/src/lib/modules/validation/validation.ts
bradenbiz: fixed in commit: [remove console.log](https://github.com/blackbaud/skyux/pull/201/commits/0e2d671b33715959842d0b569fe34661dce30042)

libs/components/validation/src/lib/modules/validation/validation.ts
bradenbiz: fixed in commit: [a54e032 - simply parameter name](https://github.com/blackbaud/skyux/pull/201/commits/a54e032ec6b15616ce3ad0adc6a71000f0cdfe52)

libs/components/validation/src/lib/modules/validation/validation.ts
bradenbiz: Due to changes I've made elsewhere (see [remove redundant default options](https://github.com/blackbaud/skyux/pull/201/commits/ac166e3bf088260a6d436796ee5e6719e96b63a5)), I believe this TypeScript issue is fixed. Can you confirm?

package.json
bradenbiz: fixed in commit: [add validator types](https://github.com/blackbaud/skyux/pull/201/commits/c84c9450d0de3c0b85863151a909aba2e6ddd538)

bradenbiz:
> It'd be nice if I could run `nx serve code-examples` and manually interact with the validation code examples.

I did not add any new code-examples for this change. Shall I do so?

libs/components/validation/src/lib/modules/validation/validation.ts
bradenbiz: fixed in commit: [use switch statement rather than if. add new tests for 100% code cove…](https://github.com/blackbaud/skyux/pull/201/commits/9298cb3ca626ef78b73e70c79e322fa54abe8d49)

libs/components/validation/src/lib/modules/validation/validation.ts
bradenbiz: completed in commit: [overload url function so parameters are optional](https://github.com/blackbaud/skyux/pull/201/commits/6d3a0dc7af8801aea103cd746a81b9f69f6cb389)

bradenbiz: fixed in commit: [add unit tests for specific string cases](https://github.com/blackbaud/skyux/pull/201/commits/08c13be9952348c8a13ee8b5468059b5c55b4b9f)

libs/components/validation/src/lib/modules/url-validation/url-validation-options.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Specifies options for the URL validator component.
+ */
+export interface SkyUrlValidationOptions {
+  /**
+   * Indicates which ruleset version of URL validation to use.
+   */
+  rulesetVersion: 1 | 2;
```
Blackbaud-SteveBrush, line 8: I built these changes and put them in our docs site. Here's the result:

Screen Shot 2022-05-12 at 9 38 07 AM

We talked offline: we're going to add a few sentences to the JSDoc comments describing each ruleset version.

apps/code-examples/src/app/code-examples/validation/url-validation/control-validator/url-validation-demo.component.html
```diff
@@ -24,3 +26,13 @@
     </div>
   </ng-container>
 </form>
+
+<button
```
Blackbaud-SteveBrush: We talked about this offline, but for posterity, we're going to move this toggle to the "playground" SPA and just demonstrate using the version 2 ruleset in the code example.

libs/components/validation/src/lib/modules/validation/validation.ts
```diff
@@ -8,8 +12,19 @@ export abstract class SkyValidation {
     return regex.test(emailAddress);
   }
 
-  public static isUrl(url: string): boolean {
+  public static isUrl(url: string, options?: SkyUrlValidationOptions): boolean {
     const regex = /^((http|https):\/\/)?([\w-]+\.)+[\w-]+/i;
-    return regex.test(url);
+    url = url.toString();
```
Blackbaud-SteveBrush: What's the reasoning behind calling `toString` when the parameter is already of type `string`?

libs/components/validation/src/lib/modules/url-validation/url-validation.directive.spec.ts
```diff
@@ -107,3 +117,257 @@ describe('Url validation', () => {
...
+    const inputEvent = document.createEvent('Event');
+    const params = {
+      bubbles: false,
+      cancelable: false,
+    };
+
+    inputEvent.initEvent('input', params.bubbles, params.cancelable);
+
+    const changeEvent = document.createEvent('Event');
+    changeEvent.initEvent('change', params.bubbles, params.cancelable);
```
Blackbaud-SteveBrush: Anywhere you're triggering custom events, you can use the `SkyAppTestUtility` to fire those events for you:
```
import { SkyAppTestUtility } from '@skyux-sdk/testing';

SkyAppTestUtility.fireDomEvent(inputEl, 'change');
```

https://github.com/blackbaud/skyux/blob/main/libs/sdk/testing/src/lib/test-utility/test-utility.ts#L20

libs/components/validation/src/lib/modules/url-validation/url-validation.directive.ts
```diff
@@ -21,23 +28,33 @@ const SKY_URL_VALIDATION_VALIDATOR = {
...
+    this._skyUrlValidationOptions = value;
+    this._validatorChange();
+  }
+
+  private _skyUrlValidationOptions: SkyUrlValidationOptions | undefined;
+
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  private _validatorChange = () => {};
+
+  public validate(control: AbstractControl): ValidationErrors | undefined {
```
Blackbaud-SteveBrush: The `Validator` interface's `validate` method returns `ValidationErrors | null`. Can we do that here, too? This will prepare us appropriately for when we eventually activate "strict" mode.

libs/components/validation/src/lib/modules/validation/validation.spec.ts
```diff
@@ -19,14 +21,83 @@ describe('Validation', () => {
...
+        'sub.domain.ly/here#anchor',
+        'sub.domain#anchor',
+        'sub.domain#anchor?query=here&here=now',
+        "domain.ly?word&amp;$,.%25+'/\\",
+        'sub.domain.ly//',
+      ];
+      strings.forEach((string) => {
+        const isUrlValid = SkyValidation.isUrl(string, optionsRulesetV2);
+        expect(isUrlValid)
+          .withContext('expected "' + string + '" to be valid')
```
Blackbaud-SteveBrush, line 58: Beautiful.

libs/components/validation/src/lib/modules/validators/validators.ts
```diff
@@ -32,15 +33,53 @@ export class SkyValidators {
    * immediately when users enter values.
    * @param control
    */
-  public static url(control: AbstractControl): ValidationErrors | null {
-    const value = control.value;
+  public static url(
+    abstractControl: AbstractControl
+  ): ValidationErrors | undefined;
```
Blackbaud-SteveBrush: Why the switch from `ValidationErrors | null`?

libs/components/validation/src/lib/modules/validation/validation.ts
```diff
@@ -8,8 +12,19 @@ export abstract class SkyValidation {
...
+    if (options) {
+      switch (options.rulesetVersion) {
+        case 1:
+          return regex.test(url);
+        case 2:
+          // we are using Validator's default options
```
Blackbaud-SteveBrush: `Validator` is a type of Angular, maybe say, 
```
we are using the `validator` package's default options
```

libs/components/validation/src/lib/modules/validators/validators.ts
Blackbaud-SteveBrush: Since this is a public method, we'll need to keep the return type the same. If we absolutely need to include `undefined` as a return type, you could do `: ValidationErrors | null | undefined`, but we probably want to keep it as `null` if possible to avoid confusion.

libs/components/validation/src/lib/modules/url-validation/url-validation-options.ts
bradenbiz: resolved in: [describe each ruleset in JSDocs](https://github.com/blackbaud/skyux/pull/201/commits/be5746df7b6dd69968b0109872fae81f6997bdda)

libs/components/validation/src/lib/modules/url-validation/url-validation.directive.spec.ts
bradenbiz: resolved in: [use SkyAppTestUtility to fire custom events](https://github.com/blackbaud/skyux/pull/201/commits/03bc316fdcd54bd8eb5f855b1385384f889ddec8)

libs/components/validation/src/lib/modules/validation/validation.ts
bradenbiz: resolved in: [update language/syntax](https://github.com/blackbaud/skyux/pull/201/commits/08014f8affe8beb1b116dd9ddc7d1c5f13080141)

libs/components/validation/src/lib/modules/validation/validation.ts
bradenbiz: I was getting a compile error in some cases claiming that the `.test(_)` function only accepts `string`s, and adding that line fixed that error for me.

apps/code-examples/src/app/code-examples/validation/url-validation/control-validator/url-validation-demo.component.html
bradenbiz: resolved in: [add URL Validation toggeable example to Playground](https://github.com/blackbaud/skyux/pull/201/commits/c9715fe8700fb2abe1dcbe9dd818bddf91d77125)

bradenbiz: and: [remove toggle from code examples & make them v2.](https://github.com/blackbaud/skyux/pull/201/commits/1f0b02fdc964592d1d03c429f9c5f5cfc1f0659a)

libs/components/validation/src/lib/modules/url-validation/url-validation.directive.ts
bradenbiz: resolved in: [use null instead of undefined.](https://github.com/blackbaud/skyux/pull/201/commits/f89f707df545fe8518df07cadd60b48fe6ac5a90)

libs/components/validation/src/lib/modules/validators/validators.ts
bradenbiz: resolved in: [use null instead of undefined.](https://github.com/blackbaud/skyux/pull/201/commits/f89f707df545fe8518df07cadd60b48fe6ac5a90)

apps/code-examples/src/app/code-examples/validation/url-validation/control-validator/url-validation-demo.component.ts
```diff
@@ -21,6 +21,10 @@ export class UrlValidationDemoComponent implements OnInit {
 
   constructor(private formBuilder: FormBuilder) {}
 
+  public skyUrlValidationOptions: SkyUrlValidationOptions = {
+    rulesetVersion: 2,
+  };
```
Blackbaud-SteveBrush: Is this property being used? I think it's meant to be put in the `FormControl`, below?
```
SkyValidators.url({
  rulesetVersion: 2,
})
```

libs/components/validation/src/lib/modules/url-validation/url-validation-options.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Specifies options for the URL validator component.
+ */
+export interface SkyUrlValidationOptions {
+  /**
+   * Indicates which ruleset version of URL validation to use.
+   *
+   * Ruleset v1 (the default) uses a simple, custom regex.
+   *
+   * Ruleset v2 uses the stricter, 3rd party [validator.js library](https://github.com/validatorjs/validator.js/).
```
Blackbaud-SteveBrush: Since this will be added to the docs, I would make it all one paragraph for readability (with a few minor changes):
```
Indicates the ruleset to use for URL validation. Ruleset 1 uses a regular expression and ruleset 2 uses the third-party [validator.js library](https://github.com/validatorjs/validator.js/).
```

libs/components/validation/src/lib/modules/validation/validation.ts
Blackbaud-SteveBrush: I see what's going on. The `SkyUrlValidationDirective` is calling `SkyValidation.isUrl` with the value of its form control. The value of a form control is `any`, so if the consumer sets the value to anything other than a `string`, the `isUrl` check fails (because it's expecting a `string`). 

To make the `isUrl` method more robust (and to better explain the scenario when the `url` value is not a string), I would change the first parameter to a type of `unknown` because it's type really is unknown, coming from a form control.
```
public static isUrl(
  value: unknown,
  options?: SkyUrlValidationOptions
): boolean {
  const regex = /^((http|https):\/\/)?([\w-]+\.)+[\w-]+/i;

  if (typeof value !== 'string') {
    return false;
  }

  const url: string = value;

  if (options) {
...
```

libs/components/validation/src/lib/modules/validators/validators.ts
```diff
@@ -10,37 +11,71 @@ export class SkyValidators {
...
 
-    if (!value) {
-      return undefined;
-    }
+  /**
+   * Validates URLs in reactive forms. Add this validator with ruleset parameters directly to
+   * the form control model in the component class. If users enter values that are not valid
+   * URLs, the validator throws an error. Since this is a sync validator, it returns a set of
+   * validation errors or `null` immediately when users enter values.
+   * @param control
```
Blackbaud-SteveBrush: Ditto, about the `@param`.

libs/components/validation/src/lib/modules/validators/validators.ts
```diff
@@ -10,37 +11,71 @@ export class SkyValidators {
...
       : { skyEmail: { invalid: value } };
   }
 
   /**
    * Validates URLs in reactive forms. Add this validator directly to the form control model in
    * the component class. If users enter values that are not valid URLs, the validator throws an
-   * error. Since this is a sync validator, it returns a set of validation errors or `undefined`
+   * error. Since this is a sync validator, it returns a set of validation errors or `null`
    * immediately when users enter values.
    * @param control
```
Blackbaud-SteveBrush: This `@param` should probably be removed. TypeDoc does a good job of determining the method's parameters when it generates the documentation.

libs/components/validation/src/lib/modules/validation/validation.ts
bradenbiz: resolved in: [make incoming value typed as unknown and then check](https://github.com/blackbaud/skyux/pull/201/commits/b5eac3a25445552abb8040b7e159e84f74067f58)

apps/code-examples/src/app/code-examples/validation/url-validation/control-validator/url-validation-demo.component.ts
bradenbiz: resolved in: [needs to use v2 ACTUALLY in code examples for control validator](https://github.com/blackbaud/skyux/pull/201/commits/94080c1c7d0d594eeb613b7d92272a34a1d5434a)

libs/components/validation/src/lib/modules/url-validation/url-validation-options.ts
bradenbiz: resolved in: [update jsdoc for rulesetVersion](https://github.com/blackbaud/skyux/pull/201/commits/6b6eb2ec5b3c2f5d06dcaa4acc524d9de414abac)

libs/components/validation/src/lib/modules/validators/validators.ts
bradenbiz: resolved in: [remove param from jsdoc comments](https://github.com/blackbaud/skyux/pull/201/commits/445a36e575afc25770140edcc5217ce6a4f36cc6)

libs/components/validation/src/lib/modules/validators/validators.ts
bradenbiz: resolved in: [remove param from jsdoc comments](https://github.com/blackbaud/skyux/pull/201/commits/445a36e575afc25770140edcc5217ce6a4f36cc6)

Blackbaud-SteveBrush (approved):

PR #209, opened 2022-05-02
fix(components/core): add null checks to core adapter focusing methods

Blackbaud-TrevorBurch:


libs/components/core/src/lib/modules/adapter-service/adapter.service.spec.ts
```diff
@@ -27,6 +27,11 @@ describe('Core adapter service', () => {
       expect(actual.length).toEqual(8);
     });
 
+    it('should return an empty array if an undefined value is queries for focusable children', () => {
```
Blackbaud-SteveBrush: Not sure I understand this sentence?

Blackbaud-TrevorBurch: Sigh. That is a typo. Standby

Blackbaud-TrevorBurch: Fixed

Blackbaud-SteveBrush (approved):

PR #213, opened 2022-05-03
refactor(components/popovers): enable strict mode (#205)

Blackbaud-SteveBrush:


libs/components/popovers/testing/src/popover/popover-fixture.ts
```diff
@@ -62,30 +62,30 @@ export class SkyPopoverFixture {
...
 
-  private get bodyElement(): HTMLElement {
-    return this.queryOverlay('.sky-popover-body');
+  private get bodyElement(): HTMLElement | undefined {
+    return this.queryOverlay('.sky-popover-body') || undefined;
   }
 
-  private getOverlay(): HTMLElement {
-    return document.querySelector('sky-overlay');
+  private getOverlay(): Element | undefined {
```
johnhwhite, line 81: It seems like we should keep this consistent as `HTMLElement` or change other `HTMLElement` references to `Element`.

Blackbaud-SteveBrush: This is a cherry-pick commit, so we'll need to fix it in `5.x.x` as well.

In any case, when querying against the `document`, TypeScript returns either `Element | null`, whereas when querying against an `HTMLElement` TypeScript returns `HTMLElement | null`. I figured we should respect what the native API returns instead of casting it as an extended type. (This is also a private method so it's not a breaking change.)

johnhwhite (approved):

PR #220, opened 2022-05-03
docs(apps/code-examples): cleanup CSS class usage (#161)

Blackbaud-SteveBrush:


apps/code-examples/src/app/code-examples/lists/repeater/basic/repeater-demo.component.scss
```diff
@@ -2,7 +2,7 @@
   display: flex;
   flex-wrap: wrap;
 
-  div.sky-emphasized {
+  div.sky-font-emphasized {
```
Blackbaud-PaulCrowder: This is a weird thing to do, overriding a CSS class that just happens to be on an element that needs this style. Can you change this so it has its own CSS class instead of piggybacking off an unrelated SKY UX CSS class?

Blackbaud-PaulCrowder: Looks like this happens in the `add-remove` and `inline-form` examples as well.

Blackbaud-SteveBrush: Sure thing.

Blackbaud-PaulCrowder (approved):

PR #221, opened 2022-05-03
deprecation(components/layout): `SkyCardComponent` is deprecated; use a different container from the content container guidelines instead

Blackbaud-TrevorBurch:


libs/components/layout/src/lib/modules/card/card-content.component.ts
```diff
@@ -1,7 +1,8 @@
 import { Component } from '@angular/core';
 
 /**
- * Specifies the content to display in the body of the card.
+ * Specifies the content to display in the body of the card.\
```
Blackbaud-SteveBrush: Got a rogue backward slash.

Blackbaud-SteveBrush (approved):

PR #232, opened 2022-05-04
fix(components/core): convert `SkyNumericOptions` from a class to an interface

Blackbaud-SteveBrush:


libs/components/core/src/lib/modules/numeric/numeric.pipe.spec.ts
```diff
@@ -18,7 +18,7 @@ describe('Numeric pipe', () => {
       markForCheck: jasmine.createSpy('markForCheck'),
     };
 
-    expectedConfig = new SkyNumericOptions();
+    expectedConfig = new NumericOptions();
```
Blackbaud-PaulCrowder, line 21: We should have a test that only uses the interface to validate that whatever default values set by the `NumericOptions` constructor are carried over to a plain JavaScript object before formatting the number.

Blackbaud-SteveBrush: We have a few existing unit tests that pass in an object (not an instance of `NumericOptions`). Is that what you mean? https://github.com/blackbaud/skyux/blob/1ca682353c08d64df610d93e1b05d3ed600bac60/libs/components/core/src/lib/modules/numeric/numeric.pipe.spec.ts#L39-L45

Blackbaud-SteveBrush: And there are these:
https://github.com/blackbaud/skyux/blob/1ca682353c08d64df610d93e1b05d3ed600bac60/libs/components/core/src/lib/modules/numeric/numeric.pipe.spec.ts#L55-L77

Blackbaud-SteveBrush: Happy to write another one if I'm misunderstanding.

Blackbaud-PaulCrowder: Those should be fine.

Blackbaud-PaulCrowder (approved):

PR #241, opened 2022-05-06
fix(components/lists): repeater a11y improvements for aria role and keyboard interaction

johnhwhite:
[AB#2096738](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2096738)

### DIFF
--- a/libs/components/lists/package.json
+++ b/libs/components/lists/package.json
@@ -16,6 +16,7 @@
   },
   "homepage": "https://github.com/blackbaud/skyux#readme",
   "peerDependencies": {
+    "@angular/cdk": "^12.2.13",
     "@angular/common": "^12.2.16",
     "@angular/core": "^12.2.16",
     "@angular/platform-browser": "^12.2.16",
--- a/libs/components/lists/src/lib/modules/repeater/fixtures/repeater.component.fixture.ts
+++ b/libs/components/lists/src/lib/modules/repeater/fixtures/repeater.component.fixture.ts
@@ -1,9 +1,7 @@
 import { Component, ViewChild } from '@angular/core';
 
-import {
-  SkyRepeaterComponent,
-  SkyRepeaterRoleType,
-} from '../repeater.component';
+import { SkyRepeaterRoleType } from '../repeater-role.type';
+import { SkyRepeaterComponent } from '../repeater.component';
 
 let nextItemId = 0;
 
@@ -19,8 +17,6 @@ export class RepeaterTestComponent {
   public get activeIndex(): number {
     return this._activeIndex;
   }
-
-  public ariaRole: SkyRepeaterRoleType | 'auto' | undefined = 'auto';
 
   public disableFirstItem = false;
 
--- a/libs/components/lists/src/lib/modules/repeater/fixtures/repeater.component.fixture.ts
+++ b/libs/components/lists/src/lib/modules/repeater/fixtures/repeater.component.fixture.ts
@@ -1,6 +1,5 @@
 import { Component, ViewChild } from '@angular/core';
 
-import { SkyRepeaterRoleType } from '../repeater-role.type';
 import { SkyRepeaterComponent } from '../repeater.component';
 
 let nextItemId = 0;
--- a/libs/components/lists/src/lib/modules/repeater/fixtures/repeater.component.fixture.ts
+++ b/libs/components/lists/src/lib/modules/repeater/fixtures/repeater.component.fixture.ts
@@ -1,6 +1,9 @@
 import { Component, ViewChild } from '@angular/core';
 
-import { SkyRepeaterComponent } from '../repeater.component';
+import {
+  SkyRepeaterComponent,
+  SkyRepeaterRoleType,
+} from '../repeater.component';
 
 let nextItemId = 0;
 
@@ -16,6 +19,10 @@ export class RepeaterTestComponent {
   public get activeIndex(): number {
     return this._activeIndex;
   }
+
+  public ariaRole: SkyRepeaterRoleType | 'auto' | undefined = 'auto';
+
+  public disableFirstItem = false;
 
   public expandMode = 'single';
 
--- a/libs/components/lists/src/lib/modules/repeater/repeater-item-roles.type.ts
+++ b/libs/components/lists/src/lib/modules/repeater/repeater-item-roles.type.ts
@@ -1,5 +1,5 @@
 export type SkyRepeaterItemRolesType = {
-  item?: 'listitem' | 'option' | 'row' | undefined;
-  title?: 'rowheader' | undefined;
-  content?: 'gridcell' | undefined;
+  item: 'listitem' | 'option' | 'row' | undefined;
+  title: 'rowheader' | undefined;
+  content: 'gridcell' | undefined;
 };
--- a/libs/components/lists/src/lib/modules/repeater/repeater-item-roles.type.ts
+++ b/libs/components/lists/src/lib/modules/repeater/repeater-item-roles.type.ts
@@ -1,0 +1,5 @@
+export type SkyRepeaterItemRolesType = {
+  item?: 'listitem' | 'option' | 'row' | undefined;
+  title?: 'rowheader' | undefined;
+  content?: 'gridcell' | undefined;
+};
--- a/libs/components/lists/src/lib/modules/repeater/repeater-item.component.html
+++ b/libs/components/lists/src/lib/modules/repeater/repeater-item.component.html
@@ -2,7 +2,7 @@
   class="sky-repeater-item sky-padding-even-default"
   [attr.aria-label]="itemName"
   [attr.aria-selected]="selectable ? isSelected : null"
-  [attr.role]="(this.repeaterService.itemRole | async).item"
+  [attr.role]="(itemRole$ | async).item"
   [ngClass]="{
     'sky-repeater-item-active': isActive,
     'sky-repeater-item-collapsible': isCollapsible,
@@ -97,7 +97,7 @@
       <div
         class="sky-repeater-item-title sky-emphasized"
         #titleRef
-        [attr.role]="(this.repeaterService.itemRole | async).title"
+        [attr.role]="(itemRole$ | async).title"
       >
         <ng-content select="sky-repeater-item-title"></ng-content>
       </div>
@@ -146,7 +146,7 @@
       [id]="contentId"
       [@skyAnimationSlide]="slideDirection"
       #itemContentRef
-      [attr.role]="(this.repeaterService.itemRole | async).content"
+      [attr.role]="(itemRole$ | async).content"
     >
       <ng-content select="sky-repeater-item-content"></ng-content>
     </div>
--- a/libs/components/lists/src/lib/modules/repeater/repeater-item.component.html
+++ b/libs/components/lists/src/lib/modules/repeater/repeater-item.component.html
@@ -1,7 +1,7 @@
 <div
   class="sky-repeater-item sky-padding-even-default"
   [attr.aria-label]="itemName"
-  [attr.aria-selected]="selectable ? isSelected : null"
+  [attr.aria-selected]="selectable ? isSelected : undefined"
   [attr.role]="(itemRole$ | async).item"
   [ngClass]="{
     'sky-repeater-item-active': isActive,
--- a/libs/components/lists/src/lib/modules/repeater/repeater-item.component.html
+++ b/libs/components/lists/src/lib/modules/repeater/repeater-item.component.html
@@ -1,8 +1,8 @@
 <div
   class="sky-repeater-item sky-padding-even-default"
-  role="option"
   [attr.aria-label]="itemName"
   [attr.aria-selected]="selectable ? isSelected : null"
+  [attr.role]="(this.repeaterService.itemRole | async).item"
   [ngClass]="{
     'sky-repeater-item-active': isActive,
     'sky-repeater-item-collapsible': isCollapsible,
@@ -94,7 +94,11 @@
       [hidden]="titleRef.children.length === 0"
       (click)="headerClick()"
     >
-      <div class="sky-repeater-item-title sky-emphasized" #titleRef>
+      <div
+        class="sky-repeater-item-title sky-emphasized"
+        #titleRef
+        [attr.role]="(this.repeaterService.itemRole | async).title"
+      >
         <ng-content select="sky-repeater-item-title"></ng-content>
       </div>
       <button
@@ -142,6 +146,7 @@
       [id]="contentId"
       [@skyAnimationSlide]="slideDirection"
       #itemContentRef
+      [attr.role]="(this.repeaterService.itemRole | async).content"
     >
       <ng-content select="sky-repeater-item-content"></ng-content>
     </div>
--- a/libs/components/lists/src/lib/modules/repeater/repeater-item.component.html
+++ b/libs/components/lists/src/lib/modules/repeater/repeater-item.component.html
@@ -9,7 +9,6 @@
     'sky-repeater-item-selected': isSelected
   }"
   (click)="onRepeaterItemClick($event)"
-  (keydown)="onItemKeyDown($event)"
   #itemRef
 >
   <div #inlineDelete>
--- a/libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
+++ b/libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
@@ -25,24 +25,16 @@ import {
   SkyInlineFormConfig,
 } from '@skyux/inline-form';
 
-import { Subject, forkJoin as observableForkJoin } from 'rxjs';
+import { Observable, Subject, forkJoin as observableForkJoin } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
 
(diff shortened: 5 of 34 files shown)


libs/components/lists/package.json
```diff
@@ -16,6 +16,7 @@
   },
   "homepage": "https://github.com/blackbaud/skyux#readme",
   "peerDependencies": {
+    "@angular/cdk": "^12.2.13",
```
Blackbaud-SteveBrush, line 19: This would normally be a breaking change, but we can add the dependency to consumer projects using this change: https://github.com/blackbaud/skyux/pull/249

libs/components/lists/src/lib/modules/repeater/repeater-item.component.html
```diff
@@ -1,8 +1,8 @@
 <div
   class="sky-repeater-item sky-padding-even-default"
-  role="option"
   [attr.aria-label]="itemName"
   [attr.aria-selected]="selectable ? isSelected : null"
+  [attr.role]="(this.repeaterService.itemRole | async).item"
```
Blackbaud-SteveBrush: Is the `this` needed? If not, it should be removed throughout the template.

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
```diff
@@ -34,6 +35,15 @@ import { SkyRepeaterService } from './repeater.service';
 
 let nextContentId = 0;
 
+export type SkyRepeaterItemRoleType = 'listitem' | 'option' | 'row' | undefined;
+export type SkyRepeaterItemTitleRoleType = 'rowheader' | undefined;
+export type SkyRepeaterItemContentRoleType = 'gridcell' | undefined;
+export type SkyRepeaterItemRolesType = {
```
Blackbaud-SteveBrush: If these types truly need to be exported, put them in separate files, otherwise, you can remove the `export`.

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
```diff
@@ -226,12 +273,14 @@ export class SkyRepeaterItemComponent
...
 
+  private _isDisabled = false;
+
   private _isExpanded = true;
 
   private _isSelected = false;
 
   constructor(
-    private repeaterService: SkyRepeaterService,
+    public repeaterService: SkyRepeaterService,
```
Blackbaud-SteveBrush: Instead of making this service--and all of its members--public to the template (and anything else that has access to this component), we should create a new property or method on the component that returns the needed values from the private service. This would also prevent our consumers from using a deep import to reach the component type, and then call methods on the repeater service directly.

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
```diff
@@ -287,6 +336,90 @@ export class SkyRepeaterItemComponent
...
+      const items = this.repeaterService.items
+        .slice(0, currentIndex)
+        .filter((item) => !item.disabled);
+      const activateItem = items.shift();
+      /* istanbul ignore else */
+      if (activateItem) {
+        this.repeaterService.activateItem(activateItem);
+        activateItem.elementRef.nativeElement.focus();
+      }
+    }
```
Blackbaud-SteveBrush, line 413: There's a bunch of repeated code in these `keydown` callbacks; it might be more appropriate to have a single callback with a `switch` statement to handle the `key` value.

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
```diff
@@ -20,11 +20,16 @@ import { Subject } from 'rxjs';
...
-import { SkyRepeaterItemComponent } from './repeater-item.component';
+import {
+  SkyRepeaterItemComponent,
+  SkyRepeaterItemRolesType,
+} from './repeater-item.component';
 import { SkyRepeaterService } from './repeater.service';
 
 let uniqueId = 0;
 
+export type SkyRepeaterRoleType = 'list' | 'listbox' | 'grid';
```
Blackbaud-SteveBrush: If this type needs to be exported, let's put it in a separate file.

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
```diff
@@ -56,6 +61,19 @@ export class SkyRepeaterComponent
...
 
+  /**
+   * Specifies an ARIA role for the repeater list.
+   * This sets the repeater list's `role` attribute and determines the appropriate role for repeater items
+   * [to support accessibility](https://developer.blackbaud.com/skyux/learn/accessibility).
+   * The default `auto` value will determine one of `list`, `listbox`, or `grid`.
+   * @default "auto"
+   */
+  @Input()
+  public set ariaRole(value: SkyRepeaterRoleType | 'auto' | undefined) {
```
Blackbaud-SteveBrush: Does this need to be a public input since the role is auto-determined by the features currently active on the repeater? What scenario would we want the consumer to override the role attribute themselves?

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
```diff
@@ -105,19 +123,23 @@ export class SkyRepeaterComponent
    * This event emits an ordered array of the `tag` properties that the consumer provides for each repeater item.
    */
   @Output()
-  public orderChange = new EventEmitter<any[]>();
+  public orderChange = new EventEmitter<unknown[]>();
```
Blackbaud-SteveBrush: This is a breaking change since it would require the consumer to cast the output's argument as a specific type. (Today it returns `any`, but after this change it would return `unknown`.)

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
```diff
@@ -105,19 +123,23 @@ export class SkyRepeaterComponent
...
 
+  public role: 'list' | 'listbox' | 'grid' | '' = '';
+
   private dragulaUnsubscribe = new Subject<void>();
 
   private ngUnsubscribe = new Subject<void>();
 
   private _expandMode = 'none';
 
+  private _ariaRole: 'auto' | 'list' | 'listbox' | 'grid' | '' = 'auto';
```
Blackbaud-SteveBrush: I was initially having difficulty discerning the differences between `_ariaRole` and `role`. What if you had the `ariaRole` setter require a specific type for its `value` parameter, and an `ariaRole` getter that returns a more specific type to be used by the template?

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
```diff
@@ -105,19 +123,23 @@ export class SkyRepeaterComponent
...
   @Output()
-  public orderChange = new EventEmitter<any[]>();
+  public orderChange = new EventEmitter<unknown[]>();
 
   @ContentChildren(SkyRepeaterItemComponent)
   public items: QueryList<SkyRepeaterItemComponent>;
 
   public dragulaGroupName: string;
 
+  public role: 'list' | 'listbox' | 'grid' | '' = '';
```
Blackbaud-SteveBrush: Is it appropriate to set a role as `''`? Would it be better to set it as `null` or `undefined`?

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
```diff
@@ -232,6 +257,10 @@ export class SkyRepeaterComponent
     this.destroyDragAndDrop();
   }
 
+  public checkContent() {
```
Blackbaud-SteveBrush: Needs a return type.

Also, it's not immediately clear what this method is used for (or who calls it). What do you think about renaming it to `onCdkObserveContent` to match our other template event handlers (e.g. `onClick`, `onKeyDown`)?

libs/components/lists/package.json
johnhwhite: Thanks.

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
johnhwhite: Good catch.

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
johnhwhite: Consolidated the keydown handlers.

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
johnhwhite: Removed the input as we discussed.

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
johnhwhite: Reverted.

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
johnhwhite: This is no longer needed without the input.

libs/components/lists/src/lib/modules/repeater/fixtures/repeater.component.fixture.ts
```diff
@@ -1,5 +1,6 @@
 import { Component, ViewChild } from '@angular/core';
 
+import { SkyRepeaterRoleType } from '../repeater-role.type';
```
Blackbaud-SteveBrush: Needed?

libs/components/lists/src/lib/modules/repeater/repeater-item-roles.type.ts
```diff
@@ -0,0 +1,5 @@
+export type SkyRepeaterItemRolesType = {
+  item?: 'listitem' | 'option' | 'row' | undefined;
```
Blackbaud-SteveBrush: Including `undefined` in the union type is redundant when using the `?` character. 

If you want to keep the `undefined` type in the union, you could remove the `?` character, especially if you want the consumer of the type to deliberately set each property (even if that deliberate value is `undefined`).

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
```diff
@@ -47,6 +49,43 @@ let nextContentId = 0;
...
+      if (value) {
+        this.isSelected = false;
+        this._isDisabled = true;
+      } else {
+        this._isDisabled = false;
+      }
+      if (this.isActive) {
+        this.repeaterService.activateItemByIndex(undefined);
+      }
+      /* istanbul ignore next */
```
Blackbaud-SteveBrush: Curious why this isn't testable?

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
```diff
@@ -91,7 +130,7 @@ export class SkyRepeaterItemComponent
    */
   @Input()
   public set isSelected(value: boolean) {
-    if (value !== this._isSelected) {
+    if (!this._isDisabled && value !== this._isSelected) {
```
Blackbaud-SteveBrush: We should use the getter when reading the value: `!this.disabled`.

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
```diff
@@ -112,6 +115,8 @@ export class SkyRepeaterComponent
 
   public dragulaGroupName: string;
 
+  public role: 'list' | 'listbox' | 'grid' | undefined = undefined;
```
Blackbaud-SteveBrush: I believe this should be `public role: SkyRepeaterRoleType | undefined`. A value of `undefined` is assumed.

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
```diff
@@ -333,4 +349,40 @@ export class SkyRepeaterComponent
...
+      .concat([`.sky-repeater[role="grid"]`, `skyux-dropdown`])
+      .join(', ');
+    const hasInteraction =
+      this.reorderable ||
+      !!(this.elementRef.nativeElement as HTMLElement).querySelector(
+        interactionSelector
+      );
+    if (hasInteraction) {
+      autoRole = 'grid';
+    } else if (hasSelectables) {
```
Blackbaud-SteveBrush: Should we check `hasSelectables` first? If it's `true`, then we don't have to waste resources querying the DOM to see if there are interactable elements in the repeater item.
```
if (hasSelectables) {
  // ...
} else {
  const hasInteraction = ...
  if (hasInteraction) { ... }
}
```

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
```diff
@@ -333,4 +349,40 @@ export class SkyRepeaterComponent
...
+    };
+    const hasSelectables = this.items?.some((item) => item.selectable);
+    const interactionSelector = [
+      'a[href]',
+      'button',
+      'select',
+      'textarea',
+      'input',
+    ]
+      .map((selector) => `sky-repeater-item-right ${selector}`)
```
Blackbaud-SteveBrush: Is `sky-repeater-item-right` a component element? I'm not seeing it anywhere. Should this be `.sky-repeater-item-right`?

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
```diff
@@ -333,4 +349,40 @@ export class SkyRepeaterComponent
...
+      grid: { item: 'row', title: 'rowheader', content: 'gridcell' },
+    };
+    const hasSelectables = this.items?.some((item) => item.selectable);
+    const interactionSelector = [
+      'a[href]',
+      'button',
+      'select',
+      'textarea',
+      'input',
+    ]
```
Blackbaud-SteveBrush, line 388: ☝️ Curious, how did we decide which selectors to include in this array?

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
```diff
@@ -47,6 +49,43 @@ let nextContentId = 0;
 export class SkyRepeaterItemComponent
   implements OnDestroy, OnInit, AfterViewInit
 {
+  /**
+   * The first, non-disabled item is tab focusable.
+   */
+  @HostBinding()
+  public get tabindex(): 0 | -1 {
+    return this.repeaterService.items.filter((item) => !item.disabled)[0] ===
```
Blackbaud-SteveBrush, line 59: So only the first, non-disabled item should be focusable? The other items are not allowed to be focusable? Just trying to understand the a11y rules behind this. If there's any information about this requirement on W3C (or somewhere similar) we should put a link in this property's JSDocs comment.

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
```diff
@@ -287,6 +332,90 @@ export class SkyRepeaterItemComponent
...
+      $event.target &&
+      ($event.target as Element).matches(
+        [
+          'sky-repeater-item',
+          '.sky-repeater-item',
+          '.sky-repeater-item-left',
+          '.sky-repeater-item-left *',
+          '.sky-repeater-item-title',
+          '.sky-repeater-item-title *',
+        ].join(',')
```
Blackbaud-SteveBrush: Is this array required to avoid capturing events from the `.sky-repeater-item-right` DIV? If so, would it be easier to check that the event does not come from one element, versus checking if the event exists in any number of other elements?

Either way, it would be helpful to include a comment explaining the array's purpose.

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
johnhwhite: Added a test.

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
johnhwhite: I added a comment for this. The `hasInteraction` takes precedence if there are also selectable items, and I added additional logic to look at component interaction settings first (negate the need to query the DOM) and to tighten the DOM selector.

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
johnhwhite: Fixed.

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
johnhwhite: I found a source and added a comment.

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
johnhwhite: Added comments.

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
johnhwhite: The idea was to prevent e.g. someone pressing down arrow in an input and that causing focus to change to the next item. I've shifted the code so that might be more obvious.

Blackbaud-SteveBrush (approved):

PR #244, opened 2022-05-09
fix(components/ag-grid): data manager not persisting column order

johnhwhite:
[AB#2143855](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2143855)


libs/components/data-manager/src/lib/modules/data-manager/data-manager-column-picker/data-manager-column-picker.component.ts
```diff
@@ -76,11 +76,6 @@ export class SkyDataManagerColumnPickerComponent implements OnDestroy, OnInit {
   ) {}
 
   public ngOnInit(): void {
-    this.libResources
```
Blackbaud-SteveBrush, line 79: We should remove the `private libResources: SkyLibResourcesService` argument from the constructor, too? If not, this will cause an error in the `main` branch due to `noUnusedLocals` being set.

Blackbaud-SteveBrush (approved):

PR #247, opened 2022-05-10
fix(sdk/prettier-schematics): include `.angular/cache` and `src/assets` in Prettier's ignore file

Blackbaud-TrevorBurch:


libs/sdk/prettier-schematics/src/schematics/ng-add/ng-add.schematic.spec.ts
```diff
@@ -130,6 +130,12 @@ dist
 node_modules
 package-lock.json
 
+# Ignore: assets
+src/assets
```
Blackbaud-SteveBrush: What about assets for libraries or applications that live in the `projects` directory?

libs/sdk/prettier-schematics/src/schematics/ng-add/ng-add.schematic.spec.ts
```diff
@@ -130,6 +130,16 @@ dist
 node_modules
 package-lock.json
 
+# Ignore assets
+/src/assets
+/projects/*/src/assets/
```
Blackbaud-SteveBrush, line 135: You have a trailing slash here, but not on `/src/assets`. Does that produce a different effect?

Blackbaud-SteveBrush: From `.gitignore`'s docs: "If there is a separator at the end of the pattern then the pattern will only match directories, otherwise the pattern can match both files and directories."
https://git-scm.com/docs/gitignore#_pattern_format
Maybe we should put a trailing slash on the first one, too, then.

Blackbaud-SteveBrush (approved):

PR #248, opened 2022-05-10
fix(components/packages): add `src/assets` to existing Prettier ignore files

Blackbaud-TrevorBurch:


libs/components/packages/src/schematics/migrations/prettier-ignore-assets/prettier-ignore-assets.schematic.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export default function addAssetsToPrettierIgnore(): Rule {
+  return (tree) => {
+    const assetsPath = 'src/assets';
+    const prettierIgnorePath = '/.prettierignore';
+    const prettierIgnore = tree.read(prettierIgnorePath);
+    const assetsRegEx = new RegExp('^[\\\\/]?src[\\\\/]assets$', 'm');
+
+    console.log(prettierIgnore.toString());
+    console.log(assetsRegEx.exec(prettierIgnore.toString()));
```
Blackbaud-SteveBrush: Console logs.

libs/components/packages/src/schematics/migrations/prettier-ignore-assets/prettier-ignore-assets.schematic.ts
```diff
@@ -0,0 +1,16 @@
+import { Rule } from '@angular-devkit/schematics';
+
+export default function addAssetsToPrettierIgnore(): Rule {
+  return (tree) => {
+    const assetsPath = 'src/assets';
```
Blackbaud-SteveBrush: What about library assets? Or applications in the `projects` directory? Is there an "Angular way" to determine each project's assets directory?

libs/components/packages/src/schematics/migrations/prettier-ignore-assets/prettier-ignore-assets.schematic.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export default function addAssetsToPrettierIgnore(): Rule {
+  return (tree) => {
+    const spaAssetsPath = 'src/assets';
+    const libraryAssetsPath = 'projects/*/src/assets';
+    const prettierIgnorePath = '/.prettierignore';
+    const prettierIgnore = tree.read(prettierIgnorePath);
+    const spaAssetsRegEx = new RegExp('^[\\\\/]?src[\\\\/]assets$', 'm');
+    const libraryAssetsRegEx = new RegExp(
+      '^[\\\\/]?projects[\\\\/]\\*[\\\\/]src[\\\\/]assets$',
```
Blackbaud-SteveBrush: Prettier uses the `.gitignore` syntax, so directories should always use the forward slash. The backward slash actually has special meaning in `.gitignore` syntax. You can keep this simple and expect folks to use the forward slash for directories. See: https://prettier.io/docs/en/ignore.html#ignoring-files-prettierignore

libs/components/packages/src/schematics/migrations/prettier-ignore-assets/prettier-ignore-assets.schematic.ts
```diff
@@ -0,0 +1,29 @@
+import { Rule } from '@angular-devkit/schematics';
+
+export default function addAssetsToPrettierIgnore(): Rule {
+  return (tree) => {
+    const spaAssetsPath = 'src/assets';
+    const libraryAssetsPath = 'projects/*/src/assets';
```
Blackbaud-SteveBrush: It'd be nice to be able to add more paths in the future as needed; maybe an array of ignored paths would be appropriate? I'd also start the paths with a root-relative `/` so we're specific about which paths we're ignoring.
```
const ignorePaths = ['/.angular/cache', '/projects/*/src/assets', '/src/assets'];
for (const ignorePath of ignorePaths) {...}
```

libs/components/packages/src/schematics/migrations/prettier-ignore-updates/prettier-ignore-updates.schematic.ts
```diff
@@ -0,0 +1,30 @@
...
+      '/src/assets',
+      '/projects/*/src/assets',
+      '/src/app/lib',
+    ];
+    const prettierIgnorePath = '/.prettierignore';
+    const prettierIgnore = tree.read(prettierIgnorePath);
+
+    let replacementString = prettierIgnore.toString();
+
+    for (const path of pathsToIgnore) {
```
Blackbaud-SteveBrush: `path` is a reserved word in Node.js. Maybe use `ignorePaths` and `ignorePath`?

libs/components/packages/src/schematics/migrations/prettier-ignore-updates/prettier-ignore-updates.schematic.ts
```diff
@@ -0,0 +1,30 @@
+import { Rule } from '@angular-devkit/schematics';
+
+export default function addNewItemsToPrettierIgnore(): Rule {
+  return (tree) => {
+    const pathsToIgnore = [
+      '/src/assets',
+      '/projects/*/src/assets',
```
Blackbaud-SteveBrush: These paths should be consistent with whatever we choose to do here: https://github.com/blackbaud/skyux/pull/247/files#r871305701

libs/components/packages/src/schematics/migrations/prettier-ignore-updates/prettier-ignore-updates.schematic.ts
```diff
@@ -0,0 +1,30 @@
...
+      '/src/app/lib',
+    ];
+    const prettierIgnorePath = '/.prettierignore';
+    const prettierIgnore = tree.read(prettierIgnorePath);
+
+    let replacementString = prettierIgnore.toString();
```
Blackbaud-SteveBrush: Some projects do not have Prettier installed. We need to add a conditional check before we update the content:
```
const prettierIgnorePath = '/.prettierignore';
const prettierIgnore = tree.read(prettierIgnorePath)?.toString();

if (prettierIgnore) {
  for (const ignorePath of ignorePaths) {
```

libs/components/packages/src/schematics/migrations/prettier-ignore-updates/prettier-ignore-updates.schematic.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    for (const path of pathsToIgnore) {
+      if (
+        prettierIgnore &&
+        !new RegExp(`^` + path.replace('*', '\\*') + `$`, 'm').exec(
+          replacementString.toString()
```
Blackbaud-SteveBrush: `replacementString` is already a `string`. Do we need this `toString()` call?

libs/components/packages/src/schematics/migrations/prettier-ignore-updates/prettier-ignore-updates.schematic.ts
```diff
@@ -0,0 +1,30 @@
...
+    ];
+    const prettierIgnorePath = '/.prettierignore';
+    const prettierIgnore = tree.read(prettierIgnorePath);
+
+    let replacementString = prettierIgnore.toString();
+
+    for (const path of pathsToIgnore) {
+      if (
+        prettierIgnore &&
+        !new RegExp(`^` + path.replace('*', '\\*') + `$`, 'm').exec(
```
Blackbaud-SteveBrush: The `.gitignore` patterns may include other characters that need escaping; to make this future-proof, you can copy an escape function I use in the Builders for this purpose; maybe put it in a new `/utility/regex.ts` file? See: https://dev.azure.com/blackbaud/Products/_git/skyux-angular-builders?path=/src/tools/webpack/plugins/app-assets/app-assets.plugin.ts&version=GB5.x.x&line=17&lineEnd=24&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

libs/components/packages/src/schematics/migrations/prettier-ignore-updates/prettier-ignore-updates.schematic.ts
Blackbaud-TrevorBurch: Done

libs/components/packages/src/schematics/migrations/prettier-ignore-updates/prettier-ignore-updates.schematic.ts
Blackbaud-TrevorBurch: Done

libs/components/packages/src/schematics/migrations/prettier-ignore-updates/prettier-ignore-updates.schematic.ts
Blackbaud-TrevorBurch: Good catch. That existed prior to my last change and I apparently accidentally removed it. Fixed.

libs/components/packages/src/schematics/migrations/prettier-ignore-updates/prettier-ignore-updates.schematic.ts
Blackbaud-TrevorBurch: Good catch. Prior to the last change `replacementString` was a `Buffer` object because of how it got set up. I have fixed this throughout the file.

libs/components/packages/src/schematics/migrations/prettier-ignore-updates/prettier-ignore-updates.schematic.ts
Blackbaud-TrevorBurch: Good call. Done

Blackbaud-SteveBrush (approved):

PR #252, opened 2022-05-13
fix(components/modals): resize observable media query service

johnhwhite:
[AB#2156544](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2156544)

### DIFF
--- a/libs/components/modals/src/lib/modules/modal/modal-content.component.ts
+++ b/libs/components/modals/src/lib/modules/modal/modal-content.component.ts
@@ -1,4 +1,8 @@
 import { Component, ViewEncapsulation } from '@angular/core';
+import {
+  SkyMediaQueryService,
+  SkyResizeObserverMediaQueryService,
+} from '@skyux/core';
 
 /**
  * Specifies content to display in the modal's body.
@@ -8,5 +12,11 @@ import { Component, ViewEncapsulation } from '@angular/core';
   templateUrl: './modal-content.component.html',
   styleUrls: ['./modal-content.component.scss'],
   encapsulation: ViewEncapsulation.None,
+  providers: [
+    {
+      provide: SkyMediaQueryService,
+      useExisting: SkyResizeObserverMediaQueryService,
+    },
+  ],
 })
 export class SkyModalContentComponent {}
--- a/libs/components/modals/src/lib/modules/modal/modal-content.component.ts
+++ b/libs/components/modals/src/lib/modules/modal/modal-content.component.ts
@@ -1,8 +1,4 @@
 import { Component, ViewEncapsulation } from '@angular/core';
-import {
-  SkyMediaQueryService,
-  SkyResizeObserverMediaQueryService,
-} from '@skyux/core';
 
 /**
  * Specifies content to display in the modal's body.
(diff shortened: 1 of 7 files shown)


libs/components/modals/src/lib/modules/modal/modal.component.ts
```diff
@@ -211,6 +202,8 @@ export class SkyModalComponent implements AfterViewInit, OnDestroy {
     /* istanbul ignore next */
     if (this.mediaQueryService) {
       this.mediaQueryService.observe(this.modalContentWrapperElement);
+    } else {
+      console.log(`no this.mediaQueryService`);
```
Blackbaud-SteveBrush: Console log.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -78,6 +85,17 @@ export class SkyModalHostComponent {
       provide: SkyModalConfiguration,
       useValue: params,
     });
+    params.providers.push({
```
Blackbaud-SteveBrush, line 88: I removed this line and the demo seemed to work. Do we need to redefine the `SkyResizeObserverMediaQueryService` provider?

libs/components/modals/src/lib/modules/modal/modal-content.component.ts
```diff
@@ -8,5 +12,11 @@ import { Component, ViewEncapsulation } from '@angular/core';
   templateUrl: './modal-content.component.html',
   styleUrls: ['./modal-content.component.scss'],
   encapsulation: ViewEncapsulation.None,
+  providers: [
+    {
+      provide: SkyMediaQueryService,
+      useExisting: SkyResizeObserverMediaQueryService,
+    },
+  ],
```
Blackbaud-SteveBrush: I removed this provider and the demo still worked. Is this needed since we're providing `SkyMediaQueryService` in the host service?

Blackbaud-SteveBrush (commented):
Just a few observations. There are some good changes in here!

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
johnhwhite: There definitely needs to be a provider with `{ provide: SkyMediaQueryService, useExisting: SkyResizeObserverMediaQueryService }` so that there is a single instance per modal and `SkyResizeObserverMediaQueryService` stands in for the `SkyMediaQueryService`. This one is required for the demo to work.

libs/components/modals/src/lib/modules/modal/modal-content.component.ts
johnhwhite: This was redundant to what `modal-host.component.ts` is providing and is not needed.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
Blackbaud-SteveBrush: We talked offline; probably can get away with removing this line after all.

Blackbaud-SteveBrush (approved):

PR #257, opened 2022-05-17
fix(components/lookup): show more modal opens when triggered from a results dropdown

Blackbaud-TrevorBurch:


libs/components/lookup/src/lib/modules/autocomplete/autocomplete-input.directive.ts
```diff
@@ -92,6 +92,8 @@ export class SkyAutocompleteInputDirective
     this.inputTextValue = this.getValueByKey();
   }
 
+  // TODO: Dead code - remove in a future major version to be safe.
+  // istanbul ignore next
```
Blackbaud-SteveBrush: Since this is not something we "protect" in our public API, I'd be more willing to just remove the "dead" code in this PR and deal with the ramifications on a case-by-case.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.html
```diff
@@ -133,7 +133,7 @@
       tabindex="0"
       type="button"
       [class.focused]="isElementFocused(addButtonEl)"
-      (click)="addButtonClicked()"
+      (mousedown)="addButtonClicked()"
```
Blackbaud-SteveBrush, line 136: This seems odd? If I click (and hold) my mouse button the `moreButtonClicked` event is fired.

Blackbaud-TrevorBurch: Yep. I agree. I did this to match what we already do for the results in the dropdown. 

This one is odd because the issue is caused in-part because of the fact that autocomplete/lookup leaves focus on the input field and keyboard controls fake focus on the dropdown items for styling purposes. This, in a round about way, caused a race condition where we close the dropdown when the input loses focus; however, the click for the modal wasn't registered in time before the dropdown closed and so nothing was happening.

Prior to the last version - we used to try to use clicks on the overlay background to know when to close the dropdown - but that had its own issues (which is why we did the change that led to this bug)

Mousedown happens before `blur` and thus doesn't exhibit the bad behavior. I can leave click - but if I do - would have to add a `setTimeout` to wait on closing the modal and then I'm guessing on how long that should last and it isn't guaranteed to be perfect.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete-input.directive.ts
Blackbaud-TrevorBurch: Removed

Blackbaud-SteveBrush (approved):

PR #264, opened 2022-05-19
fix(components/action-bars): persist focus on summary action bar chevrons after animations

Blackbaud-TrevorBurch:


libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar-adapter.service.ts
```diff
@@ -22,6 +22,14 @@ export class SkySummaryActionBarAdapterService {
     this.renderer = this.rendererFactory.createRenderer(undefined, undefined);
   }
 
+  public focusChevron(chevronElement: ElementRef) {
+    /* sanity-check */
+    /* istanbul ignore else */
+    if (chevronElement) {
```
Blackbaud-SteveBrush: In what scenario would the `chevronElement` be undefined?

Blackbaud-TrevorBurch: I really shouldn’t. However, I did hit this on accident because I had forgotten to add a check for if the animation had fired when the element was destroyed (Angular just does that). 

I added a check to the animation handler so it doesn’t call this code in that case now, but I left the null check here just as a safety net since this is a public function that could be called with `undefined`.

Blackbaud-TrevorBurch: I ignored the branch because there is no current path that leads to `undefined` being passed in and we don’t have any tests that call adapter methods directly.

libs/components/action-bars/package.json
```diff
@@ -16,6 +16,7 @@
   },
   "homepage": "https://github.com/blackbaud/skyux#readme",
   "peerDependencies": {
+    "@angular/animations": "^12.2.16",
```
Blackbaud-TrevorBurch, line 19: A note on this - this is not a breaking change even though it looks like it may be. I have to add this because I now import from the library. However, there is already a peer on `@skyux/animations` that had a peer of `@angular/animations` - so it isn't a breaking change as `@angular/animations` was already required by a required library.

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar-adapter.service.ts
Blackbaud-SteveBrush: I'd either:
1. Remove the `undefined` check, or
2. Make the method's parameter `ElementRef | undefined`.

I'd personally prefer the first option, and add a check for undefined later when a specific scenario is encountered that can be unit tested.

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.ts
```diff
@@ -148,16 +153,26 @@ export class SkySummaryActionBarComponent implements AfterViewInit, OnDestroy {
...
+      if (
+        this.type === SkySummaryActionBarType.Page ||
+        this.type === SkySummaryActionBarType.Tab
+      ) {
+        this.adapterService.styleBodyElementForActionBar(this.elementRef);
+      }
+
+      // Ensure that the correct chevron is fully rendered prior to setting focus.
+      setTimeout(() => {
+        this.adapterService.focusChevron(this.chevronElementRef);
```
Blackbaud-SteveBrush, line 175: Possible to write a unit test that covers this scenario? We have a lot of code being added that's not getting tested.

Blackbaud-TrevorBurch: Facepalm..... that was tired Trevor. Sorry about that. I'll get one added

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar-adapter.service.ts
Blackbaud-TrevorBurch: Removed

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.spec.ts
```diff
@@ -354,23 +376,56 @@ describe('Summary Action Bar component', () => {
...
-          .nativeElement.click();
+        clickExpandButton(debugElement);
         fixture.detectChanges();
         tick();
         fixture.detectChanges();
         expect(cmp.summaryActionBar.isSummaryCollapsed).toBeFalsy();
         expect(cmp.summaryActionBar.slideDirection).toBe('down');
       }));
+
+      it(`should move focus to the collapsed summary's chevron after collapsing`, async () => {
```
Blackbaud-TrevorBurch, line 393: @Blackbaud-SteveBrush This and the next test are the new ones. Everything else here was general cleanup that I did as this file was hard to read in a lot of places and had various `any` and deprecation issues. I'm happy to move the refactors to a different PR if you'd like.

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.ts
Blackbaud-TrevorBurch: Done

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.spec.ts
Blackbaud-SteveBrush: Love it.

Blackbaud-SteveBrush (approved):

PR #265, opened 2022-05-19
fix(components/datetime): update `moment` to version `2.29.3`

Blackbaud-TrevorBurch:


Blackbaud-SteveBrush (commented):
Since this is a dependency of `datetime`, I think we should make a note of this in the changelog:
```
fix(components/datetime): update `moment` to version `2.29.3`
```

Blackbaud-SteveBrush (approved):

PR #279, opened 2022-06-07
fix(components/packages): fix the `ng add` schematic to install essential SKY UX packages

Blackbaud-SteveBrush:


libs/components/packages/src/schematics/ng-add/ng-add.schematic.ts
```diff
@@ -1,70 +1,51 @@
...
     const { workspace } = await getWorkspace(tree);
 
-    const packageJsonPath = 'package.json';
-    const packageJson = JSON.parse(readRequiredFile(tree, packageJsonPath));
-
-    const { packageGroup } = fs.readJsonSync(
-      path.resolve(__dirname, '../../../package.json')
-    )['ng-update'];
-
-    await ensureLatestVersions(context, packageJson.dependencies, packageGroup);
```
Blackbaud-SteveBrush, line 56: ☝️ We were finding the latest versions of _existing_ SKY UX packages when the user ran `ng add @skyux/packages`. This is strange since the consumer will have run `ng add @skyux/packages` before installing any other SKY UX package. I'm assuming this was implemented before we had the `skyux update` command for our internal teams, so I think we can safely remove it.

Blackbaud-TrevorBurch (approved):

PR #282, opened 2022-06-08
fix(components/layout): hide action button with inaccessible skyhref

johnhwhite:
[AB#2181712](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2181712)

### DIFF
--- a/libs/components/layout/src/lib/modules/action-button/action-button-container.component.ts
+++ b/libs/components/layout/src/lib/modules/action-button/action-button-container.component.ts
@@ -65,6 +65,8 @@ export class SkyActionButtonContainerComponent
 
   private ngUnsubscribe = new Subject<void>();
 
+  private syncMaxHeightTimeout?: number;
+
   private set themeName(value: string) {
     this._themeName = value;
     this.updateResponsiveClass();
@@ -163,12 +165,13 @@ export class SkyActionButtonContainerComponent
     );
     if (this._themeName === 'modern') {
       // Wait for children components to complete rendering before height is determined.
-      setTimeout(() => {
+      clearTimeout(this.syncMaxHeightTimeout);
+      this.syncMaxHeightTimeout = setTimeout(() => {
         this.coreAdapterService.syncMaxHeight(
           this.containerRef,
           '.sky-action-button'
         );
-      });
+      }) as unknown as number;
     }
   }
 
--- a/libs/components/layout/src/lib/modules/action-button/action-button-container.component.ts
+++ b/libs/components/layout/src/lib/modules/action-button/action-button-container.component.ts
@@ -6,7 +6,6 @@ import {
   ElementRef,
   HostListener,
   Input,
-  NgZone,
   OnDestroy,
   OnInit,
   Optional,
@@ -14,7 +13,7 @@ import {
   ViewChild,
   ViewEncapsulation,
 } from '@angular/core';
-import { MutationObserverService, SkyCoreAdapterService } from '@skyux/core';
+import { SkyCoreAdapterService } from '@skyux/core';
 import { SkyThemeService } from '@skyux/theme';
 
 import { Subject } from 'rxjs';
@@ -61,8 +60,6 @@ export class SkyActionButtonContainerComponent
   })
   private containerRef: ElementRef<any>;
 
-  private mutationObserver: MutationObserver;
-
   private ngUnsubscribe = new Subject<void>();
 
   private syncMaxHeightTimeout?: number;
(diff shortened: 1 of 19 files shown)


libs/components/layout/src/lib/modules/action-button/action-button-container.component.ts
```diff
@@ -163,12 +165,13 @@ export class SkyActionButtonContainerComponent
...
       // Wait for children components to complete rendering before height is determined.
-      setTimeout(() => {
+      clearTimeout(this.syncMaxHeightTimeout);
+      this.syncMaxHeightTimeout = setTimeout(() => {
         this.coreAdapterService.syncMaxHeight(
           this.containerRef,
           '.sky-action-button'
         );
-      });
+      }) as unknown as number;
```
Blackbaud-TrevorBurch, line 141: Why are we double casting this?

libs/components/layout/src/lib/modules/action-button/action-button.component.spec.ts
```diff
@@ -257,7 +258,7 @@ describe('Action button component modern theme', () => {
     expect(flexParent).not.toHaveCssClass('sky-action-button-flex-align-left');
   });
 
-  it(`should sync all child action buttons to have the same height as the tallest action button`, async(() => {
+  it(`should sync all child action buttons to have the same height as the tallest action button`, fakeAsync(() => {
```
Blackbaud-TrevorBurch, line 276: Won't changing this to `fakeAsync` cause the `expect` inside the `.then` to not matter? I seem to remember that if you were to add `expect(2).toBe(3);` in that `.then` that things will still pass with this setup.

libs/components/router/src/lib/modules/href/href.directive.ts
```diff
@@ -39,12 +41,23 @@ export class SkyHrefDirective {
...
   public set skyHrefElse(value: 'hide' | 'unlink') {
     this._skyHrefElse = value;
     this.applyChanges(this.getChanges());
   }
 
+  /**
+   * Emits whether the link is available (true) or not (false).
+   */
+  @Output()
+  public skyHrefChange = new EventEmitter<boolean>();
```
Blackbaud-TrevorBurch: Should we consider making this return an object that includes the available state to future proof this so that we could use the same event in the future if we discover more information that we want to update about?

libs/components/layout/src/lib/modules/action-button/action-button.component.spec.ts
```diff
@@ -100,17 +99,19 @@ describe('Action button component', () => {
...
     const actionButton = el.querySelectorAll('.sky-action-button').item(1);
-    expect(actionButton.tagName === 'a');
+    expect(actionButton.tagName.toLowerCase() === 'a').toBeTrue();
     expect(actionButton.getAttribute('href')).toBe(
       'https://developer.blackbaud.com/skyux/components'
     );
   }));
 
-  it('should see if there is a permalink route included as an input to the element', () => {
+  it('should see if there is a permalink route included as an input to the element', fakeAsync(() => {
```
Blackbaud-TrevorBurch, line 108: I'm seeing changes to existing tests but not a new test to make sure that we are properly hiding the one new button that you added to the fixture and letting the other one through. Should we have a new test for that?

Blackbaud-TrevorBurch (changes requested):
@johnhwhite - just a few questions for you

libs/components/layout/src/lib/modules/action-button/action-button-container.component.ts
johnhwhite: `@types/node` has a different response type, but the browser response type is a number, and that's preferable to referencing a node type.

libs/components/layout/src/lib/modules/action-button/action-button.component.spec.ts
johnhwhite: Good catch.

libs/components/router/src/lib/modules/href/href.directive.ts
johnhwhite: Good call.

libs/components/layout/src/lib/modules/action-button/action-button.component.spec.ts
johnhwhite: Added a test.

Blackbaud-TrevorBurch (approved):
@johnhwhite code looks good. However, I did want to point out that adding the `peerDependency` is technically a breaking change. We have a migration schematic for it though. We should probably update what version that schematic runs on and add `layout` as a library that that schematic looks for. Most consumers will probably have `cdk` but we should probably run that schematic since it isn't 100% that all `layout` consumers will have it.

PR #306, opened 2022-06-22
chore(apps/playground): refresh playground SPA routing and homepage

Blackbaud-TrevorBurch:


apps/playground/src/app/app-routing.module.ts
```diff
@@ -9,7 +9,9 @@ const routes: Routes = [
   {
     path: 'components',
     loadChildren: () =>
-      import('./components/components.module').then((m) => m.ComponentsModule),
+      import('./components/components.module').then(
+        (m) => m.ComponentsRoutingModule
```
Blackbaud-SteveBrush: Should be `ComponentsModule`?

apps/playground/src/app/home/home-filter.component.ts
```diff
@@ -0,0 +1,46 @@
...
+  templateUrl: './home-filter.component.html',
+})
+export class HomeFiltersModalDemoComponent {
+  public libraries: { name: string; isSelected: boolean }[];
+
+  constructor(
+    public context: SkyDataManagerFilterModalContext,
+    public instance: SkyModalInstance
+  ) {
+    console.log(this.context);
```
Blackbaud-SteveBrush: ?

apps/playground/src/app/home/home.component.ts
```diff
@@ -1,7 +1,201 @@
...
+            this.filterItems(this.searchItems(this.componentData))
+          );
+          this.changeDetector.detectChanges();
+        }
+      });
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private createComponentData(routes: ComponentRouteInfo[], parentPath: string): Promise<any> {
+    console.log(routes);
```
Blackbaud-SteveBrush: ?

apps/playground/src/app/components/components.module.ts
```diff
@@ -1,26 +1,29 @@
...
 
-const routes: Routes = [
+import { AgGridRoutingModule } from './ag-grid/ag-grid-routing.module';
+
+export const componentRoutes: Routes = [
   {
     path: 'a11y',
-    loadChildren: () => import('./a11y/a11y.module').then((m) => m.A11yModule),
+    loadChildren: () =>
+      import('./a11y/a11y.module').then((m) => m.A11yRoutingModule),
```
Blackbaud-SteveBrush: 👇 There are a few places where you're importing the routing module, instead of the feature module. Intentional?

apps/playground/src/app/components/components.module.ts
```diff
@@ -86,10 +89,12 @@ const routes: Routes = [
 ];
 
 @NgModule({
-  imports: [RouterModule.forChild(routes)],
+  imports: [RouterModule.forChild(componentRoutes)],
   exports: [RouterModule],
 })
-export class ComponentsRoutingModule {}
+export class ComponentsRoutingModule {
+  public static routes = componentRoutes;
```
Blackbaud-SteveBrush: Is there not a way to get at the route configs using the `Router` injector, instead of using a static property on each module?

apps/playground/src/app/components/ag-grid/readonly-grid/readonly-grid.module.ts
```diff
@@ -21,4 +21,6 @@ import { ReadonlyGridComponent } from './readonly-grid.component';
     SkyDropdownModule,
   ],
 })
-export class ReadonlyGridModule {}
+export class ReadonlyGridModule {
+  public static routes = ReadonlyGridRoutingModule.routes;
```
Blackbaud-SteveBrush, line 25: Putting the static `routes` property on the feature modules as well as the routing module?

apps/playground/src/app/components/components.module.ts
Blackbaud-TrevorBurch: No. Because we lazy load the modules this isn’t possible. I didn’t quite a bit of research on this. The only thing returned from the router is the routes relevant to either the base or current route. These routes then have a `loadChildren` method that returns back the feature module type but not the inner routes. Putting that static property allows us to grab the information we need without having an instance of the module as that is what Angular gives us.

apps/playground/src/app/components/ag-grid/readonly-grid/readonly-grid.module.ts
Blackbaud-TrevorBurch: Yes. This is only needed if the routing module is in a different file from the feature module. We need the information to be on the feature module. However, the routes are often declared with the routing module and so we need to be able to grab them to put them on the feature module. I could have just exported a `const` but I opted for just putting them on a static property like we were doing elsewhere.

apps/playground/src/app/components/components.module.ts
Blackbaud-TrevorBurch: No. That was unintentional and was from an earlier version. I’ll fix that.

apps/playground/src/app/home/home.component.ts
Blackbaud-TrevorBurch: These console logs were meant to be removed. I’ll do that.

apps/playground/src/app/app-routing.module.ts
Blackbaud-TrevorBurch: Yes. I meant to revert that. I’ll do that.

Blackbaud-TrevorBurch: Fixed all four of these.

apps/playground/src/app/home/home-filter.component.ts
Blackbaud-TrevorBurch: Fixed

apps/playground/src/app/home/home.component.ts
Blackbaud-TrevorBurch: Fixed

apps/playground/src/app/components/components.module.ts
Blackbaud-TrevorBurch: Fixed all four of these.

Blackbaud-SteveBrush (approved):

PR #308, opened 2022-06-24
fix(components/theme): use 6px border radius for sky-rounded-corners in Modern Visual Style

bradenbiz:


Blackbaud-SteveBrush (commented):
I'd remove the `package-lock.json` from the PR since it will cause the build cache to restart on the PR pipelines.

PR #310, opened 2022-06-27
refactor(components/lookup): implement 100% coverage, strict-mode, and private instance fields for autocomplete

Blackbaud-SteveBrush:


libs/components/lookup/src/lib/modules/autocomplete/autocomplete-input.directive.ts
```diff
@@ -121,29 +121,29 @@ export class SkyAutocompleteInputDirective
...
 
-  private _autocompleteAttribute: string;
+  private _autocompleteAttribute: string | undefined;
 
   private _blur = new Subject<void>();
 
   private _disabled = false;
 
-  private _displayWith: string;
+  private _displayWith = '';
```
Blackbaud-PaulCrowder: Why not change the properties with underscores to use the private `#` designation as well?

Blackbaud-SteveBrush: I didn't want to override out convention of using the underscore for getter/setter properties.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete-input.directive.ts
```diff
@@ -199,8 +199,8 @@ export class SkyAutocompleteInputDirective
   }
 
   public ngOnDestroy(): void {
-    this._blur.complete();
-    this._textChanges.complete();
+    this.#_blur.complete();
+    this.#_textChanges.complete();
```
Blackbaud-PaulCrowder: This seems wrong; I'd only expect properties that start with an underscore to be referenced in property accessors, hence the pattern. I'd expect any other private field not exclusively used inside property accessors to not have an underscore prefix. There are lots of these instances across the files in the PR.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.spec.ts
```diff
@@ -561,28 +561,22 @@ describe('Autocomplete component', () => {
...
-      }
+      }).toThrowError(
+        'The SkyAutocompleteComponent requires a ContentChild input or textarea bound with the SkyAutocomplete directive. ' +
+          'For example: `<input type="text" skyAutocomplete>`.'
+      );
     }));
 
     it('should set the width of the dropdown when a search is performed', fakeAsync(() => {
       const adapterSpy = spyOn(
         autocomplete['adapterService'],
```
Blackbaud-PaulCrowder: This seems like a place where we should either use `TestBed.overrideComponent()` to provide our own adapter service that we spy on, or use `fixture.debugElement.injector.get()` to get the adapter service, instead of accessing a private member.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete-input.directive.ts
Blackbaud-SteveBrush: That's what I get for doing a find/replace. I'll go through the files with more care. 👍

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.spec.ts
Blackbaud-SteveBrush: Good call. Done.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -470,12 +482,16 @@ export class SkyAutocompleteComponent implements OnDestroy, AfterViewInit {
     this.resultsListId = `sky-autocomplete-list-${id}`;
     this.resultsWrapperId = `sky-autocomplete-wrapper-${id}`;
 
+    this.#activeElementIndex = -1;
```
Blackbaud-PaulCrowder: I don't think you need to change it in this PR, but for future reference I think it's OK to assign static values when declaring private fields. It's only when you assign one private field's value to another private field's value during declaration that creates some ambiguity as to which is set first.

Blackbaud-PaulCrowder (approved):

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
Blackbaud-SteveBrush: Reverted. I was thinking that since we're assigning some values in the constructor, _all_ of them should be. But now I don't like that as much.

Blackbaud-PaulCrowder (approved):

PR #315, opened 2022-06-28
refactor(components/lookup): implement strict-mode for lookup, search, and country field

Blackbaud-SteveBrush:


libs/components/indicators/src/lib/modules/tokens/tokens.component.ts
```diff
@@ -207,7 +207,7 @@ export class SkyTokensComponent implements OnDestroy {
   private _disabled: boolean;
   private _dismissible: boolean;
   private _focusable: boolean;
-  private _tokens: SkyToken[];
+  private _tokens: SkyToken[] | undefined;
```
Blackbaud-SteveBrush, line 210: ☝️ I'll create a separate PR to refactor the Indicators stuff.

libs/components/lookup/src/lib/modules/search/search.component.ts
```diff
@@ -207,7 +207,7 @@ export class SkySearchComponent implements OnDestroy, OnInit, OnChanges {
     if (searchText !== this.searchText) {
       this.searchText = searchText;
     }
-    this.clearButtonShown = searchText && searchText !== '';
+    this.clearButtonShown = !!(searchText && searchText !== '');
```
Blackbaud-SteveBrush, line 210: ☝️ Search and country field will be done in a future PR.

libs/components/indicators/src/lib/modules/tokens/tokens.component.ts
```diff
@@ -207,7 +207,7 @@ export class SkyTokensComponent implements OnDestroy {
   private _disabled: boolean;
```
Blackbaud-PaulCrowder, line 207: Should we convert these to private-private in this PR?

libs/components/lookup/src/lib/modules/lookup/types/lookup-show-more-custom-picker-context.ts
```diff
@@ -6,13 +6,13 @@ export class SkyLookupShowMoreCustomPickerContext {
   /**
    * Specifies the options that users can select in the custom picker.
    */
-  public items: any[];
+  public items!: any[];
```
Blackbaud-PaulCrowder, line 9: What's the purpose of `!` here?

Blackbaud-SteveBrush: The `items` property is always set when we create the context instance, so it won't ever be undefined. Better way to handle?

libs/components/indicators/src/lib/modules/tokens/tokens.component.ts
Blackbaud-SteveBrush: I was going to tackle `@skyux/indicators` in a separate PR. I had to make these changes to get Lookup working.

Blackbaud-PaulCrowder (approved):

PR #320, opened 2022-06-30
fix(components/datetime): datepicker handles reactive forms which are disabled during initialization

Blackbaud-TrevorBurch:


libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -347,6 +347,10 @@ export class SkyDatepickerInputDirective
   public validate(control: AbstractControl): ValidationErrors {
     if (!this.control) {
       this.control = control;
+      // Account for any date converstion that may have occurred prior to validation.
```
Blackbaud-SteveBrush: 'converstion'

libs/components/datetime/src/lib/modules/datepicker/datepicker.component.spec.ts
```diff
@@ -1253,6 +1253,25 @@ describe('datepicker', () => {
...
+        nativeElement = fixture.nativeElement as HTMLElement;
+        component = fixture.componentInstance;
+
+        // Default to US long date format to avoid any test runners that are using a different locale.
+        component.dateFormat = 'MM/DD/YYYY';
+
+        component.initialValue = '5/12/2017';
+        component.disableFormOnCreation = true;
+        detectChanges(fixture);
+
```
Blackbaud-SteveBrush, line 1268: Shouldn't we expect the form control's value to also be a `Date` value before we enable the form control? I put the following expectation on this line and the test failed.
```
expect(component.dateControl.value).toEqual(new Date('5/12/2017'));
```
The value of the form control should always be correct, regardless if it's disabled or not.

Blackbaud-TrevorBurch: So that is where the issue comes in. I agree...... but we don't have a reference to the control internally to the input directive at the point that `writeValue` is called to update the control's value with the converted date. If you have ideas as to how to get this - I'm all ears :). But without that we have to wait until we have that reference which is in the `validate` function....

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
Blackbaud-TrevorBurch: Fixed

libs/components/datetime/src/lib/modules/datepicker/datepicker.component.spec.ts
Blackbaud-TrevorBurch: Actually, I found that some of our other components are grabbing the control a different way. Let me see if I can quickly convert datepicker to do this to solve the issue here.

Blackbaud-TrevorBurch: After some trial and error and discussions with @Blackbaud-SteveBrush I have updated the input directive to grab the control via the injector when needed and am utilizing a set timeout to address the inner "AbstractControl" being undefined if the control is initially disabled when `writeValue` is called from Angular.

Blackbaud-TrevorBurch: I reverted the previously mentioned change as it caused a race condition between when `writeValue` fully succeeded and the `validate` method.

johnhwhite (approved):

PR #325, opened 2022-07-06
fix(components/theme): modify margin-bottom styles in Default Visual Style

bradenbiz:


libs/components/theme/src/lib/styles/_sizing.scss
```diff
@@ -117,13 +117,13 @@
...
+  margin-bottom: $sky-space-md;
 }
 
 .sky-margin-stacked-xl {
   margin-bottom: $sky-space-xl;
 }
 
 .sky-margin-stacked-xxl {
-  margin-bottom: $sky-space-xl;
+  margin-bottom: 40px;
```
Blackbaud-TrevorBurch, line 128: Did we double check that we don't have any variable for 40px?

Blackbaud-TrevorBurch (approved):

libs/components/theme/src/lib/styles/_sizing.scss
bradenbiz: Correct, there are no variables for 40px

PR #333, opened 2022-07-07
fix(components/layout): keep box component's control button placement static when there is no header

bradenbiz:


libs/components/layout/src/lib/modules/box/box.component.html
```diff
@@ -9,8 +9,12 @@
   }"
 >
   <div class="sky-box-header-content">
```
Blackbaud-SteveBrush, line 11: Instead of creating more DIV elements, could you simply add `justify-content: flex-end;` to the `.sky-box-header-content` CSS rule?

bradenbiz: Unfortunately, no. Adding `justify-content: flex-end;` would move both the header and the controls to the end of the flexbox.

bradenbiz: I just found a different way to accomplish the same thing. `margin-left: auto` on the controls selector. I'll make the change as soon as I can.

bradenbiz: https://jsfiddle.net/9b372nmL/18/
see here^

bradenbiz:
closing this PR to save time reverting the changes I made

PR #337, opened 2022-07-08
ci: add e2e workflow

johnhwhite:
[AB#2194030](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2194030)


.github/workflows/e2e-branch.yml
```diff
@@ -0,0 +1,169 @@
...
+          key: cache-nx-publish-${{ hashFiles('package-lock.json') }}
+          restore-keys: |
+            cache-nx-publish-
+            cache-nx-
+      - name: Download artifacts
+        uses: actions/download-artifact@v3
+        with:
+          path: ./dist/storybooks/{{ github.ref_name }}/
+      - name: Generate Storybook Composition
+        run: npx nx g @skyux/e2e-schematics:storybook-composition --projectsJson '${{ needs.install-deps.outputs.projects }}' --baseUrl ../storybooks/${{ github.ref_name }}/
```
Blackbaud-SteveBrush: Just noticing this now, but we should have named the schematics library `@skyux-sdk/e2e-schematics`. Possible to rename in this PR?

.github/workflows/e2e-branch.yml
```diff
@@ -0,0 +1,169 @@
...
+      - name: Build Storybook Composition
+        run: npx nx run storybook:build-storybook:ci --outputDir=dist/storybook
+      - name: Checkout ${{ env.GH_PAGES_REPO }}
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ env.GH_PAGES_OWNER }}/${{ env.GH_PAGES_REPO }}
+          ref: 'main'
```
Blackbaud-SteveBrush: Is `ref: 'main'` needed? Would we need to change this value once we start running this workflow against the `6.x.x` branch?

.github/workflows/e2e-pr.yml
```diff
@@ -0,0 +1,185 @@
...
+          path: ./dist/storybooks/
+      - name: Generate Storybook Composition
+        run: npx nx g @skyux/e2e-schematics:storybook-composition --projectsJson '${{ needs.install-deps.outputs.projects }}' --baseUrl ../storybooks/
+      - name: Build Storybook Composition
+        run: npx nx run storybook:build-storybook:ci --outputDir=dist/storybook
+      - name: Checkout ${{ env.GH_PAGES_REPO }}
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ env.GH_PAGES_OWNER }}/${{ env.GH_PAGES_REPO }}
+          ref: 'main'
```
Blackbaud-SteveBrush: Is `ref: 'main'` needed? Would we need to change this value once we start running this workflow against the `6.x.x` branch?

.github/workflows/e2e-pr.yml
```diff
@@ -0,0 +1,185 @@
...
+        run: |
+          set -x
+          node ./scripts/rebase-pr.js
+          npm ci
+      - name: Derive appropriate SHAs for base and head for `nx affected` commands
+        id: setSHAs
+        uses: nrwl/nx-set-shas@v2
+      - name: Check affected
+        id: affected
+        run: npx ts-node scripts/github-workflow-e2e-matrix
```
Blackbaud-SteveBrush: Capturing the affected projects doesn't make much sense in the context of the "Install and cache dependencies" step. I'd move these outputs into a separate step designed to capture variables that will be shared across all steps.

scripts/github-workflow-e2e-matrix.ts
```diff
@@ -0,0 +1,36 @@
...
+    'npx',
+    'nx',
+    'print-affected',
+    '--target=build-storybook',
+    all ? ' --all' : ''
+  );
+  const affected = JSON.parse(affectedJson);
+  const projects = affected.tasks
+    .map((task: any) => task.target.project)
+    .filter((project: string) => project && project !== 'storybook');
```
Blackbaud-SteveBrush: Could you use `--exclude=storybook` in the CLI command instead? This will make it easier to include other excluded projects in the future.

scripts/github-workflow-e2e-matrix.ts
```diff
@@ -0,0 +1,36 @@
...
+  const affectedJson = runCommand(
+    'npx',
+    'nx',
+    'print-affected',
+    '--target=build-storybook',
+    all ? ' --all' : ''
+  );
+  const affected = JSON.parse(affectedJson);
+  const projects = affected.tasks
+    .map((task: any) => task.target.project)
```
Blackbaud-SteveBrush: You can pass `--select=tasks.target.project` to the CLI command to return only the project names.
See: https://github.com/blackbaud/skyux-dev-infra/blob/main/lib/test-affected/handler.ts#L41

scripts/github-workflow-e2e-matrix.ts
```diff
@@ -0,0 +1,36 @@
...
+    .filter((project: string) => project && project !== 'storybook');
+  const include = projects.map((project: string) => {
+    return {
+      project,
+      token: `PERCY_TOKEN_${project.toUpperCase().replace(/-/g, '_')}_E2E`,
+    };
+  });
+  console.log(`::set-output name=projects::${JSON.stringify(projects)}`);
```
Blackbaud-SteveBrush: The contents of these `console.log`s are critical to the functionality of your GitHub workflow, but to an untrained eye I doubt a future engineer would understand their importance. What if we moved the `::set-output` bits back into the workflow file (so that they communicate proper context to a future developer reading through the YAML file), and create a new `dev-infra` command that prints the affected projects in JSON format like this:

**e2e-pr.yml**
```
run: |
  echo ::set-output name=affectedStorybookProjects::$(npx skyux-dev print-affected --target=build-storybook)
```
The `skyux-dev print-affected` command would:
1. echo a JSON-parsable string of all the affected projects to the console in the following format: `{[projectName: string]: {screenshotAccessToken: string;}}` (we can add more fields in the future to accommodate more workflow tasks)
2. automatically exclude certain projects from the output (such as `ci` and `storybook`)

This same functionality could be applied to the existing workflow files so that all workflows use the same logic to determine affected projects.

**NOTE**
For the sake of this PR, we can keep the logic in the `/scripts/github-workflow-e2e-matrix.ts` file if you'd like, with the expectation that both scripts would be eventually moved to the `dev-infra` repo (so they can be properly unit tested). If you choose this route, the command in your YML file would look (something) like:
```
run: |
  echo ::set-output name=affectedStorybookProjects::$(npx ts-node scripts/github-workflow-e2e-matrix)
```

.github/workflows/e2e-pr.yml
```diff
@@ -0,0 +1,185 @@
+name: E2E
```
Blackbaud-SteveBrush: These two YML files are nearly identical; could we attempt to combine them into one and use a conditional to determine if the run is a pull request or a push?

.github/workflows/e2e-branch.yml
johnhwhite: On a GitHub Pages repo, only one branch can be published, and it is set to the default `main` branch. Multiple branch Storybooks publish to `blackbaud/skyux-storybook` and multiple PR previews publish to `blackbaud/skyux-pr-preview`, but they all publish to the `main` branch of the respective repo.

.github/workflows/e2e-pr.yml
johnhwhite: On a GitHub Pages repo, only one branch can be published, and it is set to the default `main` branch. Multiple branch Storybooks publish to `blackbaud/skyux-storybook` and multiple PR previews publish to `blackbaud/skyux-pr-preview`, but they all publish to the `main` branch of the respective repo.

.github/workflows/e2e-branch.yml
johnhwhite: I've renamed the schematics library.

.github/workflows/e2e-pr.yml
johnhwhite: The name wasn't accurate for what this step does. I've updated the name.

scripts/github-workflow-e2e-matrix.ts
johnhwhite: I switched to `--exclude=storybook`, and this code has moved to the `dev-infra` repo.

scripts/github-workflow-e2e-matrix.ts
johnhwhite: I added `--select=tasks.target.project`, and this code has moved to the `dev-infra` repo.

.github/workflows/e2e-pr.yml
johnhwhite: I've consolidated the workflows into one.

scripts/github-workflow-e2e-matrix.ts
johnhwhite: The JSON is now generated in a `dev-infra` command, with one JSON document specifying the two workflow matrices as well as some simple variables.

Blackbaud-PaulCrowder (approved):

PR #338, opened 2022-07-08
fix(components/layout): keep box component's control button placement static when there is no header

bradenbiz:


Blackbaud-SteveBrush (approved):
Nice.

PR #342, opened 2022-07-14
feat(components/pages): support action hub settings

johnhwhite:
[AB#2226988](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2226988)

### DIFF
--- a/libs/components/packages/src/schematics/migrations/migration-collection.json
+++ b/libs/components/packages/src/schematics/migrations/migration-collection.json
@@ -59,6 +59,11 @@
       "version": "6.0.1",
       "factory": "./update-6/prettier-ignore-updates.schematic",
       "description": "Add new items to .prettierignore"
+    },
+    "add-pages-peer-dependency": {
+      "version": "6.7.0",
+      "factory": "./update-6/add-pages-peer-dependency.schematic",
+      "description": "Add '@skyux/pages' peer dependency"
     }
   }
 }
--- a/libs/components/packages/src/schematics/migrations/migration-collection.json
+++ b/libs/components/packages/src/schematics/migrations/migration-collection.json
@@ -60,10 +60,10 @@
       "factory": "./update-6/prettier-ignore-updates.schematic",
       "description": "Add new items to .prettierignore"
     },
-    "add-pages-peer-dependency": {
+    "add-pages-modals-peer-dependency": {
       "version": "6.7.0",
-      "factory": "./update-6/add-pages-peer-dependency.schematic",
-      "description": "Add '@skyux/pages' peer dependency"
+      "factory": "./update-6/add-pages-modals-peer-dependency.schematic",
+      "description": "Add '@skyux/modals' peer dependency to '@skyux/pages' package"
     }
   }
 }
--- a/libs/components/pages/src/lib/modules/action-hub/types/page-link.ts
+++ b/libs/components/pages/src/lib/modules/action-hub/types/page-link.ts
@@ -7,10 +7,6 @@ import { SkyPageLinkInterface } from './page-link-interface';
  */
 export interface SkyPageLink extends SkyPageLinkInterface {
   /**
-   * Specifies the link text.
-   */
-  label: string;
-  /**
    * Specifies the link destination.
    */
   permalink: {
--- a/libs/components/pages/src/lib/modules/action-hub/types/page-link.ts
+++ b/libs/components/pages/src/lib/modules/action-hub/types/page-link.ts
@@ -1,9 +1,11 @@
 import { NavigationExtras } from '@angular/router';
 
+import { SkyPageLinkInterface } from './page-link-interface';
+
 /**
  * Displays links to related information or recently accessed items.
  */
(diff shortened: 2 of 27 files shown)


libs/components/packages/src/schematics/migrations/migration-collection.json
```diff
@@ -59,6 +59,11 @@
       "version": "6.0.1",
       "factory": "./update-6/prettier-ignore-updates.schematic",
       "description": "Add new items to .prettierignore"
+    },
+    "add-pages-peer-dependency": {
```
Blackbaud-TrevorBurch: Should we be a little more specific here and maybe say `add-pages-modal-peer-dependency`? That way if we add another in the future we don't have to deal with having this one? Not a big deal at all as we could just be more specific later - just thinking that the change might help avoid future confusion

libs/components/pages/src/lib/modules/modal-link-list/modal-link-list.component.html
```diff
@@ -0,0 +1,41 @@
...
+    >
+      <li
+        *ngFor="let link of links"
+        [skyThemeClass]="{
+          'sky-margin-stacked-default': 'default',
+          'sky-margin-stacked-sm': 'modern'
+        }"
+      >
+        <button
+          (click)="openModal(link)"
```
Blackbaud-TrevorBurch, line 42: The requirements in the story say "It should be able to open links and modals" - but this code seems to only allow for modals. Was this intentional?

libs/components/pages/src/lib/modules/modal-link-list/modal-link-list.component.html
```diff
@@ -0,0 +1,41 @@
...
+  <sky-wait [isWaiting]="true"></sky-wait>
+  <ng-container [ngTemplateOutlet]="headingTemplateRef"></ng-container>
+</ng-template>
+<ng-template #linksTemplateRef>
+  <ng-container *ngIf="links?.length > 0">
+    <ng-container [ngTemplateOutlet]="headingTemplateRef"></ng-container>
+    <ul
+      class="sky-modal-link-list"
+      [skyThemeClass]="{
+        'sky-margin-stacked-separate': 'default',
```
Blackbaud-TrevorBurch: Per @Blackbaud-AdamFunderburk - looks like the list "blocks" of links are using one the deprecated spacing classes `sky-margin-stacked-separate`

libs/components/pages/src/lib/modules/action-hub/types/page-modal-link.ts
```diff
@@ -0,0 +1,20 @@
...
+import { SkyPageLinkInterface } from './page-link-interface';
+
+/**
+ * Displays links to related information or recently accessed items.
+ */
+export interface SkyPageModalLink extends SkyPageLinkInterface {
+  /**
+   * Specifies the link text.
+   */
+  label: string;
```
Blackbaud-TrevorBurch: Since `label` is a part of `SkyPageLinkInterface` it is redundant to list it here too, correct?

libs/components/pages/src/lib/modules/action-hub/types/page-link.ts
```diff
@@ -1,9 +1,11 @@
...
+import { SkyPageLinkInterface } from './page-link-interface';
+
 /**
  * Displays links to related information or recently accessed items.
  */
-export interface SkyPageLink {
+export interface SkyPageLink extends SkyPageLinkInterface {
   /**
    * Specifies the link text.
    */
```
Blackbaud-TrevorBurch: I don't believe we need to have `label` in this interface any more as it is a part of `SkyPageLinkInterface`

Blackbaud-TrevorBurch (changes requested):
Just a few questions

libs/components/packages/src/schematics/migrations/migration-collection.json
johnhwhite: Renamed it.

libs/components/pages/src/lib/modules/modal-link-list/modal-link-list.component.html
johnhwhite: Good catch.

libs/components/pages/src/lib/modules/modal-link-list/modal-link-list.component.html
johnhwhite: Updated.

libs/components/pages/src/lib/modules/action-hub/types/page-modal-link.ts
johnhwhite: Good catch.

libs/components/pages/src/lib/modules/action-hub/types/page-link.ts
johnhwhite: Good catch.

Blackbaud-TrevorBurch (approved):

PR #347, opened 2022-07-19
fix(components/modals): viewkept toolbars now style correctly at the top of modern theme modals

Blackbaud-TrevorBurch:


libs/components/modals/src/lib/modules/modal/modal.component.scss
```diff
@@ -188,6 +189,23 @@
     padding-top: initial;
     width: 100%;
   }
+
+  .sky-modal-viewkeeper {
+    .sky-modal-header {
+      box-shadow: none !important;
+    }
+
+    .sky-modal-content
```
johnhwhite, line 198: This seems like it would be better off in `modal-content.component.scss`, which has view encapsulation disabled and wouldn't require `::ng-deep`.

johnhwhite: The alternatives weren't better. Sticking with `::ng-deep`.

johnhwhite (approved):

PR #348, opened 2022-07-20
fix(components/data-manager): add column picker message and disable button when no columns selected

bradenbiz:

PR #349, opened 2022-07-20
fix(components/data-manager): fix column picker once-ability and disable & warn when no columns selected

bradenbiz:


libs/components/data-manager/src/lib/modules/data-manager/data-manager-column-picker/data-manager-column-picker.component.html
```diff
@@ -25,9 +26,17 @@
     </sky-data-manager>
   </sky-modal-content>
   <sky-modal-footer>
+    <sky-alert
```
Blackbaud-TrevorBurch: An alert seems odd here. Did we check that with Adam? I would think that we would at least use a status indicator here instead of an alert based on our form design guidelines https://developer.blackbaud.com/skyux/design/guidelines/form-design#validation-and-error-handling

Blackbaud-TrevorBurch (dismissed):
@bradenbiz - one question

libs/components/data-manager/src/lib/modules/data-manager/data-manager-column-picker/data-manager-column-picker.component.html
bradenbiz: resolved in the comments of a previous PR that is pulled into this: https://github.com/blackbaud/skyux/pull/348

libs/components/data-manager/src/lib/modules/data-manager/data-manager-column-picker/data-manager-column-picker.component.scss
```diff
@@ -3,3 +3,7 @@
 .sky-data-manager-column-picker-content sky-repeater-item-content {
   margin-top: $sky-margin-inline-compact;
 }
+
+.sky-data-manager-column-picker-footer-stacked-indicator {
+  margin: 0px 5px 10px 0px;
```
Blackbaud-PaulCrowder: You should use the appropriate SCSS variables for the `5px` and `10px` values (see the class right above it for reference). Also you don't need to specify `px` when the value is `0` since `0` is the same in any unit (e.g. `margin: 0 $some-var $some-other-var 0;`

libs/components/data-manager/src/lib/modules/data-manager/data-manager-column-picker/data-manager-column-picker.component.ts
```diff
@@ -96,6 +89,22 @@ export class SkyDataManagerColumnPickerComponent implements OnDestroy, OnInit {
...
+    );
+
+    if (this.dataState.onlyShowSelected) {
+      this.displayedColumnData = this.displayedColumnData.filter(
+        (col) => col.isSelected
+      );
+    }
+
+    this.isAnyDisplayedColumnSelected = this.displayedColumnData.some(
+      (aDisplayedColumn) => aDisplayedColumn.isSelected
```
Blackbaud-PaulCrowder: Change `aDisplayedColumn` to `col` to match the two functions above this one.

libs/components/data-manager/src/lib/modules/data-manager/data-manager-column-picker/data-manager-column-picker.component.scss
bradenbiz: fixed in commit: [use scss variables](https://github.com/blackbaud/skyux/pull/349/commits/4e57053cf6e5bed05e7d0196b4fe42f2cddf002c)

libs/components/data-manager/src/lib/modules/data-manager/data-manager-column-picker/data-manager-column-picker.component.ts
bradenbiz: fixed in commit: [rename](https://github.com/blackbaud/skyux/pull/349/commits/8e6bdadfec94e26f666d83e38fedad1d3b1ca731)

libs/components/data-manager/src/lib/modules/data-manager/data-manager-column-picker/data-manager-column-picker.component.scss
```diff
@@ -3,3 +3,7 @@
 .sky-data-manager-column-picker-content sky-repeater-item-content {
   margin-top: $sky-margin-inline-compact;
 }
+
+.sky-data-manager-column-picker-footer-stacked-indicator {
+  margin: 0 $sky-margin-stacked-compact $sky-margin-stacked-default 0px;
```
Blackbaud-PaulCrowder: You've still got a `0px` hanging out here.

Blackbaud-PaulCrowder: Also, I would think you'd use one of the `$sky-margin-inline-` variables for the right margin instead of `$sky-margin-stacked-` which is for top/bottom margins.

bradenbiz: fixed in commit: [correct styles](https://github.com/blackbaud/skyux/pull/349/commits/5f739edfa7c29a842e8c586138458f7c64b3d4ec)

Blackbaud-PaulCrowder (approved):

PR #352, opened 2022-07-21
refactor(components/layout): replace description-list mutation observer with css solution

johnhwhite:


libs/components/layout/src/lib/modules/description-list/description-list.component.spec.ts
```diff
@@ -251,7 +254,8 @@ describe('Description list component', () => {
     expect(termEls.length).toEqual(1);
     expect(descriptionEls.length).toEqual(1);
     expect(termEls[0]).toHaveText('boo');
-    expect(descriptionEls[0]).toHaveText('far');
+    expect(descriptionEls[0].firstChild).toHaveText('far');
+    expect(descriptionEls[0].childNodes[1]).not.toBeVisible();
```
Blackbaud-TrevorBurch, line 259: I don't see a test where we added a `toBeVisible` check for when that is the case. Should we have one just to verify that the class isn't applying when it shouldn't be? Maybe it would go in the test at line 152?

Blackbaud-TrevorBurch (changes requested):
@johnhwhite one small test suggestion

libs/components/layout/src/lib/modules/description-list/description-list.component.spec.ts
johnhwhite: Added the additional assertion.

Blackbaud-TrevorBurch (approved):

PR #364, opened 2022-07-28
feat(components/lookup): add lookup testing harness

Blackbaud-SteveBrush:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/364/storybook/)

Component Storybooks:

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/364/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/364/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/364/playground/)

libs/components/lookup/src/lib/modules/lookup/lookup.component.spec.ts
```diff
@@ -421,6 +425,11 @@ describe('Lookup component', function () {
     tick();
   }
 
+  function verifyPickerId() {
+    const modalEl = getModalEl();
+    expect(modalEl.id).toMatch(/sky-id-gen__[0-9]+__[0-9]/);
```
Blackbaud-PaulCrowder: This regex depends on the internal implementation of the ID directive. That implementation changed recently, which would have broken this test. This should either mock out ID service to return a predictable ID in the test or make this more generic, like testing that `id` is not null.

libs/components/lookup/src/lib/modules/lookup/lookup.component.ts
```diff
@@ -605,6 +606,8 @@ export class SkyLookupComponent
 
   public openPicker(initialSearch: string): void {
     if (this.showMoreConfig?.customPicker) {
+      this.showMorePickerId = null;
```
Blackbaud-PaulCrowder: Why are we using `null` here instead of `undefined`?

libs/components/lookup/testing/src/lookup/lookup-show-more-picker-harness.ts
```diff
@@ -0,0 +1,164 @@
...
+import { SkySearchHarness } from '../search/search-harness';
+
+import { SkyLookupShowMorePickerHarnessFilters } from './lookup-show-more-picker-harness-filters';
+import { SkyLookupShowMorePickerSearchResultHarness } from './lookup-show-more-picker-search-result-harness';
+import { SkyLookupShowMorePickerSearchResultHarnessFilters } from './lookup-show-more-picker-search-result-harness-filters';
+
+/**
+ * Harness for interacting with a lookup's "Show more" picker in tests.
+ */
+export class SkyLookupShowMorePickerHarness extends ComponentHarness {
```
Blackbaud-PaulCrowder, line 16: Does this work with both the `search` modal and the `searchAsync` modal? If not and we have to pick one, I'd pick the async version since that's the one we support going forward.

libs/components/lookup/src/lib/modules/lookup/lookup.component.spec.ts
Blackbaud-SteveBrush: Good call. I changed the expectation to `toBeTruthy()` since we run a11y checks in this spec and those IDs need to be unique.

libs/components/lookup/src/lib/modules/lookup/lookup.component.ts
Blackbaud-SteveBrush: Ugh, this is a bad habit I formed from an older version of Angular that required `null` to be provided to an attribute if you wanted the attribute removed from the element altogether. Updated to use `undefined` instead.

libs/components/lookup/testing/src/lookup/lookup-show-more-picker-harness.ts
Blackbaud-SteveBrush: It does, but I reworked the tests so that both the standard and async pickers run the same suite of tests so we can be 100% sure.

Blackbaud-PaulCrowder (approved):

PR #365, opened 2022-07-28
refactor(components/modals): enable modals strict mode and ensure 100% code coverage

Blackbaud-TrevorBurch:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/365/storybook/)

Component Storybooks:

- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/365/storybooks/modals-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/365/storybooks/tabs-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/365/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/365/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/365/playground/)

apps/code-examples/src/app/code-examples/modals/confirm/confirm-demo.component.ts
```diff
@@ -15,10 +15,14 @@ export class ConfirmDemoComponent {
 
   public selectedText: string | undefined;
 
-  constructor(private confirmService: SkyConfirmService) {}
+  #confirmService: SkyConfirmService;
```
Blackbaud-PaulCrowder: Since these are code examples I think it would be good to use a consistent naming convention for services. For instance, here we use `confirmService`, but in the modal demo we use `modal`. My suggestion would be to use `confirmSvc` and `modalSvc`.

libs/components/modals/src/lib/modules/confirm/confirm.component.spec.ts
```diff
@@ -49,7 +49,10 @@ describe('Confirm component', () => {
     // Cleanup after unit tests.
     // TODO: Figure out a better way to do this!
     const element = document.querySelector('.sky-confirm');
-    element.parentNode.removeChild(element);
+
+    if (element && element.parentNode) {
```
Blackbaud-PaulCrowder: You should be able to use `if (element?.parentNode)` here. You might even be able to use `element.parentNode?.removeChild(element)` and eliminate the `if` statement altogether (I can't remember if it's C# that allows this but TS doesn't or vice versa).

libs/components/modals/src/lib/modules/confirm/confirm.component.ts
```diff
@@ -8,58 +8,67 @@ import { SkyModalInstance } from '../modal/modal-instance';
...
+  ) {
+    this.#config = config;
+    this.#modal = modal;
+    this.#resourcesService = resourcesService!;
 
-  public ngOnInit() {
     if (
-      this.config.type === SkyConfirmType.Custom &&
-      this.config.buttons.length > 0
+      this.#config.type === SkyConfirmType.Custom &&
```
Blackbaud-PaulCrowder: Since you have `config` as a local parameter, I'd update references to `this.#config` to simply use `config`. It's (admittedly immeasurably) more efficient, and it will minify better.

libs/components/modals/src/lib/modules/confirm/confirm.component.ts
```diff
@@ -8,58 +8,67 @@ import { SkyModalInstance } from '../modal/modal-instance';
...
         this.buttons = buttons;
       });
     }
 
-    this.message = this.config.message;
-    this.body = this.config.body;
-    this.preserveWhiteSpace = this.config.preserveWhiteSpace || false;
+    this.message = this.#config.message;
+    this.body = this.#config.body;
+    this.preserveWhiteSpace = this.#config.preserveWhiteSpace || false;
```
Blackbaud-PaulCrowder: This could be `this.preserveWhiteSpace = !!config.preserveWhiteSpace;`

libs/components/modals/src/lib/modules/modal/modal-host.service.ts
```diff
@@ -1,5 +1,8 @@
 import { EventEmitter, Injectable } from '@angular/core';
 
+const BASE_Z_INDEX = 1040;
+const MODAL_HOSTS: SkyModalHostService[] = [];
```
Blackbaud-PaulCrowder: ALL_CAPS seems appropriate to me for constant primitive values, but for an array where we're adding and removing items it seems less appropriate. Perhaps change this one to be camelCase instead.

libs/components/modals/src/lib/modules/modal/modal.component.ts
```diff
@@ -35,69 +36,52 @@ let skyModalUniqueIdentifier = 0;
...
    */
   @Input()
-  public set tiledBody(value: boolean) {
-    this.config.tiledBody = value;
-  }
+  public tiledBody: boolean | undefined;
 
   public get modalZIndex() {
-    return this.hostService.getModalZIndex();
+    return this.#hostService.getModalZIndex();
```
Blackbaud-PaulCrowder: This seems wrong to me; can the z-index change during the lifecycle of the component, or is it constant once the modal is displayed? If it's the latter, we should get it once and assign it to a public property so we're not executing a function in a getter. If it's the former, we should use an Observable in the modal host service when it changes instead of relying on a change detection cycle to pick up the new value.

libs/components/modals/src/lib/modules/modal/modal.component.ts
```diff
@@ -107,30 +91,68 @@ export class SkyModalComponent implements AfterViewInit, OnDestroy {
...
+    this.ariaDescribedBy = config.ariaDescribedBy || this.modalContentId;
+    this.ariaLabelledBy = config.ariaLabelledBy || this.modalHeaderId;
+    this.ariaRole = config.ariaRole;
+    this.fullPage = config.fullPage;
+    this.helpKey = config.helpKey;
+    this.tiledBody = config.tiledBody;
+    this.wrapperClass = config.wrapperClass;
+
+    if (this.fullPage) {
+      this.isLargeSize = this.isMediumSize = this.isSmallSize = false;
```
Blackbaud-PaulCrowder: Is this line necessary? These properties already have a default value of `false`, so there's no need to set them to `false` again here. Alternatively I think it would be better to eliminate the three discrete properties and have one string union property since only one can ever be true at one time.

```
this.size = config.size?.toLowerCase() || '<whatever the default value is>';
```

libs/components/modals/src/lib/modules/modal/modal.component.ts
```diff
@@ -184,56 +206,56 @@ export class SkyModalComponent implements AfterViewInit, OnDestroy {
...
   public windowResize() {
-    this.componentAdapter.handleWindowChange(this.elRef);
+    this.#componentAdapter.handleWindowChange(this.#elRef);
   }
 
   public scrollShadowChange(args: SkyModalScrollShadowEventArgs): void {
     this.scrollShadow = args;
   }
 
   public viewkeeperEnabled(): boolean {
```
Blackbaud-PaulCrowder, line 237: This function is called from the template. Does this value change during the lifecycle of the component, and if so, should we be relying purely on external change detection triggers to notify the template? Again, this should be an Observable so the template can be notified when this changes if it's outside a change detection cycle.

libs/components/modals/src/lib/modules/modal/modal.service.ts
```diff
@@ -42,20 +47,26 @@ export class SkyModalService {
...
   ): SkyModalInstance {
     const modalInstance = new SkyModalInstance();
-    this.createHostComponent();
-    const params = this.getConfigFromParameter(config);
+    this.#createHostComponent();
+    const params = this.#getConfigFromParameter(config);
+
+    /* Ignoring coverage since, while it is good to have the null-check, the only current code path always has a default empty array given. */
+    /* istanbul ignore next */
+    if (!params.providers) {
```
Blackbaud-PaulCrowder: If this is true then I'd prefer to see a non-null assertion operator used instead of having code that never executes.

libs/components/modals/testing/src/modal-fixture.ts
```diff
@@ -4,53 +4,69 @@ import { ComponentFixture } from '@angular/core/testing';
...
-    const modalDialogElement = this.getModalDiaglogElement();
-    if (modalDialogElement) {
-      return modalDialogElement.getAttribute('aria-describedby') || undefined;
+  public get ariaDescribedBy(): string | undefined {
+    const modalDialogElement = this.#getModalDiaglogElement();
+    const describedByAttribute =
+      modalDialogElement.getAttribute('aria-describedby');
+    /* We need to have the else case here; however, it wouldn't ever be hit due to our defaults */
+    /* istanbul ignore else */
+    if (describedByAttribute) {
```
Blackbaud-PaulCrowder: I get the desire to move the error-throwing to one function, but it appears to confuse TypeScript and/or ESLint as to what the valid code paths are if you have to write code that is unreachable to satisfy it. I'd prefer continuing to throw the error directly in the property and using a constant for the repeated error message than to have to jump through hoops to fool the TS compiler.

apps/code-examples/src/app/code-examples/modals/confirm/confirm-demo.component.ts
Blackbaud-TrevorBurch: Done

libs/components/modals/src/lib/modules/confirm/confirm.component.spec.ts
Blackbaud-TrevorBurch: I can do `element?.parentNode?.removeChild(element)`. Done

libs/components/modals/src/lib/modules/confirm/confirm.component.ts
Blackbaud-TrevorBurch: Done

libs/components/modals/src/lib/modules/confirm/confirm.component.ts
Blackbaud-TrevorBurch: Done

libs/components/modals/src/lib/modules/modal/modal-host.service.ts
Blackbaud-TrevorBurch: Done

libs/components/modals/src/lib/modules/modal/modal.service.ts
Blackbaud-TrevorBurch: Ya. I went back and forth on this one. I switched it to the non-null assertion.

libs/components/modals/src/lib/modules/modal/modal.component.ts
Blackbaud-TrevorBurch: I made a change to where we have a `size` variable which includes `full-page`. This does change the class declarations in the template some so let me know if you want to chat any more about it.

libs/components/modals/src/lib/modules/modal/modal.component.ts
Blackbaud-TrevorBurch: I'm not sure I follow. This is a case where we are listening for an event and calling a function based on that. Not where we are binding to a value and calling a function. I'm not sure that I understand your suggestion on this one.

libs/components/modals/src/lib/modules/modal/modal.component.ts
Blackbaud-TrevorBurch: I have reworked the modal host service logic to utilize a `ReplaySubject` and then have the modal component listen to updates from this. This allows us to listen to changes but also update the zIndex of each host correctly if a lower modal closes for some reason while another was on top of it (not a common scenario - but one I wanted to make sure was accounted for). I also added a unit test for this scenario.

libs/components/modals/testing/src/modal-fixture.ts
Blackbaud-TrevorBurch: I don't think it is the error handling as much as it is the fact that we have defaults for the `aria-describedby` if it is undefined by the consumer. I think this is a case where I should have just used the non-null assertion. I'll update that and then we can chat more if needed.

libs/components/modals/src/lib/modules/modal/modal.component.ts
Blackbaud-PaulCrowder: I was referring to `viewkeeperEnabled()`. This line is in `modal.component.html`:

```
'sky-modal-viewkeeper': viewkeeperEnabled()
```

libs/components/modals/testing/src/modal-fixture.ts
Blackbaud-PaulCrowder: If you throw the error directly in the `if` statement, then the code coverage tool will see the `else` case as unreachable and therefore not require you cover it. By throwing an error in an external function, the code coverage tool can't determine that. That's why I'm advocating keeping the `throw` statement local to the function instead of calling a separate function to throw the error.

libs/components/modals/src/lib/modules/modal/modal.component.ts
Blackbaud-PaulCrowder: My first thought was just to reduce the complexity and get the z-index when the modal closes, but the current behavior does account for a lower modal being removed, so it's good that you preserved that behavior. I think you want a `BehaviorSubject` rather than a `ReplaySubject`, as `ReplaySubject` has an infinite buffer and will replay all previous values when a new subscriber is added, whereas `BehaviorSubject` only stores and emits the last value for new subscribers. You should also complete/close the subject when the modal host is destroyed so it doesn't hold onto old values indefinitely.

Blackbaud-TrevorBurch: We chatted in standup and decided to change this just to assign the zIndex when created based on the max of the already opened modals. I have updated things to do this.

libs/components/modals/src/lib/modules/modal/modal.component.ts
Blackbaud-PaulCrowder: We'll address in a future change.

Blackbaud-PaulCrowder (approved):

PR #386, opened 2022-08-09
fix(components/pages): action hub allow callback on needs attention items

johnhwhite:
[AB#2253841](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2253841)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/386/storybook/)

Component Storybooks:

- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/386/storybooks/pages-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/386/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/386/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/386/playground/)

libs/components/pages/src/lib/modules/action-hub/types/action-hub-needs-attention.ts
```diff
@@ -16,11 +17,22 @@ export interface SkyActionHubNeedsAttention {
...
     route?: {
       commands: any[];
       extras?: NavigationExtras;
     };
     url?: string;
   };
+  /**
+   * Specifies the modal parameters.
+   */
+  modal?: {
```
Blackbaud-PaulCrowder: Do we need the `modal` option if we have a `click` option? I feel like having a specific `modal` option is going to have us chasing modal every time it adds a new feature to maintain parity here. As it is now it doesn't seem to provide the consumer with the modal instance for handling save/cancel/etc.

libs/components/pages/src/lib/modules/action-hub/types/action-hub-needs-attention.ts
```diff
@@ -16,11 +17,22 @@ export interface SkyActionHubNeedsAttention {
...
+    component: any;
+    config?: SkyModalConfigurationInterface;
+  };
+  /**
+   * Specifies a click handler for the action item.
+   */
+  click?: () => void;
```
Blackbaud-PaulCrowder, line 29: Should we pass some kind of context to this function so the consumer knows which item is clicked?

libs/components/pages/src/lib/modules/link-list/link-list.component.html
```diff
@@ -28,7 +28,16 @@ <h2 class="sky-font-heading-4 sky-margin-stacked-sm">
           'sky-margin-stacked-sm': 'modern'
         }"
       >
-        <a *ngIf="link.permalink.url" [skyHref]="link.permalink.url">
+        <a
+          *ngIf="link.permalink.url && hasProtocol(link.permalink.url)"
```
Blackbaud-PaulCrowder: Since this is in an `ngFor` I understand need to make an exception to the "don't call functions in templates" rule, but it should use a pure pipe instead of a function call. That way it will only be evaluated when the input value changes instead of every change detection cycle.

libs/components/pages/src/lib/modules/needs-attention/needs-attention.component.scss
```diff
@@ -9,30 +9,27 @@
   display: flex;
   flex-wrap: nowrap;
   align-items: center;
+  cursor: pointer;
```
Blackbaud-PaulCrowder: Is `cursor: pointer` necessary? Both the default style for `a` and `.sky-btn` already specify this.

libs/components/pages/src/lib/modules/needs-attention/needs-attention.component.scss
```diff
@@ -9,30 +9,27 @@
...
+  text-align: left;
+  width: 100%;
```
Blackbaud-PaulCrowder, line 13: I'm not sure you need this, either, if `display` is `flex`.

libs/components/pages/src/lib/modules/needs-attention/needs-attention.component.scss
```diff
@@ -9,30 +9,27 @@
...
-  }
-
-  sky-key-info-value {
     color: $sky-text-color-action-primary;
-  }
 
-  &:hover {
-    sky-key-info-value,
-    sky-key-info-label {
+    &:hover {
```
Blackbaud-PaulCrowder: Another style that should already be specified on `a` and `.sky-btn`.

libs/components/pages/src/lib/modules/action-hub/types/action-hub-needs-attention.ts
johnhwhite: Removed the `modal` option here.

libs/components/pages/src/lib/modules/action-hub/types/action-hub-needs-attention.ts
johnhwhite: I'm not sure what would be needed in the context that they don't already have. I'm inclined to keep the interface simple.

libs/components/pages/src/lib/modules/link-list/link-list.component.html
johnhwhite: Switched to pipes.

libs/components/pages/src/lib/modules/needs-attention/needs-attention.component.scss
johnhwhite: This ended up being an issue where `skyHref` no longer allowed `#` as a link, so the link was styled as non-active.

libs/components/pages/src/lib/modules/needs-attention/needs-attention.component.scss
johnhwhite: The `text-align` and `width` styles are necessary for the button.

Blackbaud-PaulCrowder: I only mentioned `width` because `flex` has a default width of `100%`. Is here another class on these items that counteracts that default?

libs/components/pages/src/lib/modules/pipes/use-href.pipe.ts
```diff
@@ -0,0 +1,15 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'useHref',
+})
+export class UseHrefPipe implements PipeTransform {
```
Blackbaud-PaulCrowder: Can these pipes all be combined into one? it could either take a second parameter as the type you want to check (href, skyAppLink, etc.) and return a boolean, or you could keep the one parameter and return a string indicating which one to use.

johnhwhite: Consolidated to one pipe. Good call.

libs/components/pages/src/lib/modules/needs-attention/needs-attention.component.scss
johnhwhite: If it were `display: block flex` it would have default width `100%` but `button` has a default `<display-outside>` of `inline` and [MDN says it's too soon](https://developer.mozilla.org/en-US/docs/Web/CSS/display#which_syntax_should_you_use_now) to adopt the newer syntax.

libs/components/pages/src/lib/modules/link-as/link-as.pipe.ts
```diff
@@ -0,0 +1,39 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'linkAs',
+})
+export class LinkAsPipe implements PipeTransform {
+  public transform(
+    value: unknown,
```
Blackbaud-PaulCrowder: It seems a little odd to accept `item` or `item.permalink` here and then have to use `uknown` as the type. Can you not always accept `item` as `SkyActionHubNeedsAttention`? That way you can check for `permalink` in the function and avoid all the bracket notation and `'prop' in value` checks.

johnhwhite: 💡 Updated.

Blackbaud-PaulCrowder (approved):

PR #397, opened 2022-08-15
fix(components/modals): hide non modal elements from screen readers

Blackbaud-SandhyaRajasabeson:
add aria-hidden to siblings of the current modal and modal host so that screen readers cannot
navigate outside the modal while it is open

### DIFF
--- a/libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
+++ b/libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
@@ -13,7 +13,7 @@ export class SkyModalAdapterService {
   #bodyEl: HTMLElement;
 
   #windowRef: SkyAppWindowRef;
-  #hostSiblingPreviousValue = new Map<Element, string | null>();
+  #hostSiblingAriaHiddenCache = new Map<Element, string | null>();
 
   constructor(windowRef: SkyAppWindowRef) {
     this.#windowRef = windowRef;
@@ -49,7 +49,11 @@ export class SkyModalAdapterService {
     this.#bodyEl.classList.remove(className);
   }
 
-  public hideHostSiblings(hostElRef: ElementRef): void {
+  /**
+   * Hides siblings of modal-host from screen readers
+   * @param hostElRef reference to modal-host element
+   */
+  public hideHostSiblingsFromScreenReaders(hostElRef: ElementRef): void {
     const hostElement = hostElRef.nativeElement;
     const hostSiblings = hostElement.parentElement.children;
 
@@ -62,7 +66,7 @@ export class SkyModalAdapterService {
         element.nodeName.toLowerCase() !== 'style'
       ) {
         // preserve previous aria-hidden status of elements outside of modal host
-        this.#hostSiblingPreviousValue.set(
+        this.#hostSiblingAriaHiddenCache.set(
           element,
           element.getAttribute('aria-hidden')
         );
@@ -71,8 +75,11 @@ export class SkyModalAdapterService {
     }
   }
 
-  public unhideOrRestoreHostSiblings(): void {
-    this.#hostSiblingPreviousValue.forEach((previousValue, element) => {
+  /**
+   * Restores modal-host siblings to screenreader status prior to modals being opened
+   */
+  public unhideOrRestoreHostSiblingsFromScreenReaders(): void {
+    this.#hostSiblingAriaHiddenCache.forEach((previousValue, element) => {
       // if element had aria-hidden status prior, restore status
       if (element.parentElement) {
         if (previousValue) {
@@ -82,16 +89,16 @@ export class SkyModalAdapterService {
         }
       }
     });
-    this.#hostSiblingPreviousValue.clear();
+    this.#hostSiblingAriaHiddenCache.clear();
   }
 
-  public hidePreviousModal(topModal: Element): void {
+  public hidePreviousModalFromScreenReaders(topModal: Element): void {
     if (topModal && topModal.previousElementSibling) {
       topModal.previousElementSibling.setAttribute('aria-hidden', 'true');
     }
   }
 
-  public unhidePreviousModal(topModal: Element): void {
+  public unhidePreviousModalFromScreenReaders(topModal: Element): void {
     if (topModal && topModal.previousElementSibling) {
       topModal.previousElementSibling.removeAttribute('aria-hidden');
     }
--- a/libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
+++ b/libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
@@ -13,7 +13,7 @@ export class SkyModalAdapterService {
   #bodyEl: HTMLElement;
 
   #windowRef: SkyAppWindowRef;
-  #hiddenElements = new Map<Element, string | null>();
+  #hostSiblingPreviousValue = new Map<Element, string | null>();
 
   constructor(windowRef: SkyAppWindowRef) {
     this.#windowRef = windowRef;
@@ -49,8 +49,8 @@ export class SkyModalAdapterService {
     this.#bodyEl.classList.remove(className);
   }
 
-  public hideHostSiblings(elRef: ElementRef): void {
-    const hostElement = elRef.nativeElement;
+  public hideHostSiblings(hostElRef: ElementRef): void {
+    const hostElement = hostElRef.nativeElement;
     const hostSiblings = hostElement.parentElement.children;
 
     for (let i = 0; i < hostSiblings.length; i++) {
@@ -62,14 +62,17 @@ export class SkyModalAdapterService {
         element.nodeName.toLowerCase() !== 'style'
       ) {
         // preserve previous aria-hidden status of elements outside of modal host
-        this.#hiddenElements.set(element, element.getAttribute('aria-hidden'));
+        this.#hostSiblingPreviousValue.set(
+          element,
+          element.getAttribute('aria-hidden')
+        );
         element.setAttribute('aria-hidden', 'true');
       }
     }
   }
 
-  public unhideHostSiblings(): void {
-    this.#hiddenElements.forEach((previousValue, element) => {
+  public unhideOrRestoreHostSiblings(): void {
+    this.#hostSiblingPreviousValue.forEach((previousValue, element) => {
       // if element had aria-hidden status prior, restore status
       if (previousValue) {
         element.setAttribute('aria-hidden', previousValue);
@@ -77,18 +80,18 @@ export class SkyModalAdapterService {
         element.removeAttribute('aria-hidden');
       }
     });
-    this.#hiddenElements.clear();
+    this.#hostSiblingPreviousValue.clear();
   }
 
-  public hidePreviousModal(modal: Element): void {
-    if (modal && modal.previousElementSibling) {
-      modal.previousElementSibling.setAttribute('aria-hidden', 'true');
+  public hidePreviousModal(topModal: Element): void {
+    if (topModal && topModal.previousElementSibling) {
+      topModal.previousElementSibling.setAttribute('aria-hidden', 'true');
     }
   }
 
-  public unhidePreviousModal(modal: Element): void {
-    if (modal && modal.previousElementSibling) {
-      modal.previousElementSibling.removeAttribute('aria-hidden');
+  public unhidePreviousModal(topModal: Element): void {
+    if (topModal && topModal.previousElementSibling) {
+      topModal.previousElementSibling.removeAttribute('aria-hidden');
     }
   }
 }
--- a/libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
+++ b/libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
@@ -74,10 +74,12 @@ export class SkyModalAdapterService {
   public unhideOrRestoreHostSiblings(): void {
     this.#hostSiblingPreviousValue.forEach((previousValue, element) => {
       // if element had aria-hidden status prior, restore status
-      if (previousValue) {
-        element.setAttribute('aria-hidden', previousValue);
-      } else {
-        element.removeAttribute('aria-hidden');
+      if (element.parentElement) {
+        if (previousValue) {
+          element.setAttribute('aria-hidden', previousValue);
+        } else {
+          element.removeAttribute('aria-hidden');
+        }
       }
     });
     this.#hostSiblingPreviousValue.clear();
--- a/libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
+++ b/libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
@@ -47,4 +47,36 @@ export class SkyModalAdapterService {
   #removeClassFromBody(className: string): void {
     this.#bodyEl.classList.remove(className);
   }
+
+  public addAriaHidden(
+    elementsToHide: HTMLCollection,
+    self: Element | unknown
+  ): Map<Element, string | null> {
+    const hiddenElements = new Map<Element, string | null>();
+
+    for (let i = 0; i < elementsToHide.length; i++) {
+      const element = elementsToHide[i];
+      if (
+        element !== self &&
+        !element.hasAttribute('aria-live') &&
+        element.nodeName.toLowerCase() !== 'script' &&
+        element.nodeName.toLowerCase() !== 'style'
+      ) {
+        hiddenElements.set(element, element.getAttribute('aria-hidden'));
+        element.setAttribute('aria-hidden', 'true');
+      }
+    }
+
+    return hiddenElements;
+  }
+
+  public removeAriaHidden(hiddenElements: Map<Element, string | null>): void {
+    hiddenElements.forEach((previousValue, element) => {
+      if (previousValue) {
+        element.setAttribute('aria-hidden', previousValue);
+      } else {
+        element.removeAttribute('aria-hidden');
+      }
+    });
+  }
 }
--- a/libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
+++ b/libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
@@ -92,15 +92,20 @@ export class SkyModalAdapterService {
     this.#hostSiblingAriaHiddenCache.clear();
   }
 
-  public hidePreviousModalFromScreenReaders(topModal: Element): void {
-    if (topModal && topModal.previousElementSibling) {
-      topModal.previousElementSibling.setAttribute('aria-hidden', 'true');
+  public hidePreviousModalFromScreenReaders(topModal: ElementRef): void {
+    if (topModal && topModal.nativeElement.previousElementSibling) {
+      topModal.nativeElement.previousElementSibling.setAttribute(
+        'aria-hidden',
+        'true'
+      );
     }
   }
 
-  public unhidePreviousModalFromScreenReaders(topModal: Element): void {
-    if (topModal && topModal.previousElementSibling) {
-      topModal.previousElementSibling.removeAttribute('aria-hidden');
+  public unhidePreviousModalFromScreenReaders(topModal: ElementRef): void {
+    if (topModal && topModal.nativeElement.previousElementSibling) {
+      topModal.nativeElement.previousElementSibling.removeAttribute(
+        'aria-hidden'
+      );
     }
   }
 }
--- a/libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
+++ b/libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
@@ -62,6 +62,7 @@ export class SkyModalAdapterService {
         element.nodeName.toLowerCase() !== 'script' &&
         element.nodeName.toLowerCase() !== 'style'
       ) {
+        // preserve previous aria-hidden status of elements outside of modal host
         hiddenElements.set(element, element.getAttribute('aria-hidden'));
         element.setAttribute('aria-hidden', 'true');
       }
@@ -72,6 +73,7 @@ export class SkyModalAdapterService {
 
   public removeAriaHidden(hiddenElements: Map<Element, string | null>): void {
     hiddenElements.forEach((previousValue, element) => {
+      // if element had aria-hidden status prior, restore status
       if (previousValue) {
         element.setAttribute('aria-hidden', previousValue);
       } else {
--- a/libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
+++ b/libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
@@ -1,4 +1,4 @@
-import { Injectable } from '@angular/core';
+import { ElementRef, Injectable } from '@angular/core';
 import { SkyAppWindowRef } from '@skyux/core';
 
 /**
@@ -13,6 +13,7 @@ export class SkyModalAdapterService {
   #bodyEl: HTMLElement;
 
   #windowRef: SkyAppWindowRef;
+  #hiddenElements = new Map<Element, string | null>();
 
   constructor(windowRef: SkyAppWindowRef) {
     this.#windowRef = windowRef;
@@ -48,31 +49,27 @@ export class SkyModalAdapterService {
     this.#bodyEl.classList.remove(className);
   }
 
-  public addAriaHidden(
-    elementsToHide: HTMLCollection,
-    self: Element | unknown
-  ): Map<Element, string | null> {
-    const hiddenElements = new Map<Element, string | null>();
+  public hideHostSiblings(elRef: ElementRef): void {
+    const hostElement = elRef.nativeElement;
+    const hostSiblings = hostElement.parentElement.children;
 
-    for (let i = 0; i < elementsToHide.length; i++) {
-      const element = elementsToHide[i];
+    for (let i = 0; i < hostSiblings.length; i++) {
+      const element = hostSiblings[i];
       if (
-        element !== self &&
+        element !== hostElement &&
         !element.hasAttribute('aria-live') &&
         element.nodeName.toLowerCase() !== 'script' &&
         element.nodeName.toLowerCase() !== 'style'
       ) {
         // preserve previous aria-hidden status of elements outside of modal host
-        hiddenElements.set(element, element.getAttribute('aria-hidden'));
+        this.#hiddenElements.set(element, element.getAttribute('aria-hidden'));
         element.setAttribute('aria-hidden', 'true');
       }
     }
-
-    return hiddenElements;
   }
 
-  public removeAriaHidden(hiddenElements: Map<Element, string | null>): void {
-    hiddenElements.forEach((previousValue, element) => {
+  public unhideHostSiblings(): void {
+    this.#hiddenElements.forEach((previousValue, element) => {
       // if element had aria-hidden status prior, restore status
       if (previousValue) {
         element.setAttribute('aria-hidden', previousValue);
@@ -80,5 +77,18 @@ export class SkyModalAdapterService {
         element.removeAttribute('aria-hidden');
       }
     });
+    this.#hiddenElements.clear();
+  }
+
+  public hidePreviousModal(modal: Element): void {
+    if (modal && modal.previousElementSibling) {
+      modal.previousElementSibling.setAttribute('aria-hidden', 'true');
+    }
+  }
+
+  public unhidePreviousModal(modal: Element): void {
+    if (modal && modal.previousElementSibling) {
+      modal.previousElementSibling.removeAttribute('aria-hidden');
+    }
   }
 }
--- a/libs/components/modals/src/lib/modules/modal/modal-host.component.ts
+++ b/libs/components/modals/src/lib/modules/modal/modal-host.component.ts
@@ -2,6 +2,7 @@ import {
   ChangeDetectorRef,
   Component,
   ComponentFactoryResolver,
+  ElementRef,
   Injector,
   OnDestroy,
   ViewChild,
@@ -57,8 +58,10 @@ export class SkyModalHostComponent implements OnDestroy {
   #injector: Injector;
   #router: Router;
   #changeDetector: ChangeDetectorRef;
-
   #modalHostContext: SkyModalHostContext;
+  #elRef: ElementRef;
+
+  #ariaPreviousValueSiblings = new Map<Element, string | null>();
 
   #modalInstances: SkyModalInstance[] = [];
 
@@ -68,7 +71,8 @@ export class SkyModalHostComponent implements OnDestroy {
     injector: Injector,
     router: Router,
     changeDetector: ChangeDetectorRef,
-    modalHostContext: SkyModalHostContext
+    modalHostContext: SkyModalHostContext,
+    elRef: ElementRef
   ) {
     this.#resolver = resolver;
     this.#adapter = adapter;
@@ -76,12 +80,23 @@ export class SkyModalHostComponent implements OnDestroy {
     this.#router = router;
     this.#changeDetector = changeDetector;
     this.#modalHostContext = modalHostContext;
+    this.#elRef = elRef;
   }
 
   public ngOnDestroy(): void {
     // Close all modal instances before disposing of the host container.
     this.#closeAllModalInstances();
     this.#modalHostContext.args.teardownCallback();
+  }
+
+  public hideModalHostSiblings(): void {
+    // hidding all elements at the modal-host level from screenreaders
+    const hostElement = this.#elRef.nativeElement;
+    const hostSiblings = hostElement.parentElement.children;
+    this.#ariaPreviousValueSiblings = this.#adapter.addAriaHidden(
+      hostSiblings,
+      hostElement
+    );
   }
 
   public open(
@@ -104,6 +119,7 @@ export class SkyModalHostComponent implements OnDestroy {
 
     const adapter = this.#adapter;
     const modalOpener: HTMLElement = adapter.getModalOpener();
+    let ariaPreviousValueSiblings = this.#ariaPreviousValueSiblings;
 
     let isOpen = true;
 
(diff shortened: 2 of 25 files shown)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/397/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/397/storybooks/ag-grid-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/397/storybooks/modals-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/397/storybooks/pages-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/397/storybooks/tabs-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/397/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/397/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/397/playground/)

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -14,6 +15,7 @@ import {
 } from '@skyux/core';
 
 import { takeWhile } from 'rxjs/operators';
+import { __values } from 'tslib';
```
Blackbaud-SteveBrush: Eh?

libs/components/modals/src/lib/modules/modal/modal.component.ts
```diff
@@ -25,7 +25,7 @@ import { SkyModalScrollShadowEventArgs } from './modal-scroll-shadow-event-args'
 import { skyAnimationModalState } from './modal-state-animation';
 
 let skyModalUniqueIdentifier = 0;
-
+//get modal host siblings and put either aria hidden or display hidden in css
```
Blackbaud-TrevorBurch: This comment seems to be for nothing. Am I missing something?

Blackbaud-SteveBrush: Should this comment be here?

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -109,15 +117,89 @@ export class SkyModalHostComponent implements OnDestroy {
...
+      this.hostElement.nativeElement.parentElement.children
+    );
+
     const modalComponentRef = this.target.createComponent(
       factory,
       undefined,
       injector
     );
 
+    //const modalsAfter = this.hostElement.nativeElement.children;
```
Blackbaud-TrevorBurch: Extra commented out line?

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -109,15 +117,89 @@ export class SkyModalHostComponent implements OnDestroy {
       parent: this.injector,
     });
 
+    const modalsBefore = Array.from(
+      this.hostElement.nativeElement.parentElement.children
```
Blackbaud-TrevorBurch: Is there a reason that we can't  just use `this.target.element.nativeElement` here? It feels odd that we are adding another element just for grabbing the parent element when we already have another sibling element.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -109,15 +117,89 @@ export class SkyModalHostComponent implements OnDestroy {
...
+    const modalElement = modalsAfter.filter(
+      (el) => modalsBefore.indexOf(el) < 0
+    )[0];
+
     modalInstance.componentInstance = modalComponentRef.instance;
 
+    // hidding all elements at the modal-host level from screenreaders
+    const ariaPreviousValueSiblings = new Map<Element, string | null>();
+    const siblings =
+      document.querySelector('sky-modal-host').parentElement.children;
```
Blackbaud-TrevorBurch: Any logic that is grabbing things off the actual DOM API should go into an adapter service. So in this case they should go into the `SkyModalAdapterService`.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -109,15 +117,89 @@ export class SkyModalHostComponent implements OnDestroy {
...
+
+    for (let i = 0; i < siblings.length; i++) {
+      const sibling = siblings[i];
+
+      if (
+        sibling !== document.querySelector('sky-modal-host') &&
+        !sibling.hasAttribute('aria-live') &&
+        sibling.nodeName !== 'SCRIPT' &&
```
Blackbaud-TrevorBurch: Should we do a `sibling.nodeName.toLowerCase()` just to be safe and then use 'script'? That way if some browser decides to return 'script' we will handle it.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -109,15 +117,89 @@ export class SkyModalHostComponent implements OnDestroy {
...
+          sibling.getAttribute('aria-hidden')
+        );
+
+        sibling.setAttribute('aria-hidden', 'true');
+      }
+    }
+
+    // hiding all sibling modals from screenreaders
+    const ariaPreviousValueChildren = new Map<Element, string | null>();
+    const children = document.querySelector('sky-modal-host').children;
```
Blackbaud-TrevorBurch: Nitpicky - could we call this `hostChildren` and then the later variable `hostChild` just to make it clearer what children we have?

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -109,15 +117,89 @@ export class SkyModalHostComponent implements OnDestroy {
...
+    // getting the element ref to the modal that was just opened
+    const modalElement = modalsAfter.filter(
+      (el) => modalsBefore.indexOf(el) < 0
+    )[0];
+
     modalInstance.componentInstance = modalComponentRef.instance;
 
+    // hidding all elements at the modal-host level from screenreaders
+    const ariaPreviousValueSiblings = new Map<Element, string | null>();
+    const siblings =
```
Blackbaud-TrevorBurch: Nitpicky again - could we call this `hostSiblings`? It took me a minute to figure out what we were grabbing here.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -109,15 +117,89 @@ export class SkyModalHostComponent implements OnDestroy {
...
+      document.querySelector('sky-modal-host').parentElement.children;
+
+    for (let i = 0; i < siblings.length; i++) {
+      const sibling = siblings[i];
+
+      if (
+        sibling !== document.querySelector('sky-modal-host') &&
```
Blackbaud-TrevorBurch: Can we inject `ElementRef` into the constructor of this? Then we could do `sibling !== this.elRef.nativeElement` and don't have to query the DOM again.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -109,15 +117,89 @@ export class SkyModalHostComponent implements OnDestroy {
...
+        !sibling.hasAttribute('aria-live') &&
```
Blackbaud-TrevorBurch: For my own knowledge - can you explain why we aren't placing `aria-hidden` on elements which have `aria-live`?

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -109,15 +117,89 @@ export class SkyModalHostComponent implements OnDestroy {
...
       injector
     );
 
+    //const modalsAfter = this.hostElement.nativeElement.children;
+    const modalsAfter = Array.from(
+      this.hostElement.nativeElement.parentElement.children
+    );
+
+    // getting the element ref to the modal that was just opened
+    const modalElement = modalsAfter.filter(
```
Blackbaud-TrevorBurch: It seems like the only reason we are querying the modals before and after are to get the new opened item. However, we should be able to get that by using `modalComponentRef.location.nativeElement`

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -109,15 +117,89 @@ export class SkyModalHostComponent implements OnDestroy {
...
+        child.nodeName !== 'STYLE'
+      ) {
+        ariaPreviousValueChildren.set(child, child.getAttribute('aria-hidden'));
+        child.setAttribute('aria-hidden', 'true');
+      }
+    }
+
     function closeModal() {
+      // unhiding modal-host siblings
+      ariaPreviousValueSiblings.forEach((previousValue, element) => {
```
Blackbaud-TrevorBurch: What happens if a lower modal closes before the most upper modal (not a common case - but could happen via the typescript service technically.

Blackbaud-TrevorBurch: I think that the `children` would most likely be ok as the lower modal wouldn't have the new `children` and so it probably would be fine. However, if I'm reading this correctly - closing the lowest modal would reenable all siblings even though an upper modal is still open.

Blackbaud-TrevorBurch (changes requested):
@Blackbaud-SandhyaRajasabeson - some comments for you.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
Blackbaud-SandhyaRajasabeson: i was basing this off of how angular handles `aria-hidden` with their modals. if we put aria-hidden on aria-live elements they won't announce and won't announce even after the modal is removed so it would be information that never gets to the user.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
Blackbaud-SandhyaRajasabeson: im going to add a unit test that opens two modals and then closes the first one just to confirm that that is what would happen. but if that is so, im thinking make the list of aria hidden modals into a global variable and check that when any modal is closed.

libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
```diff
@@ -47,4 +47,36 @@ export class SkyModalAdapterService {
...
+
+    for (let i = 0; i < elementsToHide.length; i++) {
+      const element = elementsToHide[i];
+      if (
+        element !== self &&
+        !element.hasAttribute('aria-live') &&
+        element.nodeName.toLowerCase() !== 'script' &&
+        element.nodeName.toLowerCase() !== 'style'
+      ) {
+        hiddenElements.set(element, element.getAttribute('aria-hidden'));
```
Blackbaud-SteveBrush: Mind putting a comment above this line to explain why we're saving the existing `aria-hidden` value? I know why (because we talked about it during Flex) but it would be good to mention it here for posterity.

libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
```diff
@@ -47,4 +47,36 @@ export class SkyModalAdapterService {
...
+        element.setAttribute('aria-hidden', 'true');
+      }
+    }
+
+    return hiddenElements;
+  }
+
+  public removeAriaHidden(hiddenElements: Map<Element, string | null>): void {
+    hiddenElements.forEach((previousValue, element) => {
+      if (previousValue) {
```
Blackbaud-SteveBrush: I would add a comment above this line explaining why there's a condition. Something to the effect of 
```
// Restore the initial value of `aria-hidden` if applicable.
```

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -139,11 +155,38 @@ export class SkyModalHostComponent implements OnDestroy {
...
     modalInstance.componentInstance = modalComponentRef.instance;
 
     this.#registerModalInstance(modalInstance);
+    // hidding all elements at the modal-host level from screenreaders
+    const hostElement = this.#elRef.nativeElement;
+    let allModals = hostElement.children;
+
+    if (SkyModalHostService.openModalCount == 1) {
+      this.hideModalHostSiblings();
+      ariaPreviousValueSiblings = this.#ariaPreviousValueSiblings;
```
Blackbaud-SteveBrush: What's the value in setting this again? Isn't it getting set up above? 

More importantly, I believe you're creating a local copy of the `#ariaPreviousValueSiblings` class property because you can't use `this` in the `closeModal` function. I would instead convert the `closeModal` function into a "fat arrow" function and just use the `this.#ariaPreviousValueSiblings` property directly, so that we can avoid adding needless "shadow" variables.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -139,11 +155,38 @@ export class SkyModalHostComponent implements OnDestroy {
...
+    const modalElement = modalComponentRef.location.nativeElement;
+
     modalInstance.componentInstance = modalComponentRef.instance;
 
     this.#registerModalInstance(modalInstance);
+    // hidding all elements at the modal-host level from screenreaders
+    const hostElement = this.#elRef.nativeElement;
+    let allModals = hostElement.children;
+
+    if (SkyModalHostService.openModalCount == 1) {
```
Blackbaud-SteveBrush: I could be wrong, but why would we check if there's only one modal opened before hiding all siblings? What happens if an element is added to the BODY tag when a second modal has been opened? Shouldn't we hide any newly created siblings when that happens?

libs/components/modals/src/lib/modules/modal/modal.service.spec.ts
```diff
@@ -176,9 +176,137 @@ describe('Modal service', () => {
...
+    document.body.appendChild(div);
+    div.setAttribute('aria-live', 'true');
+
+    openModal(ModalTestComponent, { fullPage: false });
+
+    expect(div.getAttribute('aria-hidden')).toBe(null);
+    div.remove();
+  }));
+
+  it('what does it do with layered modals', fakeAsync(() => {
```
Blackbaud-SteveBrush: All `it` blocks should start with the word "should". You also have two tests named the same thing.

libs/components/modals/src/lib/modules/modal/modal.service.spec.ts
```diff
@@ -176,9 +176,137 @@ describe('Modal service', () => {
...
+
+    expect(div.getAttribute('aria-hidden')).toBe('true');
+
+    closeModal(modal);
+
+    expect(div.getAttribute('aria-hidden')).toBe(null);
+    div.remove();
+  }));
+
+  it('should keep hidden status elements at the host level consistent', fakeAsync(() => {
```
Blackbaud-SteveBrush: For clarity, I would reword the test to something like, "should restore sibling elements' initial 'aria-hidden' values after all modals are closed".

libs/components/modals/src/lib/modules/modal/modal.service.spec.ts
```diff
@@ -176,9 +176,137 @@ describe('Modal service', () => {
...
     const wrapperClass = 'custom-wrapper-class';
-    openModal(ModalTestComponent, { wrapperClass });
+    const modal = openModal(ModalTestComponent, { wrapperClass });
     applicationRef.tick();
 
     expect(document.body.querySelector(`sky-modal.${wrapperClass}`)).toExist();
+    closeModal(modal);
+  }));
+
+  it('should not hide modal host from screen readers when opening a modal', fakeAsync(() => {
```
Blackbaud-SteveBrush: It might be beneficial to wrap these tests in their own `describe` block so they're easier to digest.

libs/components/modals/src/lib/modules/modal/modal.service.spec.ts
```diff
@@ -176,9 +176,137 @@ describe('Modal service', () => {
...
+  }));
+
+  it('what does it do with layered modals', fakeAsync(() => {
+    openModal(ModalTestComponent, { fullPage: false });
+    document.querySelector('sky-test-cmp').id = 'sibling';
+    const host = document.querySelector('sky-modal-host');
+
+    const lowerModal = openModal(ModalTestComponent, { fullPage: false });
+    const list = host.children;
+    list[3].id = 'lowerModal';
```
Blackbaud-SteveBrush: I was initially very confused that the host had more than 2 children (since you only opened two modals in this test). Once I looked in the host's template I could see that we have a few child elements baked-in. That being said, simply reading the host's children can be a bit brittle (what happens when we decide to add another element to the host's template?). Instead, I would create a helper function `getModalEls()` that would return a `NodeListOf<Element>` for all modal elements so that future contributors to this test are not confused.
```
function getModalEls() {
  return document.querySelectorAll('sky-test-cmp');
}
```
Then, in your tests, you can do the following:
```
const modalEls = getModalEls();
modalEls.item(0).blah
```

libs/components/modals/src/lib/modules/modal/modal.service.spec.ts
```diff
@@ -176,9 +176,137 @@ describe('Modal service', () => {
...
+
+    openModal(ModalTestComponent, { fullPage: false });
+
+    expect(div.getAttribute('aria-hidden')).toBe(null);
+    div.remove();
+  }));
+
+  it('what does it do with layered modals', fakeAsync(() => {
+    openModal(ModalTestComponent, { fullPage: false });
+    document.querySelector('sky-test-cmp').id = 'sibling';
```
Blackbaud-SteveBrush: Can we add more descriptive IDs than "sibling" and "lowerModal"? I was having trouble tracking the logic of the test with those names. Also, the "lowerModal" is opened after the "sibling" modal; shouldn't it be the "upperModal" instead?

libs/components/modals/src/lib/modules/modal/modal.service.spec.ts
```diff
@@ -176,9 +176,137 @@ describe('Modal service', () => {
...
+    expect(
+      document.getElementById('lowerModal').getAttribute('aria-hidden')
+    ).toBe(null);
+  }));
+
+  // modal siblings
+
+  it('should hide and unhide modal siblings from screen readers', fakeAsync(() => {
+    openModal(ModalTestComponent, { fullPage: false });
+    document.querySelector('sky-test-cmp').id = 'sibling';
```
Blackbaud-SteveBrush: Any place where you're querying for modals, you can use the new `getModalEls()` function (see above).

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
Blackbaud-SteveBrush: (I would add a unit test that predicts that exact scenario to confirm.)

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -139,11 +155,38 @@ export class SkyModalHostComponent implements OnDestroy {
...
+      this.hideModalHostSiblings();
+      ariaPreviousValueSiblings = this.#ariaPreviousValueSiblings;
+    }
+    if (
+      SkyModalHostService.openModalCount > 1 &&
+      allModals[allModals.length - 1] == modalElement
+    ) {
+      allModals[allModals.length - 2].setAttribute('aria-hidden', true);
+    }
```
Blackbaud-SteveBrush, line 160: I would use either `nextElementSibling` or `previousElementSibling` for this. https://developer.mozilla.org/en-US/docs/Web/API/Element/previousElementSibling

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -139,11 +155,38 @@ export class SkyModalHostComponent implements OnDestroy {
...
 
     function closeModal() {
+      // unhide siblings if last modal is closing
+      if (SkyModalHostService.openModalCount == 1) {
+        adapter.removeAriaHidden(ariaPreviousValueSiblings);
+        ariaPreviousValueSiblings.clear();
+      } else if (SkyModalHostService.topModal == hostService) {
+        // if there are more than 1 modal then unhide the one behind this one before closing it
+        allModals = hostElement.children;
+        allModals[allModals.length - 2].removeAttribute('aria-hidden');
```
Blackbaud-SteveBrush: Same here, RE: `nextElementSibling` or `previousElementSibling`.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -139,11 +155,38 @@ export class SkyModalHostComponent implements OnDestroy {
...
+      allModals[allModals.length - 1] == modalElement
+    ) {
+      allModals[allModals.length - 2].setAttribute('aria-hidden', true);
+    }
 
     function closeModal() {
+      // unhide siblings if last modal is closing
+      if (SkyModalHostService.openModalCount == 1) {
+        adapter.removeAriaHidden(ariaPreviousValueSiblings);
+        ariaPreviousValueSiblings.clear();
```
Blackbaud-SteveBrush: This feels like a wholistic action, so I would put this in the host's `ngOnDestroy()` method.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -57,8 +58,10 @@ export class SkyModalHostComponent implements OnDestroy {
   #injector: Injector;
   #router: Router;
   #changeDetector: ChangeDetectorRef;
-
   #modalHostContext: SkyModalHostContext;
+  #elRef: ElementRef;
+
+  #ariaPreviousValueSiblings = new Map<Element, string | null>();
```
Blackbaud-SteveBrush: Since the `SkyModalAdapterService` is provided by this component, its state will be tied to this component as well. Meaning, the adapter service could manage all of the sibling elements itself. This would also keep the modal component from being "polluted" with DOM features and types (the adapter service is designed to abstract away the DOM-specific aspects of our component class).

If possible, I would move the `#ariaPreviousValueSiblings` property to the adapter service and have it manage the `aria-hidden` states and elements itself. The only thing you should be calling in this component would be:
```
this.#adapter.hideHostSiblingsForScreenReaders(this.#elRef);
this.#adapter.showHostSiblingsForScreenReaders(this.#elRef);

this.#adapter.hidePreviousModalForScreenReaders();
this.#adapter.showPreviousModalForScreenReaders();
```

libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
```diff
@@ -47,4 +47,36 @@ export class SkyModalAdapterService {
   #removeClassFromBody(className: string): void {
     this.#bodyEl.classList.remove(className);
   }
+
+  public addAriaHidden(
```
Blackbaud-TrevorBurch: Nitpicky - could we add a comment here just stating that we return back the initial value of the elements' `aria-hidden` attribute?

libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
```diff
@@ -47,4 +47,36 @@ export class SkyModalAdapterService {
...
+    elementsToHide: HTMLCollection,
+    self: Element | unknown
+  ): Map<Element, string | null> {
+    const hiddenElements = new Map<Element, string | null>();
```
Blackbaud-TrevorBurch: Would a better name for this be `elementsPreviousValue` or something similar since this map isn't just the elements but also is that.

libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
```diff
@@ -47,4 +47,36 @@ export class SkyModalAdapterService {
...
+      ) {
+        hiddenElements.set(element, element.getAttribute('aria-hidden'));
+        element.setAttribute('aria-hidden', 'true');
+      }
+    }
+
+    return hiddenElements;
+  }
+
+  public removeAriaHidden(hiddenElements: Map<Element, string | null>): void {
```
Blackbaud-TrevorBurch: I wonder if we should call this `restoreOrRemoveAriaHidden` and then name the parameter something similar to whatever we name the variable in the field above (see my comment there).

We aren't always removing it in the case where it previously was set.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -104,6 +119,7 @@ export class SkyModalHostComponent implements OnDestroy {
 
     const adapter = this.#adapter;
     const modalOpener: HTMLElement = adapter.getModalOpener();
+    let ariaPreviousValueSiblings = this.#ariaPreviousValueSiblings;
```
Blackbaud-TrevorBurch: I don't follow why we need this local variable? It seems like all that is happening is that it is set to the class level field and then cleared at the end. Couldn't we just reference the class level field all the time?

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
Blackbaud-TrevorBurch: What if there are three modals and the middle once closes. Wouldn't this unhide the 1st and leave both the 1st and 3rd unhidden?

libs/components/modals/src/lib/modules/modal/modal.service.spec.ts
Blackbaud-TrevorBurch: We tend to write our spec descriptions in a `should...` format that describes what should happen. I don't know based on the description here or in the next test what they "should" be doing.

Blackbaud-TrevorBurch (changes requested):

libs/components/modals/src/lib/modules/modal/modal.service.spec.ts
Blackbaud-SandhyaRajasabeson: do we not want to have the "should" statements be generic with no "code" language?

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
Blackbaud-SandhyaRajasabeson: my thinking was that since the user is in the modal there wouldn't be things added behind it so if we just hide things the first time a modal is opened we would be good. but i guess that's idealic thinking.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
Blackbaud-SandhyaRajasabeson: that would only be called when `sky-modal-host` is destroyed right? `sky-modal-host` isn't removed from the dom even when all the modals all closed so wouldn't it make more sense for this to be called when the last modal is closed and not when host is destroyed

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
Blackbaud-SandhyaRajasabeson: nope! it checks that the top modal is the one that is closing and only then unhides any modals

Blackbaud-SandhyaRajasabeson: there is a unit test for this scenario also just to make sure no changes break that expectation https://github.com/blackbaud/skyux/blob/modal-navigation-accessibility/libs/components/modals/src/lib/modules/modal/modal.service.spec.ts#L239

Blackbaud-TrevorBurch (dismissed):

libs/components/modals/src/lib/modules/modal/modal.service.spec.ts
Blackbaud-SteveBrush: Does this `it` need to be in the `describe` block below?

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
Blackbaud-SteveBrush: Yeah, we can't control what other scripts are doing on the page. Unless I'm missing something, it wouldn't hurt to hide all siblings every time a new modal opens. Would be good to provide a test for this scenario as well.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -139,11 +142,34 @@ export class SkyModalHostComponent implements OnDestroy {
...
+      SkyModalHostService.openModalCount > 1 &&
+      SkyModalHostService.topModal == hostService
+    ) {
+      // hiding the lower modals when more than one modal is opened
+      this.#adapter.hidePreviousModal(modalElement);
+    }
+
+    const closeModal = () => {
+      // unhide siblings if last modal is closing
+      if (SkyModalHostService.openModalCount == 1) {
```
Blackbaud-SteveBrush: Needs a triple-equals `===`.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -139,11 +142,34 @@ export class SkyModalHostComponent implements OnDestroy {
...
+        this.#adapter.unhideOrRestoreHostSiblings();
+      } else if (SkyModalHostService.topModal == hostService) {
```
Blackbaud-SteveBrush: Needs a triple-equals `===`.

libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
```diff
@@ -47,4 +48,50 @@ export class SkyModalAdapterService {
   #removeClassFromBody(className: string): void {
     this.#bodyEl.classList.remove(className);
   }
+
+  public hideHostSiblings(hostElRef: ElementRef): void {
```
Blackbaud-SteveBrush: For posterity, let's put a JSDocs comment above this method describing what it does.

libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
```diff
@@ -47,4 +48,50 @@ export class SkyModalAdapterService {
...
+        this.#hostSiblingPreviousValue.set(
+          element,
+          element.getAttribute('aria-hidden')
+        );
+        element.setAttribute('aria-hidden', 'true');
+      }
+    }
+  }
+
+  public unhideOrRestoreHostSiblings(): void {
```
Blackbaud-SteveBrush: For posterity, let's put a JSDocs comment above this method describing what it does.

libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
```diff
@@ -13,6 +13,7 @@ export class SkyModalAdapterService {
   #bodyEl: HTMLElement;
 
   #windowRef: SkyAppWindowRef;
+  #hostSiblingPreviousValue = new Map<Element, string | null>();
```
Blackbaud-SteveBrush: This property name is ambiguous. Got a better name that explains what kinds of values it's tracking?

libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
```diff
@@ -47,4 +48,50 @@ export class SkyModalAdapterService {
...
+      if (previousValue) {
+        element.setAttribute('aria-hidden', previousValue);
+      } else {
+        element.removeAttribute('aria-hidden');
+      }
+    });
+    this.#hostSiblingPreviousValue.clear();
+  }
+
+  public hidePreviousModal(topModal: Element): void {
```
Blackbaud-SteveBrush: For all four of these methods, the word "hide" is misleading since the modal is not being hidden for sighted users. Even though it's a tad wordy, I would add the qualifier "FromScreenReaders" to the method names so it's 100% clear.
```
public hidePreviousModalFromScreenReaders
```

libs/components/modals/src/lib/modules/modal/modal-adapter.service.ts
```diff
@@ -47,4 +48,50 @@ export class SkyModalAdapterService {
...
+    if (topModal && topModal.previousElementSibling) {
+      topModal.previousElementSibling.setAttribute('aria-hidden', 'true');
+    }
+  }
+
+  public unhidePreviousModal(topModal: Element): void {
```
Blackbaud-SteveBrush: Adapter services should never receive an `Element` but rather an `ElementRef`. The idea behind this decision is that these adapter services can be switched out with other platform adapters (besides the DOM). For example, if we wanted to use the modal stuff in a mobile app, we could switch the DOM adapter with a mobile phone adapter. Since the mobile phone adapter wouldn't know what an `Element` is, we should always pass the `ElementRef` since that will represent either a DOM element or a mobile phone "element".

This should be updated in a few places above, too.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -139,11 +142,34 @@ export class SkyModalHostComponent implements OnDestroy {
...
 
+    // modal element that was just opened
+    const modalElement = modalComponentRef.location.nativeElement;
+
     modalInstance.componentInstance = modalComponentRef.instance;
 
     this.#registerModalInstance(modalInstance);
 
-    function closeModal() {
+    if (SkyModalHostService.openModalCount == 1) {
```
Blackbaud-SteveBrush: Needs a triple-equals `===`.

libs/components/modals/src/lib/modules/modal/modal-host.component.ts
```diff
@@ -139,11 +142,34 @@ export class SkyModalHostComponent implements OnDestroy {
...
+      // hidding all elements at the modal-host level from screenreaders when the first modal is opened
+      this.#adapter.hideHostSiblings(this.#elRef);
+    }
+    if (
+      SkyModalHostService.openModalCount > 1 &&
+      SkyModalHostService.topModal == hostService
```
Blackbaud-SteveBrush: Needs a triple-equals `===`.

Blackbaud-SteveBrush (approved):

Blackbaud-SteveBrush (approved):

PR #403, opened 2022-08-16
refactor(components/indicators): label strict mode support

Blackbaud-TrevorBurch:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/403/storybook/)

Component Storybooks:

- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/403/storybooks/indicators-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/403/storybooks/tabs-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/403/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/403/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/403/playground/)

libs/components/indicators/src/lib/modules/label/label.component.ts
```diff
@@ -25,13 +25,19 @@ export class SkyLabelComponent {
     return this._labelType || 'info';
   }
 
-  public baseIcon: SkyIconStackItem;
+  // The next three properties are set in the constructor. However, due to the function - the build doesn't see them as being set.
+  // We set a definite assignment assersion here on the properties here because of this.
```
Blackbaud-SteveBrush: Can we just remove the `updateIcon` method altogether and put the contents in the constructor?

Blackbaud-TrevorBurch: No - it is also called by the `labelType` setter. I could copy the logic both places, but that seemed wrong and like unneeded clutter.

Blackbaud-SteveBrush: Gotcha. It looks like you added the constructor call in this pull request, meaning those public properties were potentially undefined before this change. What's the harm in making those public properties | undefined? I'd rather do that (and keep it functionally the same before the changes introduced in this PR) instead of adding a comment explaining an exceptional use case.

Blackbaud-TrevorBurch: Hopped on a call about this. I backed out my bug fix and fixed the default logic to not use the getter.

Blackbaud-SteveBrush (approved):

PR #415, opened 2022-08-18
fix(components/ag-grid): top scroll not visible

johnhwhite:
[AB#2262309](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2262309)
[AB#2195399](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2195399)

- [ ] Changed top scroll to use CSS solution rather than DOM changes
- [ ] Added visual tests, including variations that affect top scroll


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/415/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/415/storybooks/ag-grid-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/415/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/415/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/415/playground/)

apps/ag-grid-storybook/src/app/ag-grid/ag-grid-stories.component.stories.ts
```diff
@@ -0,0 +1,27 @@
+import { Meta, Story, moduleMetadata } from '@storybook/angular';
```
Blackbaud-ErikaMcVey, line 1: the file pattern of these stories seems different than all the others...is that intentional? Others are like `ag-grid.component.ts` and `ag-grid.component.stories.ts`

apps/ag-grid-storybook/src/app/ag-grid/ag-grid-stories.component.scss
```diff
@@ -0,0 +1,43 @@
+:root {
```
Blackbaud-ErikaMcVey, line 1: looks like there is more styling here than the typical storybook. Did you run into any gotchas that would be worth noting in the e2e testing docs? Seems like its pretty ag grid specific so maybe not

apps/playground/src/app/components/ag-grid/data-manager-large/data-manager-large.component.ts
```diff
@@ -12,16 +14,24 @@ import {
...
 
 @Component({
   selector: 'app-data-manager-large',
   templateUrl: './data-manager-large.component.html',
+  styleUrls: ['./data-manager-large.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush,
+  encapsulation: ViewEncapsulation.None,
 })
 export class DataManagerLargeComponent implements OnInit {
+  @HostBinding('class.use-normal-dom-layout')
```
Blackbaud-ErikaMcVey, line 30: do consumers have to add this info (the styling above, the host binding, turning off view encapsulation) in order to use top scroll? Or was this just to support switching between normal and autoheight in the demo?

Blackbaud-ErikaMcVey:
It looks like the first visual test sreenshot might've failed when it shouldn't have--looks like fonts might not have loaded? Icons aren't loaded and the font is definitely different.

oh, storybook also calls out an accessibility violation--"Scrollable region must have keyboard access". Top scroll works with up/down arrow keys but horizontal scroll does not work with left/right arrow keys. They can move horizontally if they focus on a cell and move across them so maybe that is enough for accessibility?

apps/ag-grid-storybook/src/app/ag-grid/ag-grid-stories.component.stories.ts
johnhwhite: The name was to avoid duplicate `AgGridModule` definitions.

apps/ag-grid-storybook/src/app/ag-grid/ag-grid-stories.component.scss
johnhwhite: When not using auto-height, AG Grid needs a height style. With or without auto-height, setting exact sizes helped minimize the overflow displayed in Percy. Otherwise Percy attempts to find the right viewport size.

apps/playground/src/app/components/ag-grid/data-manager-large/data-manager-large.component.ts
johnhwhite: This was to support switching and to set heights when not using auto-height.

Blackbaud-ErikaMcVey (approved):

PR #426, opened 2022-08-22
fix(components/toast): toast shows icon when default state is being used

Blackbaud-TrevorBurch:


libs/components/toast/src/lib/modules/toast/fixtures/toast-with-toaster-service.component.fixture.ts
```diff
@@ -8,7 +8,7 @@ import { SkyToasterService } from '../toaster.service';
 
 @Component({
   selector: 'sky-test-cmp',
-  templateUrl: './toast.component.fixture.html',
+  templateUrl: './toast-with-toaster-service.component.fixture.html',
```
Blackbaud-TrevorBurch, line 11: I found this when working... we actually had the `html` file but the fixture was pointing at the wrong one. I didn't research the existing tests any more but they still passed when I changed it. We may just not need this fixture (shrug)

libs/components/toast/src/lib/modules/toast/toast.component.ts
```diff
@@ -45,13 +46,18 @@ export class SkyToastComponent implements OnInit, OnDestroy {
    * Specifies a `SkyToastType` type for the toast to determine the color and icon to display.
    */
   @Input()
-  public set toastType(value: SkyToastType) {
-    this._toastType = value;
+  public set toastType(value: SkyToastType | undefined) {
```
Blackbaud-TrevorBurch, line 49: I did a couple strict mode fixes in this file such as getting rid of the getter here and using the `OrDefault` convention. However, I did not strict mode everything including doing private private things as that was not in scope and there are a lot of violations in this component. I started to do them and it became too much for this unrelated fix and so I backed them out.

blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/426/storybook/)

Component Storybooks:

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/426/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/426/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/426/playground/)

libs/components/toast/src/lib/modules/toast/toast.component.ts
```diff
@@ -45,13 +46,18 @@ export class SkyToastComponent implements OnInit, OnDestroy {
...
+    if (value === undefined) {
```
Blackbaud-PaulCrowder: Did strict mode prevent you from using `this.toastTypeOrDefault = value || SKY_TOAST_TYPE_DEFAULT`? If so, I'd prefer the ternary operator here:

```
this.toastTypeOrDefault = value === undefined ? SKY_TOAST_TYPE_DEFAULT : value;
```

libs/components/toast/src/lib/modules/toast/toast.component.ts
```diff
@@ -45,13 +46,18 @@ export class SkyToastComponent implements OnInit, OnDestroy {
...
+      this.toastTypeOrDefault = SKY_TOAST_TYPE_DEFAULT;
+    } else {
+      this.toastTypeOrDefault = value;
+    }
     this.updateIcon();
-  }
 
-  public get toastType(): SkyToastType {
-    return this._toastType === undefined ? SkyToastType.Info : this._toastType;
+    this.ariaLive =
```
Blackbaud-PaulCrowder: Wouldn't the ARIA live and ARIA role properties also need to be set when the initial state of the component is established with the default toast type? If so, I'd rename `updateIcon()` to `updateForToastType()` and move this logic there.

libs/components/toast/src/lib/modules/toast/toast.component.ts
```diff
@@ -112,12 +116,12 @@ export class SkyToastComponent implements OnInit, OnDestroy {
...
   private ngUnsubscribe = new Subject<void>();
 
-  private _toastType: SkyToastType;
-
   constructor(
     private changeDetector: ChangeDetectorRef,
     @Optional() private toasterService?: SkyToasterService
-  ) {}
+  ) {
+    this.updateIcon();
```
Blackbaud-PaulCrowder: I'd prefer to see this moved to `ngOnInit()` with the rest of the logic that initializes the state of the component, such as the `isOpen` state.

libs/components/toast/src/lib/modules/toast/toast.component.ts
Blackbaud-TrevorBurch: Nope. That was just me coding too fast.

libs/components/toast/src/lib/modules/toast/toast.component.ts
Blackbaud-TrevorBurch: Done

libs/components/toast/src/lib/modules/toast/toast.component.ts
Blackbaud-TrevorBurch: Fixed

libs/components/toast/src/lib/modules/toast/toast.component.ts
Blackbaud-TrevorBurch: In theory - the defaults that I set on the fields are all that are needed as they are correct for the default. However, I don't think what you suggested hurts and does give us some confidence so I went ahead and did that.

Blackbaud-PaulCrowder (approved):

PR #440, opened 2022-08-26
refactor(components/http): http strict mode support

Blackbaud-TrevorBurch:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/440/storybook/)

Component Storybooks:

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/440/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/440/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/440/playground/)

libs/components/http/src/lib/modules/auth-http-client/auth-interceptor.spec.ts
```diff
@@ -297,4 +322,30 @@ describe('Auth interceptor', () => {
       jasmine.objectContaining(expectedTokenArgs)
     );
   });
+
+  // describe('with missing SkyAppConfig', () => {
```
Blackbaud-SteveBrush: Eh?

Blackbaud-TrevorBurch: Sigh - good catch. That was code that I temporarily brought over from a different spec as a template as I was working on the code above and I didn't intend to save and commit. I have removed it.

libs/components/http/src/lib/modules/auth-http-client/auth-interceptor.ts
```diff
@@ -40,21 +40,31 @@ function removeSkyParams(request: HttpRequest<any>): HttpRequest<any> {
...
+    this.#paramsProvider = paramsProvider;
+  }
 
   public intercept(
     request: HttpRequest<any>,
     next: HttpHandler
   ): Observable<HttpEvent<any>> {
-    let auth: boolean;
-    let permissionScope: string;
+    let auth: boolean | undefined;
```
Blackbaud-SteveBrush: I would hardset this to `false` instead of making a boolean have the possibility of being undefined.
```
let auth = false;
```

libs/components/http/src/lib/modules/auth-http-client/auth-interceptor.ts
```diff
@@ -40,21 +40,31 @@ function removeSkyParams(request: HttpRequest<any>): HttpRequest<any> {
...
+    let permissionScope: string | null | undefined;
```
Blackbaud-SteveBrush: Is it possible to limit this value to `string | undefined` so that we don't have to juggle two different "falsy" types? For the `params.get` method, you can do the following to set it to undefined if `null` is returned:
```
permissionScope = params.get(SKY_AUTH_PARAM_PERMISSION_SCOPE) ?? undefined;
```

libs/components/http/src/lib/modules/auth-http-client/auth-interceptor.ts
Blackbaud-TrevorBurch: Done

libs/components/http/src/lib/modules/auth-http-client/auth-interceptor.ts
Blackbaud-TrevorBurch: Done

Blackbaud-SteveBrush (approved):

PR #454, opened 2022-08-29
feat(components/tabs): tabset wizard finish nav button

Blackbaud-ErikaMcVey:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/454/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/454/storybooks/ag-grid-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/454/storybooks/modals-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/454/storybooks/pages-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/454/storybooks/tabs-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/454/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/454/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/454/playground/)

libs/components/tabs/src/lib/modules/tabs/tabset-nav-button.component.ts
```diff
@@ -23,126 +20,174 @@ const buttonTypePrevious = 'previous';
...
+        .pipe(distinctUntilChanged(), takeUntil(this.#ngUnsubscribe))
+        .subscribe((index: SkyTabIndex) => {
+          this.#activeSkyTabIndex = index;
+          this.#tabCount = this.#_tabset.tabs.length;
+          this.#updateTabToSelect();
+          this.#updateButtonVisibility();
+          this.#updateButtonProperties();
+        });
+    } else {
+      this.#logger.error(
```
Blackbaud-TrevorBurch, line 48: I wonder if we should actually throw an "Error" here instead of just logging and error and continuing execution. Thoughts?

libs/components/tabs/src/lib/modules/tabs/tabset-nav-button.component.ts
```diff
@@ -23,126 +20,174 @@ const buttonTypePrevious = 'previous';
...
-    this._buttonText = value;
+  public get buttonType(): SkyTabsetNavButtonType {
+    return this.#_buttonType;
   }
 
-  public get buttonText(): string {
-    if (this._buttonText) {
-      return this._buttonText;
-    }
+  public set buttonType(value: SkyTabsetNavButtonType) {
```
Blackbaud-TrevorBurch: This is technically a breaking change... I hate it... but it is....

We took in any string before and so there is no guarantee that consumers aren't doing something dumb. We should probably try to handle random strings like we were previously with the `default` case in the `buttonText` setter. Happy to chat around this if needed.

Blackbaud-TrevorBurch: If we would even handle this with a union of your new type and `string` for now and then do a breaking change in 7 to remove that.

libs/components/tabs/src/lib/modules/tabs/tabset-nav-button.component.ts
```diff
@@ -23,126 +20,174 @@ const buttonTypePrevious = 'previous';
...
+  public buttonText?: string | undefined;
+
+  /**
+   * Indicates whether to disable the nav button.
+   * Defaults to the disabled state of the next tab for `next`, the existence of a previous tab for `previous`,
+   * and false for `finish`.
+   */
+  @Input()
+  public set disabled(value: boolean) {
+    this.#_disabled = value;
```
Blackbaud-TrevorBurch, line 90: We should probably set a separate public property in the setter here and update it when `buttonType` updates or when `disabled` updates and then use that separate property in the template. That way we don't have logic in the getter but the getter also doesn't return something different than what was set. We do similar in other places that have defaults such as `ariaRoleOrDefault` in some places.

libs/components/tabs/src/lib/modules/tabs/tabset-nav-button.component.ts
```diff
@@ -23,126 +20,174 @@ const buttonTypePrevious = 'previous';
...
+      return this.#_disabled;
+    } else if (this.#_buttonType === buttonTypeFinish) {
+      return false;
     }
 
-    return !tabToSelect || tabToSelect.disabled;
+    return !this.#tabToSelect || this.#tabToSelect.disabled;
+  }
+
+  public buttonClassName: string;
```
Blackbaud-TrevorBurch: string | undefined

libs/components/tabs/src/lib/modules/tabs/tabset-nav-button.component.ts
```diff
@@ -23,126 +20,174 @@ const buttonTypePrevious = 'previous';
...
+  public isVisible = false;
+  public type: 'submit' | 'button';
+  public ariaControls: string;
```
Blackbaud-TrevorBurch: string | undefined

Blackbaud-TrevorBurch (changes requested):
@Blackbaud-ErikaMcVey just a few things

libs/components/tabs/src/lib/modules/tabs/tabset-nav-button.component.ts
Blackbaud-ErikaMcVey: Paul and I talked about this since the progress indicator does throw an error--we decided the intention was mostly to let developers know "hey you can't do this" so logging something would be enough.

libs/components/tabs/src/lib/modules/tabs/tabset-nav-button.component.ts
Blackbaud-ErikaMcVey: I'm not sure I'm following. The logic has changed a bit, but they can still pass any string for button text. The template renders that if it's been passed, or the resource string if it hasn't. I think the diff for this just is comparing button text and type a little oddly

Blackbaud-ErikaMcVey: ohhh nvm I see what you mean now 😬

libs/components/tabs/src/lib/modules/tabs/tabset-nav-button.component.ts
Blackbaud-ErikaMcVey: unfortunately the nav buttons rely on it being a getter. Every time something changes in the wizard form (i.e. user input), we do need to re-check if its disabled since it relies on the tab's disabled property and we don't emit any event when that changes

Blackbaud-TrevorBurch (approved):

PR #464, opened 2022-08-30
fix(components/indicators): only escape highlight directive search text once

Blackbaud-PaulCrowder:


Blackbaud-PaulCrowder:
This PR also includes strict mode changes.

libs/components/indicators/src/lib/modules/text-highlight/text-highlight.directive.ts
```diff
@@ -27,160 +114,98 @@ export class SkyTextHighlightDirective
...
-  private static cleanRegex(regex: string): string {
-    return regex.replace(/[-/\\^$*+?.()|[\]{}]/g, '\\$&');
-  }
-
-  private static markNode(node: any, searchTerms: string[]): number {
-    /* istanbul ignore else */
-    if (searchTerms) {
-      const text = node.nodeValue;
-      for (let i = 0; i < searchTerms.length; i++) {
-        searchTerms[i] = this.cleanRegex(searchTerms[i]);
```
Blackbaud-PaulCrowder, line 66: This is where the bug where search terms were escaped multiple times originated. The original array would be updated with a re-escaped version of each item for every child element, so when a certain number of child elements were present, the regular expression reached a size so large it would cause an error. For example, the first iteration would escape `'.'` with `'\.'`. The second iteration would escape the already-escaped string `'\.'` with `'\\\.'`, and so on.

libs/components/indicators/src/lib/modules/text-highlight/text-highlight.directive.ts
```diff
@@ -9,7 +9,94 @@ import {
...
+    }
+  }
+}
+
+function createSearchRegex(searchTerms: string[]): RegExp | undefined {
+  let searchRegex: RegExp | undefined;
+
+  if (searchTerms) {
+    searchTerms = searchTerms.map((searchTerm) =>
+      searchTerm.replace(/[-/\\^$*+?.()|[\]{}]/g, '\\$&')
```
Blackbaud-SteveBrush: A brief comment describing the purpose of the `RegExp` (or what it's doing) would be helpful for future debugging.

Blackbaud-PaulCrowder: Added.

Blackbaud-SteveBrush (dismissed):

Blackbaud-SteveBrush (approved):

PR #478, opened 2022-09-02
refactor(components/storybook): storybook strict mode support

Blackbaud-TrevorBurch:


libs/components/storybook/src/lib/preview-wrapper/preview-wrapper.component.ts
```diff
@@ -62,7 +62,7 @@ export class PreviewWrapperComponent implements OnInit, OnDestroy {
     SkyTheme.presets.default,
     SkyThemeMode.presets.light
   );
-  private _theme?: 'default' | 'modern-light' | 'modern-dark';
+  private _theme: 'default' | 'modern-light' | 'modern-dark' = 'default';
   private readonly _ngUnsubscribe = new Subscription();
```
Blackbaud-SteveBrush: Do these properties need the `#_` treatment?

libs/components/storybook/src/lib/preview-wrapper/preview-wrapper.component.ts
```diff
@@ -24,8 +24,8 @@ import { Subscription } from 'rxjs';
 export class PreviewWrapperComponent implements OnInit, OnDestroy {
   @Input()
   public set theme(value: 'default' | 'modern-light' | 'modern-dark') {
```
Blackbaud-SteveBrush: Input values can always be `undefined`.

libs/components/storybook/src/lib/preview-wrapper/preview-wrapper.component.ts
```diff
@@ -45,40 +45,46 @@ export class PreviewWrapperComponent implements OnInit, OnDestroy {
     }
   }
   public get theme() {
```
Blackbaud-SteveBrush: Needs a return type.

libs/components/storybook/src/lib/preview-wrapper/preview-wrapper.component.ts
```diff
@@ -45,40 +45,46 @@ export class PreviewWrapperComponent implements OnInit, OnDestroy {
...
 
-  private _themeSettings = new SkyThemeSettings(
+  #_themeSettings = new SkyThemeSettings(
     SkyTheme.presets.default,
     SkyThemeMode.presets.light
   );
-  private _theme?: 'default' | 'modern-light' | 'modern-dark';
-  private readonly _ngUnsubscribe = new Subscription();
-  private initialized = false;
+  #_theme: 'default' | 'modern-light' | 'modern-dark' = 'default';
```
Blackbaud-SteveBrush: Can we make this string union an official type, or at least a type defined at the top of this file?

libs/components/storybook/src/lib/preview-wrapper/preview-wrapper.component.ts
```diff
@@ -45,40 +45,46 @@ export class PreviewWrapperComponent implements OnInit, OnDestroy {
...
+  #ngUnsubscribe = new Subscription();
+  #initialized = false;
+
+  #themeService: SkyThemeService;
+  #renderer: Renderer2;
 
   constructor(
-    private themeService: SkyThemeService,
+    themeService: SkyThemeService,
     @Inject('BODY') private body: HTMLElement,
```
Blackbaud-SteveBrush: Still need the `private`? If so, maybe this should be an injection token instead.

libs/components/storybook/src/lib/preview-wrapper/preview-wrapper.component.ts
```diff
@@ -45,40 +45,46 @@ export class PreviewWrapperComponent implements OnInit, OnDestroy {
...
-    this.themeService.init(this.body, this.renderer, this.themeSettings);
-    this.initialized = true;
+    this.#themeService.init(this.body, this.#renderer, this.themeSettings);
+    this.#initialized = true;
   }
 
   public ngOnDestroy(): void {
-    this._ngUnsubscribe.unsubscribe();
-    this.themeService.destroy();
+    this.#ngUnsubscribe.unsubscribe();
```
Blackbaud-SteveBrush: Do we need to first run `this.#ngUnsubscribe.next()`? Should we run `complete` instead of `unsubscribe`?

libs/components/storybook/src/lib/preview-wrapper/preview-wrapper.component.ts
Blackbaud-TrevorBurch: Sigh. I actually had fixed this one and then messed it up on a merge. Fixed.

libs/components/storybook/src/lib/preview-wrapper/preview-wrapper.component.ts
Blackbaud-TrevorBurch: Fixed.

libs/components/storybook/src/lib/preview-wrapper/preview-wrapper.component.ts
Blackbaud-TrevorBurch: Done. I went ahead and just did what we do for the theme selector in the apps and made a local type. I wonder if we should have a more global way to do this defined in `theme`, but I didn't want to deal with that at this point.

libs/components/storybook/src/lib/preview-wrapper/preview-wrapper.component.ts
Blackbaud-TrevorBurch: That was just a miss. Fixed.

libs/components/storybook/src/lib/preview-wrapper/preview-wrapper.component.ts
Blackbaud-TrevorBurch: Ummm. When I looked deeper we weren't even using that field other than creating it and destroying it. I removed it all together.

libs/components/storybook/src/lib/preview-wrapper/preview-wrapper.component.ts
```diff
@@ -23,10 +23,10 @@ import { Subscription } from 'rxjs';
 })
 export class PreviewWrapperComponent implements OnInit, OnDestroy {
   @Input()
-  public set theme(value: 'default' | 'modern-light' | 'modern-dark') {
-    this._theme = value;
-    if (value && value.match(/^modern(-(light|dark))?$/)) {
-      if (value.includes('dark')) {
+  public set theme(value: PreviewWrapperThemeValue | undefined) {
+    this.#_theme = value ?? 'default';
+    if (this.#_theme && this.#_theme.match(/^modern(-(light|dark))?$/)) {
```
Blackbaud-SteveBrush: Won't the first condition, `this.#_theme`, always resolve to true?

libs/components/storybook/src/lib/preview-wrapper/preview-wrapper.component.ts
```diff
@@ -44,41 +44,47 @@ export class PreviewWrapperComponent implements OnInit, OnDestroy {
       );
     }
   }
-  public get theme() {
-    return this._theme;
+  public get theme(): PreviewWrapperThemeValue {
```
Blackbaud-SteveBrush: Inputs should always support `undefined` values, so shouldn't we do the `themeOrDefault` treatment on this property?

Blackbaud-TrevorBurch: So looking at this a little deeper - the getter wasn't even needed. It wasn't being used anywhere other than in the setter. I changed the setter to just use a local variable and removed the getter all together.

libs/components/storybook/src/lib/preview-wrapper/preview-wrapper.component.ts
Blackbaud-TrevorBurch: Good catch. Removed.

Blackbaud-SteveBrush (approved):

PR #480, opened 2022-09-02
fix(components/data-manager): update activeView on data view config updates

bradenbiz:


libs/components/data-manager/src/lib/modules/data-manager/data-manager-toolbar/data-manager-toolbar.component.spec.ts
```diff
@@ -434,16 +479,35 @@ describe('SkyDataManagerToolbarComponent', () => {
     );
   });
 
+  it('should update the active view config when any view is updated', () => {
```
Blackbaud-ErikaMcVey, line 482: So this is technically hitting that line because the code is called when the component first renders and subscribes to that, but it isn't _really_ testing that the toolbar does indeed update when something else updates the view config. We should actually call `dataManagerService.updateViewConfig()` in this one. The setup will be a little more involved than the other tests--you will have to initialize the data manager and the view. LMK if you want to go over it together.

Blackbaud-ErikaMcVey (changes requested):
everything looks pretty good, just needs one small change!

libs/components/data-manager/src/lib/modules/data-manager/data-manager-toolbar/data-manager-toolbar.component.spec.ts
bradenbiz: Good catch. Fixed in: [use dataManagerService to properly test](https://github.com/blackbaud/skyux/pull/480/commits/d04debd33481bf91cf48fb6d2ac4716206ce714f)

Blackbaud-ErikaMcVey (approved):

PR #500, opened 2022-09-09
feat(components/indicators): update alert test harness with accessibility functions

Blackbaud-ErikaMcVey:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/500/storybook/)

Component Storybooks:

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/500/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/500/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/500/playground/)

libs/components/indicators/testing/src/alert/alert-harness.ts
```diff
@@ -42,11 +46,7 @@ export class SkyAlertHarness extends SkyComponentHarness {
       return 'success';
     }
 
-    if (await alert.hasClass('sky-alert-warning')) {
-      return 'warning';
-    }
-
-    return undefined;
+    return 'warning';
```
Blackbaud-TrevorBurch, line 49: Can you explain why we are changing this? I know we have a default, but that default sets the class and without the class there is something wrong there and there is no alert type showing.

Blackbaud-TrevorBurch (commented):
@Blackbaud-ErikaMcVey one question

libs/components/indicators/testing/src/alert/alert-harness.ts
Blackbaud-ErikaMcVey: Warning is the default so there will never _not_ be a class--when we were returning undefined it was impossible to get test coverage to 100%.

If we break something to where the warning class no longer is added by default and no alert type is showing then I think it'd be the responsibility of the alert component tests to catch that, not the harness tests. What do you think?

Blackbaud-TrevorBurch: That’s fair. Ok.

Blackbaud-TrevorBurch (approved):

PR #509, opened 2022-09-12
chore(components/indicators): implement strict mode for indicators

Blackbaud-PaulCrowder:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/509/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/509/storybooks/ag-grid-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/509/storybooks/indicators-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/509/storybooks/modals-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/509/storybooks/pages-storybook/)
- [popovers-storybook](https://blackbaud.github.io/skyux-pr-preview/509/storybooks/popovers-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/509/storybooks/tabs-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/509/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/509/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/509/playground/)

libs/components/indicators/src/lib/modules/tokens/token.component.ts
```diff
@@ -19,15 +19,15 @@ export class SkyTokenComponent {
    * Indicates whether to disable the token to prevent users from selecting it, dismissing it,
    * or navigating to it with the arrow keys. When the token is disabled,
    * users can still place focus on it using the `Tab` key.
-   * @default true
+   * @default false
```
Blackbaud-SteveBrush, line 22: 😄

libs/components/indicators/src/lib/modules/tokens/token.component.ts
```diff
@@ -36,14 +36,12 @@ export class SkyTokenComponent {
    * @default "Remove item"
    */
   @Input()
-  public set ariaLabel(value: string) {
-    this._ariaLabel = value;
+  public set ariaLabel(value: string | undefined) {
+    this.#ariaLabelOrDefault = value || this.#getDefaultAriaLabel();
```
Blackbaud-SteveBrush, line 40: Move forward with the `fooOrDefault` pattern and fix later?

Blackbaud-SteveBrush, line 40: If they pass `''` I'm assuming we want to revert to the default label? If not, we can use null coalescing to only set the default when `undefined` is passed.

libs/components/indicators/src/lib/modules/tokens/token.component.ts
```diff
@@ -86,40 +80,31 @@ export class SkyTokenComponent {
...
   public tokenFocus = new EventEmitter<void>();
 
-  /**
-   * @internal
-   */
-  public get tabIndex(): number | boolean {
-    return this.focusable ? 0 : -1;
-  }
+  public tokenActive: boolean | undefined;
+  public closeActive: boolean | undefined;
```
Blackbaud-SteveBrush: Since `undefined` is falsey, could we set these deliberately to `false` and avoid using `| undefined`?

libs/components/indicators/src/lib/modules/tokens/tokens.component.ts
```diff
@@ -110,39 +111,39 @@ export class SkyTokensComponent implements OnDestroy {
...
   }
 
   /**
    * Specifies an array of tokens to include in the list.
    */
   @Input()
   public set tokens(value: SkyToken[] | undefined) {
-    this._tokens = value;
-    this.tokensChange.emit(this._tokens);
+    this.#_tokensOrDefault = value || [];
```
Blackbaud-SteveBrush: Move forward with the `fooOrDefault` pattern and fix later?

libs/components/indicators/src/lib/modules/tokens/tokens.component.ts
```diff
@@ -189,29 +188,30 @@ export class SkyTokensComponent implements OnDestroy {
...
+
+  #_disabled = false;
+  #_dismissible = true;
+  #_focusable = true;
+  #_tokensOrDefault: SkyToken[] = [];
+  #_displayWith = 'name';
 
   @ViewChildren(SkyTokenComponent)
-  private tokenComponents: QueryList<SkyTokenComponent>;
+  public tokenComponents!: QueryList<SkyTokenComponent>;
```
Blackbaud-SteveBrush: I brought this up during standup a few weeks ago (using non-null assertions on view children) and we agreed we should be using `| undefined` since they're technically undefined in the constructor. I'm personally okay using non-null assertions on view children, but we should probably be consistent. Which do you prefer?

libs/components/indicators/src/lib/modules/tokens/tokens.component.ts
```diff
@@ -110,39 +111,39 @@ export class SkyTokensComponent implements OnDestroy {
...
   @Input()
-  public set messageStream(value: Subject<SkyTokensMessage>) {
-    this._messageStream = value;
-    this.initMessageStream();
-  }
-
-  public get messageStream(): Subject<SkyTokensMessage> {
-    return this._messageStream;
+  public set messageStream(value: Subject<SkyTokensMessage> | undefined) {
+    this.#messageStreamOrDefault = value || new Subject<SkyTokensMessage>();
```
Blackbaud-SteveBrush: Instead of setting a `fooOrDefault` property, why not use a getter for the `messageStream`? If not, it would help me if we specified in our team docs when it's appropriate to use the `fooOrDefault` pattern as opposed to a normal getter.

libs/components/indicators/testing/src/label/label-harness.ts
```diff
@@ -41,18 +41,22 @@ export class SkyLabelHarness extends SkyComponentHarness {
    * Gets the `labelType` of the label component.
    */
   public async getLabelType(): Promise<SkyLabelType> {
-    const labelClasses = await (await this.#getWrapper()).getAttribute('class');
+    const labelClasses: DOMTokenList = await (
+      await this.#getWrapper()
+    ).getProperty('classList');
```
Blackbaud-SteveBrush, line 46: Good change. 👍

libs/components/indicators/testing/src/tokens/token-harness.ts
```diff
@@ -9,7 +9,7 @@ import { SkyTokenHarnessFilters } from './token-harness-filters';
 export class SkyTokenHarness extends ComponentHarness {
   public static hostSelector = 'sky-token';
 
-  #getDismissButton = this.locatorForOptional('button.sky-token-btn-close');
+  #getDismissButton = this.locatorFor('button.sky-token-btn-close');
```
Blackbaud-SteveBrush, line 12: This selector can sometimes return `null` if the token is not dismissible.

libs/components/indicators/src/lib/modules/tokens/token.component.ts
Blackbaud-PaulCrowder: I used `#ariaLabelOrDefault` here because I set its default value in the constructor (it calls a function which is why I don't set it inline to its declaration). I didn't want to have an instance where I reference an underscored private field anywhere but the `get` and `set` accessors.

Blackbaud-PaulCrowder: I could go either way. I can't really see a situation where the consumer _wouldn't_ want an ARIA label, so setting it to `''` should probably set it back to the default.

libs/components/indicators/testing/src/tokens/token-harness.ts
Blackbaud-PaulCrowder: True, but the only place where we call it is right after we check to see if the token is dismissible. I changed this code so I wouldn't have to wrap the click logic in an `if` statement that I couldn't reach with a unit test.

libs/components/indicators/src/lib/modules/tokens/token.component.ts
Blackbaud-PaulCrowder: Fixed.

libs/components/indicators/src/lib/modules/tokens/tokens.component.ts
Blackbaud-PaulCrowder: Fixed.

libs/components/indicators/src/lib/modules/tokens/tokens.component.ts
Blackbaud-PaulCrowder: Fixed.

libs/components/indicators/src/lib/modules/tokens/tokens.component.ts
Blackbaud-PaulCrowder: Fixed.

Blackbaud-SteveBrush (approved):

PR #536, opened 2022-09-19
refactor(components/forms): radio strict mode support

Blackbaud-TrevorBurch:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/536/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/536/storybooks/ag-grid-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/536/storybooks/layout-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/536/storybooks/pages-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/536/storybooks/tabs-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/536/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/536/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/536/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/536/playground/)

apps/code-examples/src/app/code-examples/forms/radio/icon/radio-demo.component.ts
```diff
@@ -1,11 +1,11 @@
...
 
 @Component({
   selector: 'app-radio-demo',
   templateUrl: './radio-demo.component.html',
 })
-export class RadioDemoComponent implements OnInit {
+export class RadioDemoComponent {
   public myForm: FormGroup;
 
   public views: any[] = [
```
Blackbaud-SteveBrush: Possible to get rid of the `any`? We could create a separate interface (exported from a separate file) like we did for Lookup's code examples.

apps/code-examples/src/app/code-examples/forms/radio/inline-help/radio-demo.component.ts
```diff
@@ -1,11 +1,11 @@
...
 
 @Component({
   selector: 'app-radio-demo',
   templateUrl: './radio-demo.component.html',
 })
-export class RadioDemoComponent implements OnInit {
+export class RadioDemoComponent {
   public myForm: UntypedFormGroup;
 
   public options: any[] = [
```
Blackbaud-SteveBrush: Same here.

apps/code-examples/src/app/code-examples/forms/radio/standard/radio-demo.component.ts
```diff
@@ -1,11 +1,11 @@
...
 
 @Component({
   selector: 'app-radio-demo',
   templateUrl: './radio-demo.component.html',
 })
-export class RadioDemoComponent implements OnInit {
+export class RadioDemoComponent {
   public myForm: FormGroup;
 
   public options: any[] = [
```
Blackbaud-SteveBrush: Same here.

libs/components/forms/src/lib/modules/radio/fixtures/radio-group-reactive.component.fixture.ts
```diff
@@ -25,27 +27,31 @@ export class SkyRadioGroupReactiveFixtureComponent implements OnInit {
     { name: 'Harry Mckenzie', disabled: false },
   ];
 
-  public radioControl: UntypedFormControl;
-  public radioForm: UntypedFormGroup;
+  public radioControl!: UntypedFormControl;
+  public radioForm!: UntypedFormGroup;
```
Blackbaud-SteveBrush: Possible to remove the `!`?

libs/components/forms/src/lib/modules/radio/fixtures/radio-group-reactive.component.fixture.ts
```diff
@@ -25,27 +27,31 @@ export class SkyRadioGroupReactiveFixtureComponent implements OnInit {
...
 
   @ViewChild(SkyRadioGroupComponent)
-  public radioGroupComponent: SkyRadioGroupComponent;
+  public radioGroupComponent: SkyRadioGroupComponent | undefined;
+
+  #fb: UntypedFormBuilder;
 
-  constructor(private fb: UntypedFormBuilder) {}
+  constructor(fb: UntypedFormBuilder) {
+    this.#fb = fb;
```
Blackbaud-SteveBrush: Let's make this variable consistent with other implementations, `formBuilder: UntypedFormBuilder` and `#formBuilder`.

libs/components/forms/src/lib/modules/radio/radio-group.component.ts
```diff
@@ -212,69 +230,69 @@ export class SkyRadioGroupComponent
...
   }
 
   #updateValue(value: any, markDirty: boolean): void {
-    const isNewValue = value !== this._value;
+    const isNewValue = value !== this.#_value;
 
     /* istanbul ignore else */
     if (isNewValue) {
-      this._value = value;
+      this.#_value = value;
```
Blackbaud-SteveBrush: Accessing a setter's property outside of the setter?

libs/components/forms/src/lib/modules/radio/radio-group.component.ts
```diff
@@ -212,69 +230,69 @@ export class SkyRadioGroupComponent
...
       if (markDirty) {
-        this.onChange(this._value);
+        this.#onChange(this.#_value);
```
Blackbaud-SteveBrush: Same here.

libs/components/forms/src/lib/modules/radio/radio.component.ts
```diff
@@ -58,46 +58,52 @@ export class SkyRadioComponent implements OnDestroy, ControlValueAccessor {
...
   @Input()
-  public set disabled(value: boolean) {
+  public set disabled(value: boolean | undefined) {
     const coercedValue = SkyFormsUtility.coerceBooleanProperty(value);
     if (coercedValue !== this.disabled) {
-      this._disabled = coercedValue;
-      this._disabledChange.next(this._disabled);
-      this.changeDetector.markForCheck();
+      this.#_disabled = coercedValue;
+      this.#_disabledChange.next(this.#_disabled);
```
Blackbaud-SteveBrush: We're accessing a setter's property outside of the setter. Maybe this should be a new value.
```
#disabledChange = new Subject<boolean>();

constructor() {
  #disabledChangeObs = this.#disabledChange.asObservable();
}

public get disabledChange(): Observable<boolean> {
  return this.#disabledChangeObs;
}
```
For example: https://github.com/blackbaud/skyux/blob/main/libs/components/core/src/lib/modules/affix/affixer.ts#L48

libs/components/forms/src/lib/modules/radio/radio.component.ts
```diff
@@ -140,21 +146,30 @@ export class SkyRadioComponent implements OnDestroy, ControlValueAccessor {
...
       'The sky-radio `tabindex` property is deprecated. Please use the `tabindex` property on the sky-radio-group component.'
     );
-    this._tabindex = value;
-    this.changeDetector.markForCheck();
+    if (value) {
+      this.#_tabindex = value;
+    } else {
+      this.#_tabindex = 0;
+    }
+    this.#changeDetector.markForCheck();
```
Blackbaud-SteveBrush, line 157: Why is this needed (since this is an input)?

libs/components/forms/src/lib/modules/radio/radio.component.ts
```diff
@@ -201,11 +216,13 @@ export class SkyRadioComponent implements OnDestroy, ControlValueAccessor {
...
   @Input()
   public get radioType(): string {
-    return this._radioType || 'info';
+    return this.#_radioType;
   }
-  public set radioType(value: string) {
+  public set radioType(value: string | undefined) {
     if (value) {
-      this._radioType = value.toLowerCase();
+      this.#_radioType = value.toLowerCase();
```
Blackbaud-SteveBrush: Probably should use `toLocaleLowerCase()`?

libs/components/forms/src/lib/modules/radio/radio.component.ts
```diff
@@ -215,55 +232,61 @@ export class SkyRadioComponent implements OnDestroy, ControlValueAccessor {
...
    * Fires when the selected value changes.
    */
   @Output()
   public get disabledChange(): Observable<boolean> {
-    return this._disabledChange;
-  }
-
-  public get inputId(): string {
-    return `sky-radio-${this.id}-input`;
+    return this.#_disabledChange;
```
Blackbaud-SteveBrush: ☝️ All of these outputs should be using the `asObservable` pattern described above.

libs/components/forms/src/lib/modules/radio/radio.component.ts
```diff
@@ -215,55 +232,61 @@ export class SkyRadioComponent implements OnDestroy, ControlValueAccessor {
...
+  #_change = new EventEmitter<SkyRadioChange>();
+  #_checked = false;
+  #_checkedChange = new BehaviorSubject<boolean>(this.#_checked);
+  #_disabled = false;
+  #_disabledChange = new BehaviorSubject<boolean>(this.#_disabled);
+  #_name: string | undefined;
+  #_radioType = 'info';
+  #_selectedValue: any;
+  #_tabindex = 0;
+  #_value: any;
```
Blackbaud-SteveBrush, line 284: Is there a better type for this?

Blackbaud-SteveBrush: Could we get away with using `unknown` here?

libs/components/forms/testing/src/radio-fixture.ts
```diff
@@ -79,12 +79,14 @@ export class SkyRadioFixture {
...
   }
 
-  private getRadioButtonInputEl(index: number): DebugElement {
+  private getRadioButtonInputEl(index: number): DebugElement | undefined {
     const allRadioInputs = this.getAllRadioInputEls();
 
     if (allRadioInputs && allRadioInputs[index]) {
       return allRadioInputs[index];
     }
+    /* istanbul ignore next */
```
Blackbaud-SteveBrush, line 88: Possible to put `/* istanbul ignore else */` on the `if` statement, instead?

apps/code-examples/src/app/code-examples/forms/radio/icon/radio-demo.component.ts
Blackbaud-TrevorBurch: Done

apps/code-examples/src/app/code-examples/forms/radio/inline-help/radio-demo.component.ts
Blackbaud-TrevorBurch: Done

apps/code-examples/src/app/code-examples/forms/radio/standard/radio-demo.component.ts
Blackbaud-TrevorBurch: Done

libs/components/forms/src/lib/modules/radio/fixtures/radio-group-reactive.component.fixture.ts
Blackbaud-TrevorBurch: Done

libs/components/forms/src/lib/modules/radio/fixtures/radio-group-reactive.component.fixture.ts
Blackbaud-TrevorBurch: Done

libs/components/forms/src/lib/modules/radio/radio-group.component.ts
Blackbaud-TrevorBurch: Ya... I don't love it either. The trick here is the difference between if we mark the form dirty or not. The setter and `writeValue` both call this function as their logic with a boolean flag. There is no way to pass that boolean flag to the setter and so we would still need to set the underscore variable in `writeValue` even if this function didn't exist.

libs/components/forms/src/lib/modules/radio/radio-group.component.ts
Blackbaud-TrevorBurch: See above

libs/components/forms/src/lib/modules/radio/radio.component.ts
Blackbaud-TrevorBurch: Fixed

libs/components/forms/src/lib/modules/radio/radio.component.ts
Blackbaud-TrevorBurch: Fixed

libs/components/forms/src/lib/modules/radio/radio.component.ts
Blackbaud-TrevorBurch: Fixed

libs/components/forms/src/lib/modules/radio/radio.component.ts
Blackbaud-TrevorBurch: Updated to `unknown`

libs/components/forms/testing/src/radio-fixture.ts
Blackbaud-TrevorBurch: Nope - I tried. Istanbul doesn't see it for some reason.

libs/components/forms/src/lib/modules/radio/radio.component.ts
Blackbaud-TrevorBurch: Because the radio group component calls the input in typescript and, as we discovered in flex the other day, that type of assignment on an input does not trigger change detection. It is only when inputs are updated through template bindings that trigger change detection.

Blackbaud-SteveBrush: Ah, right.

libs/components/forms/src/lib/modules/radio/radio.component.ts
```diff
@@ -215,66 +232,78 @@ export class SkyRadioComponent implements OnDestroy, ControlValueAccessor {
...
   public get checkedChange(): Observable<boolean> {
-    return this._checkedChange;
+    return this.#checkedChangeObs;
   }
 
   /**
    * Fires when the selected value changes.
    */
   @Output()
   public get disabledChange(): Observable<boolean> {
```
Blackbaud-SteveBrush, line 253: ☝️ Just noticed this, but Angular recommends outputs be of type `EventEmitter`, but we're using `Observable`. Can you put `// TODO: convert to EventEmitter in a breaking change` above each one?

libs/components/forms/src/lib/modules/radio/radio.component.ts
```diff
@@ -287,41 +316,41 @@ export class SkyRadioComponent implements OnDestroy, ControlValueAccessor {
...
+    this.#onTouchedCallback = fn;
   }
 
   public onInputChange(event: Event): void {
     event.stopPropagation();
 
     if (!this.disabled) {
       this.checked = true;
-      this._change.emit({
+      this.#_change.emit({
```
Blackbaud-SteveBrush: To avoid accessing this underscored property outside of the getter, maybe we can convert this to using the `asObservable` pattern the other outputs are using?

libs/components/forms/src/lib/modules/radio/radio.component.ts
```diff
@@ -287,41 +316,41 @@ export class SkyRadioComponent implements OnDestroy, ControlValueAccessor {
...
         source: this,
-        value: this._value,
+        value: this.#_value,
```
Blackbaud-SteveBrush: Should we use the getter for this?

libs/components/forms/src/lib/modules/radio/radio.component.ts
```diff
@@ -215,66 +232,78 @@ export class SkyRadioComponent implements OnDestroy, ControlValueAccessor {
...
+  #_value: unknown;
+
+  #changeDetector: ChangeDetectorRef;
+
+  constructor(changeDetector: ChangeDetectorRef) {
+    this.#changeDetector = changeDetector;
+    this.#checkedChange = new BehaviorSubject<boolean>(this.checked);
+    this.#checkedChangeObs = this.#checkedChange.asObservable();
+    this.#disabledChange = new BehaviorSubject<boolean>(this.disabled);
+    this.#disabledChangeObs = this.#disabledChange.asObservable();
```
Blackbaud-SteveBrush, line 295: ☝️ These observables need to be completed in the `ngOnDestroy`.

libs/components/forms/src/lib/modules/radio/radio-group.component.ts
```diff
@@ -116,87 +120,101 @@ export class SkyRadioGroupComponent
...
+  #_disabled = false;
 
-  private _name = `sky-radio-group-${nextUniqueId++}`;
+  #_name = '';
 
-  private _tabIndex: number;
+  #_tabIndex: number | undefined;
 
-  private _value: any;
+  #_value: any;
```
Blackbaud-SteveBrush: Possible to use `unknown`, here?

libs/components/forms/src/lib/modules/radio/radio-group.component.ts
Blackbaud-SteveBrush: I think we've uncovered some "smelly" code. You're right that there's two entry points into the form control, one for the user's input (the `value` input), and one for the control's programmatic updates (`writeValue`). We might need to rethink how we're handling the "true" value of the control, and instead of accessing the `#_value` field here, we should create a new private property called `#controlValue` (or something) that both the `value` input and the `writeValue` method update (`#_value` would only represent the value passed to the `value` input).

Historically, we haven't been the best about separating the concerns of user input versus `writeValue` calls, so worst case, I think it would be better to duplicate code to represent the two concerns (if needed), rather than have branching logic in a singular method that both the `value` input and `writeValue` method call into.

libs/components/forms/src/lib/modules/radio/radio.component.ts
Blackbaud-TrevorBurch: Done

libs/components/forms/src/lib/modules/radio/radio.component.ts
Blackbaud-TrevorBurch: Hmmm. our typing isn't even right here. I converted to use the pattern - using a `Subject` since the existing `EventEmitter` didn't hold a value.

libs/components/forms/src/lib/modules/radio/radio.component.ts
Blackbaud-TrevorBurch: Good catch. Fixed

libs/components/forms/src/lib/modules/radio/radio.component.ts
Blackbaud-TrevorBurch: Fixed

libs/components/forms/src/lib/modules/radio/radio-group.component.ts
Blackbaud-TrevorBurch: Done

Blackbaud-SteveBrush (approved):

PR #606, opened 2022-09-30
refactor(components/autonumeric): autonumeric strict mode support

bradenbiz:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/606/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/606/storybooks/ag-grid-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/606/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/606/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/606/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/606/playground/)

libs/components/autonumeric/src/lib/modules/autonumeric/fixtures/autonumeric.component.fixture.ts
```diff
@@ -19,39 +13,35 @@ import { SkyAutonumericDirective } from '../autonumeric.directive';
...
-  public ngOnInit(): void {
-    this.formGroup = this.formBuilder.group({
+  constructor(formBuilder: UntypedFormBuilder) {
+    this.#formBuilder = formBuilder;
+
+    this.formGroup = this.#formBuilder.group({
       donationAmount: new UntypedFormControl(),
     });
+
+    this.formControl = this.formGroup.controls['donationAmount'];
```
Blackbaud-TrevorBurch: The standard way of doing this would be `this.formGroup.get('donationAmount')`.

Blackbaud-TrevorBurch: Or you could create the form control right after 39 and assign it to `this.formControl` and then on line 42 just pass `this.formControl` to where you are setting `donationAmount` in the group function.

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
```diff
@@ -52,103 +52,115 @@ export class SkyAutonumericDirective
...
+    this.#autonumericOptions = this.#mergeOptions(value);
+    this.#updateAutonumericInstance();
   }
 
   /**
    * @internal
    */
   @Input()
-  public skyAutonumericFormChangesUnformatted = false;
+  public skyAutonumericFormChangesUnformatted: boolean | undefined = false;
```
Blackbaud-TrevorBurch, line 64: Since this has a default value of `false` this should be converted to a getter/setter that sets `false` to a private variable internally whenever `undefined` is given and then the getter just returns that private variable.

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
```diff
@@ -52,103 +52,115 @@ export class SkyAutonumericDirective
...
+    renderer: Renderer2,
+    changeDetector: ChangeDetectorRef
   ) {
-    this.autonumericInstance = new AutoNumeric(this.elementRef.nativeElement);
+    this.#elementRef = elementRef;
+    this.#globalConfig = globalConfig;
+    this.#renderer = renderer;
+    this.#changeDetector = changeDetector;
+
+    this.#autonumericInstance = new AutoNumeric(this.#elementRef.nativeElement);
```
Blackbaud-TrevorBurch: Minor - but passing just `elementRef.nativeElement` is slightly more efficient.

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
```diff
@@ -165,53 +177,57 @@ export class SkyAutonumericDirective
...
   }
 
-  private getNumericValue(): number | undefined {
-    const inputValue = this.getInputValue();
-    return inputValue && !this.isInputValueTheCurrencySymbol(inputValue)
-      ? this.autonumericInstance.getNumber()
+  #getNumericValue(): number | undefined {
+    const inputValue = this.#getInputValue();
+    const result = this.#autonumericInstance.getNumber();
+    const numericValueOrUndefined = !result ? undefined : result;
```
Blackbaud-TrevorBurch: I would just add `&& !!result` to the opening portion of the ternary - then you won't need the extra ternary that you added and can just return back `? result : undefined`

libs/components/autonumeric/src/lib/modules/autonumeric/fixtures/autonumeric.component.fixture.ts
bradenbiz: done in [refactor formControl creation](https://github.com/blackbaud/skyux/pull/606/commits/92308bf5459603002b8979c52aa371c0fb10f3a3)

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
bradenbiz: fixed in: [efficiency](https://github.com/blackbaud/skyux/pull/606/commits/2a39f6c5a270b87a7ea811cc81054c2742539f12)

libs/components/autonumeric/src/lib/modules/autonumeric/fixtures/autonumeric.component.fixture.ts
bradenbiz: I realized this was breaking things. doing the first way causes a warning about donationAmount existing, and the second way does not allow donationAmount to be found. reverted in: [revert form control intialization form](https://github.com/blackbaud/skyux/pull/606/commits/7c5054fa10e9c64cd1386b7e6a725845e0659556)

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
bradenbiz: fixed in: [revert form control intialization form](https://github.com/blackbaud/skyux/pull/606/commits/7c5054fa10e9c64cd1386b7e6a725845e0659556)

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
bradenbiz: good suggestion. fixed in: [simplify function](https://github.com/blackbaud/skyux/pull/606/commits/29379133a5c7980e92e027509639cb41e69accca)

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
bradenbiz: will you please confirm I did it the way you expected? thanks

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
```diff
@@ -52,103 +52,123 @@ export class SkyAutonumericDirective
...
    * @internal
    */
   @Input()
-  public skyAutonumericFormChangesUnformatted = false;
+  public get skyAutonumericFormChangesUnformatted(): boolean | undefined {
+    return this.#_skyAutonumericFormChangesUnformatted;
+  }
+
+  public set skyAutonumericFormChangesUnformatted(value: boolean | undefined) {
+    this.#_skyAutonumericFormChangesUnformatted = value || false;
```
Blackbaud-SteveBrush: A boolean value will trigger the "falsey" check of this operator. You can use null coallescing to only check against `null` or `undefined`:
```
this.#_skyAutonumericFormChangesUnformatted = value ?? false;
```

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
```diff
@@ -165,53 +185,58 @@ export class SkyAutonumericDirective
...
-    const inputValue = this.getInputValue();
-    return inputValue && !this.isInputValueTheCurrencySymbol(inputValue)
-      ? this.autonumericInstance.getNumber()
+  #getNumericValue(): number | undefined {
+    const inputValue = this.#getInputValue();
+    const result = this.#autonumericInstance.getNumber();
+    return inputValue &&
+      !this.#isInputValueTheCurrencySymbol(inputValue) &&
+      !!result
+      ? result
```
Blackbaud-SteveBrush: This operation can be simplified a bit (and account for zero values being returned from the `this.#autonumericInstance.getNumber` method):
```
#getNumericValue(): number | undefined {
  const inputValue = this.#getInputValue();
  const numericValue = this.#autonumericInstance.getNumber();

  return inputValue &&
    !this.#isInputValueTheCurrencySymbol(inputValue) &&
    typeof numericValue === 'number'
    ? numericValue
    : undefined;
}
```

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
Blackbaud-TrevorBurch: See Steve's comment about using `??` instead of `||`.

libs/components/autonumeric/src/lib/modules/autonumeric/fixtures/autonumeric.component.fixture.ts
Blackbaud-TrevorBurch: Your change that you reverted wasn't quite what I meant. You can do

```
this.#formBuilder = formBuilder;

this.formControl = new FormControl('donationAmount');

this.formGroup = this.#formBuilder.group({ donationAmount: this.formControl });
```

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.spec.ts
```diff
@@ -207,98 +189,41 @@ describe('Autonumeric directive', () => {
...
 
     input.value = '1000';
     SkyAppTestUtility.fireDomEvent(input, 'input');
     SkyAppTestUtility.fireDomEvent(input, 'keyup');
-    detectChanges();
-
-    expect(spy).toHaveBeenCalled();
-  }));
-
-  it('should not update numeric value on keyup when no change is made', fakeAsync(() => {
```
Blackbaud-TrevorBurch, line 227: It seems like we could leave this test if we want. You could just test that the input value doesn't change from `1000` to `1,000`.

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.spec.ts
```diff
@@ -386,20 +296,7 @@ describe('Autonumeric directive', () => {
...
-
-      verifyFormControlStatuses({
-        dirty: false,
-        pristine: true,
-        touched: false,
-        valid: true,
-      });
-    }));
-
-    it('should set correct statuses when initialized with falsey value', fakeAsync(() => {
```
Blackbaud-TrevorBurch, line 399: Is there a reason you got rid of this test?

bradenbiz: This test's name is a misnomer. We don't want the autonumeric directive to accept _any_ falsy value. We only want to confirm that it accepts **_numeric_** falsy values. 

Since the only numeric falsy value in JavaScript is 0, an accurate name for what the test was doing is: 

> 'should set correct statuses when initialized with **_a numeric_** falsy value'

Therefore, this test was redundant thanks to this existing test:

```
it('should properly format 0 values', fakeAsync(() => {
    detectChanges();

    setValue(0);

    detectChanges();

    const modelValue = getModelValue();
    const formattedValue = getFormattedValue();

    expect(modelValue).toEqual(0);
    expect(formattedValue).toEqual('0.00');
}));
```

libs/components/autonumeric/src/lib/modules/autonumeric/fixtures/autonumeric.component.fixture.ts
bradenbiz: Doing it this way is causing the formControl to be marked as invalid on initialization.

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
bradenbiz: Good suggestion. This breaks code coverage, so I'll need to write a test for passing in `undefined`.

bradenbiz: done in: [use null coalescing instead of logical OR for input](https://github.com/blackbaud/skyux/pull/606/commits/33851fcd14e534b5b03e409b681ff461b31e1936)

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.spec.ts
bradenbiz: What would it be testing? I think it's redundant, due to the existence of this test: 
```
it('should not update numeric value on keyup when no change is made and a currency symbol is specified', fakeAsync(() => {
    setOptions('dollar');

    const input = fixture.nativeElement.querySelector('input');

    SkyAppTestUtility.fireDomEvent(input, 'mouseenter');
    SkyAppTestUtility.fireDomEvent(input, 'input');
    SkyAppTestUtility.fireDomEvent(input, 'keyup');

    expect(fixture.componentInstance.formControl.value).toBeFalsy();
}));
```

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
bradenbiz: Ahh... good call. done in: [modify function to allow for 0 values](https://github.com/blackbaud/skyux/pull/606/commits/95bf87efd4997ed0790ad913a9d958dd39b5713b)

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.spec.ts
Blackbaud-TrevorBurch: I could go either way here. The two tests were trying to test the same functionality; however, one was testing with the currency option set and one was testing without. So they were technically checking the same thing in different configurations.

libs/components/autonumeric/src/lib/modules/autonumeric/fixtures/autonumeric.component.fixture.ts
bradenbiz: fixed in: [fix formControl init](https://github.com/blackbaud/skyux/pull/606/commits/43edc78ecbc9dc06f3223e48641439609e24d944)

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.spec.ts
bradenbiz: reverted in: [add back non-redundant test](https://github.com/blackbaud/skyux/pull/606/commits/dea3c560d06a763a25ca4dfe35c24b04f1b9764e)

Blackbaud-TrevorBurch (dismissed):

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
```diff
@@ -52,103 +52,123 @@ export class SkyAutonumericDirective
...
+  public set skyAutonumeric(value: SkyAutonumericOptions | undefined) {
+    this.#autonumericOptions = this.#mergeOptions(value);
+    this.#updateAutonumericInstance();
   }
 
   /**
    * @internal
    */
   @Input()
-  public skyAutonumericFormChangesUnformatted = false;
```
Blackbaud-SteveBrush, line 64: Any reason why this shouldn't be the following (based on our team discussion yesterday)?
```
public skyAutonumericFormChangesUnformatted: boolean | undefined = false;
```

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
bradenbiz: @Blackbaud-SteveBrush does this need to be undone?

libs/components/autonumeric/src/lib/modules/autonumeric/autonumeric.directive.ts
bradenbiz: nope. thanks for catching that.

bradenbiz: fixed in: [revert added getter/setter](https://github.com/blackbaud/skyux/pull/606/commits/36f6290353e4a70df8d6749b6ab69cd0f987578f)

Blackbaud-SteveBrush (approved):

PR #622, opened 2022-10-04
refactor(components/lookup): lookup library strict mode finishes

Blackbaud-TrevorBurch:
[AB#2195489](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2195489)
[AB#2195393](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2195393)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/622/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/622/storybooks/ag-grid-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/622/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/622/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/622/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/622/playground/)

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -109,19 +109,25 @@ export class SkyAutocompleteComponent implements OnDestroy, AfterViewInit {
    */
   @Input()
   public set descriptorProperty(value: string | undefined) {
-    this.#_descriptorProperty = value;
+    this.#_descriptorProperty = value ?? 'name';
```
Blackbaud-PaulCrowder: Using `??` changes the behavior from that of the previous code. `value ?? 'name'` returns `''` if `value` is an empty string, while `return this.#_descriptorProperty || 'name'` returns `'name'` if `value` is an empty string. I'd prefer we continue using `||` for these coalescing scenarios unless the previous behavior was fundamentally wrong.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -109,19 +109,25 @@ export class SkyAutocompleteComponent implements OnDestroy, AfterViewInit {
...
   }
 
   /**
    * @internal
    * Indicates whether to display a button in the dropdown that opens a picker where users can view all options.
    */
   @Input()
-  public enableShowMore = false;
+  public set enableShowMore(value: boolean | undefined) {
+    this.#_enableShowMore = value ?? false;
```
Blackbaud-PaulCrowder: Can we just change the declaration of this input to `enableShowMore: boolean | undefined = false`? Presumably the rest of the code functioned just fine when `enableShowMore` was undefined. Otherwise I'd prefer `this.#_enableShowMore = !!value` to preserve the previous behavior where all falsy values effectively resolved to `false` instead of only `null` or `undefined`.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -195,14 +209,16 @@ export class SkyAutocompleteComponent implements OnDestroy, AfterViewInit {
    */
   @Input()
   public set searchTextMinimumCharacters(value: number | undefined) {
-    this.#_searchTextMinimumCharacters = value;
+    // If a user gives us 0 we set the variable to 1 as we would only search once a character is given.
+    if (value === 0 || value === undefined) {
```
Blackbaud-PaulCrowder: This changes the intent of the previous code in the case where `searchTextMinimumCharacters` is set to a negative value. I'd prefer to move the previous logic for defaulting to `1` here instead of rewriting the logic.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -213,31 +229,52 @@ export class SkyAutocompleteComponent implements OnDestroy, AfterViewInit {
...
    */
   @Input()
-  public searchFilters: SkyAutocompleteSearchFunctionFilter[] | undefined;
+  public set searchFilters(
+    value: SkyAutocompleteSearchFunctionFilter[] | undefined
+  ) {
+    this.#_searchFilters = value;
+
+    // Reset default search if it is what is being used.
+    if (this.search !== this.searchOrDefault) {
```
Blackbaud-PaulCrowder: This same logic is called in the `propertiesToSearch` setter, and similar logic exists in the `search` setter. You should consider combining all that logic into one function and call it from all 3 setters.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -213,31 +229,52 @@ export class SkyAutocompleteComponent implements OnDestroy, AfterViewInit {
...
+  }
 
   /**
    * Specifies the maximum number of search results to display in the dropdown list.
    * By default, the component displays all matching results.
    */
   @Input()
   public set searchResultsLimit(value: number | undefined) {
-    this.#_searchResultsLimit = value;
+    this.#_searchResultsLimit = value ?? 0;
```
Blackbaud-PaulCrowder: Use `||` here.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -213,31 +229,52 @@ export class SkyAutocompleteComponent implements OnDestroy, AfterViewInit {
...
 
   /**
    * @internal
    * Indicates whether to display a button that lets users add options to the data source.
+   * @default false
    */
   @Input()
-  public showAddButton = false;
+  public set showAddButton(value: boolean | undefined) {
+    this.#_showAddButton = value ?? false;
```
Blackbaud-PaulCrowder: Use `public showAddButton: boolean | undefined = false` or `this.#_showAddButton = !!value` here.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -250,9 +287,18 @@ export class SkyAutocompleteComponent implements OnDestroy, AfterViewInit {
...
    * Allows async search to be disabled even when a listener is specified for
    * the `searchAsync` output.
+   * @default false
    */
   @Input()
-  public searchAsyncDisabled = false;
+  public set searchAsyncDisabled(value: boolean | undefined) {
+    // NOTE: Istanbul branch ignores don't work for the `??` operator and we don't hit both branches as this is internal - so ignoring line.
+    /* istanbul ignore next */
+    this.#_searchAsyncDisabled = value ?? false;
```
Blackbaud-PaulCrowder: Same as above.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -372,30 +418,36 @@ export class SkyAutocompleteComponent implements OnDestroy, AfterViewInit {
...
     }
   }
 
-  private get resultsRef(): ElementRef | undefined {
+  public get resultsRef(): ElementRef | undefined {
     return this.#_resultsRef;
   }
 
+  public searchOrDefault: SkyAutocompleteSearchFunction =
+    skyAutocompleteDefaultSearchFunction({
```
Blackbaud-PaulCrowder: This assignment should probably go in the constructor since it calls a function.

libs/components/lookup/src/lib/modules/country-field/country-field.component.spec.ts
```diff
@@ -519,7 +531,9 @@ describe('Country Field Component', () => {
 
         const searchResults = searchAndGetResults('Austr', fixture);
         expect(
-          searchResults[0].querySelector('.sky-deemphasized').textContent.trim()
+          searchResults[0]
```
Blackbaud-PaulCrowder: Can this use `expect(...).toHaveText()` instead?

libs/components/lookup/src/lib/modules/lookup/lookup-autocomplete-adapter.ts
```diff
@@ -30,25 +30,33 @@ export class SkyLookupAutocompleteAdapter {
...
+    if (this.search !== this.searchOrDefault) {
+      this.searchOrDefault = skyAutocompleteDefaultSearchFunction({
+        propertiesToSearch: this.propertiesToSearch,
+        searchFilters: this.searchFilters,
+      });
+    }
   }
 
   public get propertiesToSearch(): string[] {
-    return this.#_propertiesToSearch || ['name'];
```
Blackbaud-PaulCrowder, line 51: This was a particularly nasty violation of the "don't return default values in a getter" because it would return a new array every time 🤦

libs/components/lookup/src/lib/modules/lookup/lookup-autocomplete-adapter.ts
```diff
@@ -30,25 +30,33 @@ export class SkyLookupAutocompleteAdapter {
...
    * @default ["name"]
    */
   @Input()
-  public set propertiesToSearch(value: string[]) {
-    this.#_propertiesToSearch = value;
+  public set propertiesToSearch(value: string[] | undefined) {
+    this.#_propertiesToSearch = value ?? ['name'];
+
+    // Reset default search if it is what is being used.
+    if (this.search !== this.searchOrDefault) {
```
Blackbaud-PaulCrowder: Most of the same guidance from the autocomplete comment applies here as the code is very similar.

libs/components/lookup/src/lib/modules/lookup/lookup-show-more-modal.component.ts
```diff
@@ -80,6 +82,7 @@ export class SkyLookupShowMoreModalComponent
 
   constructor(
     public modalInstance: SkyModalInstance,
+    @Inject(SKY_SHOW_MORE_NATIVE_PICKER_CONTEXT)
```
Blackbaud-PaulCrowder: I think it makes sense to keep `SkyLookupShowMoreNativePickerContext` as a class since it also serves as an injection token.

libs/components/lookup/src/lib/modules/lookup/lookup.component.ts
```diff
@@ -567,22 +600,20 @@ export class SkyLookupComponent
       | typeof SkyLookupShowMoreAsyncModalComponent;
 
     if (this.searchAsync.observers.length > 0) {
-      contextProviderType = SkyLookupShowMoreNativePickerAsyncContext;
       modalComponentType = SkyLookupShowMoreAsyncModalComponent;
 
-      contextProviderValue = new SkyLookupShowMoreNativePickerAsyncContext();
+      contextProviderValue = {} as SkyLookupShowMoreNativePickerAsyncContext;
```
Blackbaud-PaulCrowder: This won't apply if we keep `SkyLookupShowMoreNativePickerAsyncContext` as a class, but for future reference we should use the object initializer syntax when replacing classes with interfaces:

```
contextProviderValue = {
  idProperty: this.idProperty,
  searchAsync: (args) => {
    this.searchAsync.emit(args);
    return args.result!;
  };
}
```

libs/components/lookup/src/lib/modules/search/search.component.ts
```diff
@@ -320,10 +346,25 @@ export class SkySearchComponent implements OnDestroy, OnInit, OnChanges {
...
       (this.isCollapsible || this.isCollapsible === undefined) &&
       this.isFullWidth !== true
     );
   }
+
+  #setupSearchChangedEvent(): void {
+    this.#searchUpdatedUnsubscribe.next();
+
+    this.#searchUpdated
+      .asObservable()
```
Blackbaud-PaulCrowder: This `asObservable()` isn't needed here. You'd only use that if you were exposing the `Subject` to an API consumer and you want to obscure the underlying `Subject` methods like `next()`.

libs/components/lookup/src/lib/modules/search/search.component.ts
```diff
@@ -320,10 +346,25 @@ export class SkySearchComponent implements OnDestroy, OnInit, OnChanges {
...
-  #searchShouldCollapse() {
+  #searchShouldCollapse(): boolean {
     return (
       (this.isCollapsible || this.isCollapsible === undefined) &&
       this.isFullWidth !== true
     );
   }
+
+  #setupSearchChangedEvent(): void {
+    this.#searchUpdatedUnsubscribe.next();
```
Blackbaud-PaulCrowder: Without calling `this.#searchUpdatedUnsubscribe.complete()` here, the previous subscription would never be unsubscribed. I'd prefer tracking the `Subscription` itself rather than a separate `Subject` since it's only used for this one `Subject`.

```
  this.#searchUpdatedSub?.unsubscribe();

 this.#searchUpdatedSub = this.#searchUpdated...subscribe(...)
```

libs/components/lookup/testing/src/lookup/lookup-harness.spec.ts
```diff
@@ -14,7 +14,7 @@ async function setupTest(options: { dataSkyId: string }) {
   const fixture = TestBed.createComponent(LookupHarnessTestComponent);
   const loader = TestbedHarnessEnvironment.loader(fixture);
 
-  let lookupHarness: SkyLookupHarness;
+  let lookupHarness: SkyLookupHarness | null;
```
Blackbaud-PaulCrowder: How is it possible that `lookupHarness` is `null`? And if it is possible, could we not throw an error before returning since all the tests need it to not be `null`?

libs/components/lookup/testing/src/search/search-harness.spec.ts
```diff
@@ -14,7 +14,7 @@ describe('Search harness', () => {
     const fixture = TestBed.createComponent(SearchHarnessTestComponent);
     const loader = TestbedHarnessEnvironment.loader(fixture);
 
-    let searchHarness: SkySearchHarness;
+    let searchHarness: SkySearchHarness | undefined;
```
Blackbaud-PaulCrowder: I don't see any calls to this function that don't supply `options.dataSkyId`, so we can remove the `| undefined` and the check for a falsy `options.dataSkyId`. Then you can get rid of all the Elvis operators.

Blackbaud-PaulCrowder:
There's a general pattern of using the nullish coalescing operator (`??`) instead of `||`. I'd prefer to use `||` (or in the case of booleans, `!!value`) to preserve the original intent of the code.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
Blackbaud-TrevorBurch: I updated the two identical ones to use a private function. The logic in the search setter was slightly different enough that I left it as it was.

libs/components/lookup/testing/src/lookup/lookup-harness.spec.ts
Blackbaud-TrevorBurch: The null check was because the `queryHarness` call on line 27 can return `null`. However, I just added a non-null there and removed the null checks elsewhere since this is in a test file and should be safe.

libs/components/lookup/src/lib/modules/lookup/types/lookup-show-more-native-picker-async-context.ts
```diff
@@ -7,20 +7,20 @@ import { SkyLookupShowMoreNativePickerConfig } from './lookup-show-more-native-p
  * @internal
  * Context for the show more native picker. These values are provided by the lookup component.
  */
-export interface SkyLookupShowMoreNativePickerAsyncContext {
-  descriptorProperty: string;
+export class SkyLookupShowMoreNativePickerAsyncContext {
```
Blackbaud-PaulCrowder, line 10: Since all of these are required, it probably makes sense to make all of these constructor parameters so an instance can't be created without setting those properties. Same goes for the other classes with the non-null assertions on their properties.

```
constructor(public descriptorProperty: string, public idProperty: string, ...) {}
```

Blackbaud-PaulCrowder (approved):

PR #626, opened 2022-10-04
refactor(components/flyout): flyout strict mode support

Blackbaud-TrevorBurch:
[AB#2195453](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2195453)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/626/storybook/)

Component Storybooks:

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/626/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/626/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/626/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/626/playground/)

apps/playground/src/app/components/flyout/flyout/flyout-responsive-demo-content.component.ts
```diff
@@ -9,8 +9,12 @@ import { SkyMediaBreakpoints, SkyMediaQueryService } from '@skyux/core';
...
   public currentMediaBreakpoint: string;
 
-  constructor(private mediaQueryService: SkyMediaQueryService) {
-    this.mediaQueryService.subscribe((breakpoint) => {
+  #mediaQueryService: SkyMediaQueryService;
+
+  constructor(mediaQueryService: SkyMediaQueryService) {
+    this.#mediaQueryService = mediaQueryService;
+
+    this.#mediaQueryService.subscribe((breakpoint) => {
```
Blackbaud-PaulCrowder: `#mediaQueryService` isn't referenced anywhere but the constructor, so there's no need for the private variable. You do want an `#ngUnsubscribe` with a `takeUntil()`, though, so this component doesn't keep responding to media query breakpoint changes after its destroyed.

libs/components/flyout/src/lib/modules/flyout/fixtures/flyout-sample-context.fixture.ts
```diff
@@ -1,6 +1,6 @@
-export class SkyFlyoutTestSampleContext {
-  public name: string;
-  public showAutofocusButton?: boolean;
-  public showIframe?: boolean;
-  public showNormalButton?: boolean;
+export interface SkyFlyoutTestSampleContext {
```
Blackbaud-PaulCrowder: I think it makes sense to keep this a class since it's also being used as an injection token.

libs/components/flyout/src/lib/modules/flyout/fixtures/flyout.component.fixture.ts
```diff
@@ -5,32 +5,36 @@ import { SkyFlyoutConfig } from '../../flyout/types/flyout-config';
...
+  }
 
   public openFlyout(options?: SkyFlyoutConfig): SkyFlyoutInstance<any> {
     if (!options) {
       options = {
         providers: [
           {
-            provide: SkyFlyoutTestSampleContext,
+            provide: SKY_FLYOUT_SAMPLE_CONTEXT,
             useValue: { name: 'Sam', showIframe: false },
```
Blackbaud-PaulCrowder: If we keep this a class, we should change this to `useFactory()` where we can properly instantiate an instance of `SkyFlyoutTestSampleContext`.

libs/components/flyout/src/lib/modules/flyout/flyout-iterator.component.ts
```diff
@@ -18,45 +18,45 @@ import { Subject } from 'rxjs';
...
+  #_previousButtonClick = new EventEmitter<void>();
 
   public ngOnDestroy(): void {
-    this.ngUnsubscribe.next();
-    this.ngUnsubscribe.complete();
-    this._previousButtonClick.complete();
-    this._nextButtonClick.complete();
+    this.#ngUnsubscribe.next();
+    this.#ngUnsubscribe.complete();
+    this.#_previousButtonClick.complete();
```
Blackbaud-PaulCrowder: This isn't needed on `EventEmitter` instances. Angular should handle unsubscribing when the consuming component is destroyed.

https://stackoverflow.com/questions/53342421/complete-output-eventemitter-in-ngondestroy-callback

libs/components/flyout/src/lib/modules/flyout/flyout-media-query.service.ts
```diff
@@ -9,21 +9,22 @@ import { BehaviorSubject, Subscription } from 'rxjs';
...
-    this.current
-  );
+  #currentSubject: BehaviorSubject<SkyMediaBreakpoints>;
 
-  private _current = SkyMediaBreakpoints.xs;
+  #_current = SkyMediaBreakpoints.xs;
 
   constructor() {
-    this.currentSubject.next(this._current);
+    this.#currentSubject = new BehaviorSubject<SkyMediaBreakpoints>(
```
Blackbaud-PaulCrowder: Why are we initializing `BehaviorSubject` with a value and then immediately emitting that same value again? Also, how do we know that `xs` is the correct initial breakpoint? It seems like this should be a `new ReplaySubject< SkyMediaBreakpoints>(1)` without an initial value.

libs/components/flyout/src/lib/modules/flyout/flyout.service.ts
```diff
@@ -33,31 +33,45 @@ import { SkyFlyoutMessageType } from './types/flyout-message-type';
...
-    private readonly _ngZone: NgZone,
+    coreAdapter: SkyCoreAdapterService,
+    windowRef: SkyAppWindowRef,
+    dynamicComponentService: SkyDynamicComponentService,
+    router: Router,
+    ngZone: NgZone,
     // NOTE: This used to be used for an `applicationRef.tick` which has since been removed.
     // We can not remove this due to it being a breaking change for those manually constructing the service.
-    @Optional() private readonly _applicationRef?: ApplicationRef
+    @Optional() applicationRef?: ApplicationRef
```
Blackbaud-PaulCrowder: Since this is for SKY UX 7, let's remove this parameter as a breaking change.

libs/components/flyout/src/lib/modules/flyout/flyout.service.spec.ts
```diff
@@ -54,10 +59,7 @@ describe('Flyout service', () => {
     });
     service.open(SkyFlyoutHostsTestComponent);
     applicationRef.tick();
```
Blackbaud-PaulCrowder, line 56: Do we still need all these `applicatioinRef.tick()` calls?

libs/components/flyout/src/lib/modules/flyout/flyout.service.spec.ts
```diff
@@ -75,10 +77,7 @@ describe('Flyout service', () => {
...
-    const spy = spyOn(
-      service['host'].instance.messageStream,
-      'next'
-    ).and.callThrough();
+    const spy = spyOn(Subject.prototype, 'next').and.callThrough();
```
Blackbaud-PaulCrowder: This is pretty gross; is there a better way to test this, such as observing the behavior of closing a flyout instead of relying on an internal implementation?

libs/components/flyout/src/lib/modules/flyout/flyout.service.spec.ts
```diff
@@ -36,7 +38,10 @@ describe('Flyout service', () => {
   ));
 
   it('should only create a single host component', () => {
-    const spy = spyOn(service as any, 'createHostComponent').and.callThrough();
+    const spy = spyOn(
```
Blackbaud-PaulCrowder: Can you not mock out `SkyDynamicComponentService` with a Jasmine SpyObj? If not, you can use `TestBed.inject()` to get the instance of `SkyDynamicComponentService` and then spy on its `createComponent` method.

libs/components/flyout/src/lib/modules/flyout/flyout.service.spec.ts
```diff
@@ -139,13 +138,13 @@ describe('Flyout service', () => {
...
   it('should remove the host when the user navigates through history if no closed event is fired in 500ms - sanity check', fakeAsync(() => {
-    service.open(SkyFlyoutHostsTestComponent);
+    const instance = service.open(SkyFlyoutHostsTestComponent);
     const dynamicService = TestBed.inject(SkyDynamicComponentService);
     const removeComponentSpy = spyOn(
       dynamicService,
       'removeComponent'
     ).and.callThrough();
-    spyOn(service['host'].instance.messageStream, 'next').and.stub();
+    spyOn(instance.closed, 'emit').and.stub();
```
Blackbaud-PaulCrowder: `.and.stub()` is the default behavior when you call `spyOn()` and is only needed to reverse another `.and` function, like `.and.returnValue()` or `.and.callFake()`.

libs/components/flyout/src/lib/modules/flyout/types/flyout-config-internal.ts
```diff
@@ -0,0 +1,30 @@
+import { SkyFlyoutConfig } from './flyout-config';
+
+/**
+ * Type to represent the flyout configuration after default values have been applied.
+ * @internal
+ */
+export interface SkyFlyoutConfigInternal extends SkyFlyoutConfig {
```
Blackbaud-PaulCrowder, line 7: I couldn't find any uses of this interface in this PR. Is it needed?

apps/playground/src/app/components/flyout/flyout/flyout-responsive-demo-content.component.ts
Blackbaud-TrevorBurch: I have removed the private variable; however, the media query service is odd in that you can't do `takeUntil` on it due to how it is written. I implemented the pattern we have in many of our components to unsubscribe in `ngOnDestroy`.

libs/components/flyout/src/lib/modules/flyout/fixtures/flyout.component.fixture.ts
Blackbaud-TrevorBurch: I think I did this right. Please let me know if not.

libs/components/flyout/src/lib/modules/flyout/flyout.service.spec.ts
Blackbaud-TrevorBurch: Yes. Because the service doesn't have a component fixture that we can detect changes from - without these calls change detection on the components that the service opens does not fire. I did remove these calls from the component test file.

libs/components/flyout/src/lib/modules/flyout/types/flyout-config-internal.ts
Blackbaud-TrevorBurch: It is used within `SkyFlyoutComponent` to account for where we have a config with optional properties but then set them in the `attach` function so that we can account that the properties will be set later. Happy to chat if you don't like what I did there.

libs/components/flyout/src/lib/modules/flyout/flyout.service.spec.ts
Blackbaud-TrevorBurch: I changed this to watch the dynamic component service like the other tests. I don't love this as it still is somewhat of an internal implementation; however, I tried changing to watch for the flyout in the DOM and was running into a lot of timing issues with when the DOM element is actually removed.

libs/components/flyout/src/lib/modules/flyout/flyout-media-query.service.ts
```diff
@@ -1,26 +1,23 @@
...
 
-  #_current = SkyMediaBreakpoints.xs;
+  #_current: SkyMediaBreakpoints | undefined;
 
   constructor() {
-    this.#currentSubject = new BehaviorSubject<SkyMediaBreakpoints>(
-      this.current
-    );
-    this.#currentSubject.next(this.#_current);
+    this.#currentSubject = new ReplaySubject<SkyMediaBreakpoints>(1);
```
Blackbaud-PaulCrowder, line 16: This initializes the `ReplaySubject` but now it doesn't immediately emit `this.#_current`. Was this intentional?

libs/components/flyout/src/lib/modules/flyout/flyout-media-query.service.ts
```diff
@@ -1,26 +1,23 @@
...
-  public get current(): SkyMediaBreakpoints {
+  public get current(): SkyMediaBreakpoints | undefined {
     return this.#_current;
   }
 
-  #currentSubject: BehaviorSubject<SkyMediaBreakpoints>;
+  #currentSubject: ReplaySubject<SkyMediaBreakpoints>;
 
-  #_current = SkyMediaBreakpoints.xs;
+  #_current: SkyMediaBreakpoints | undefined;
```
Blackbaud-PaulCrowder: It looks like we're still referencing `this.#_current` outside the getter. Since that's the case, it should be renamed `#current`.

apps/playground/src/app/components/flyout/flyout/flyout-responsive-demo-content.component.ts
Blackbaud-PaulCrowder: I don't see where we're unsubscribing from `#mediaQueryService` in this component.

Blackbaud-TrevorBurch: Oof - you're right. I apparently blanked on circling back after I dug in there. Fixed.

libs/components/flyout/src/lib/modules/flyout/flyout-media-query.service.ts
Blackbaud-TrevorBurch: Yes. I changed that based on your comment regarding `xs` and if we knew that it was the right default. `xs` is the initial size of most flyouts but not all. Our flyout code immediately calls `setBreakpointForWidth` in order to set a breakpoing (and then update the subject); however, the previous behavior of initially outputting `xs` was based on the default width and not reality. I'm happy to add the `xs` initial back as to not change things if you'd like.

libs/components/flyout/src/lib/modules/flyout/flyout-media-query.service.ts
Blackbaud-TrevorBurch: We actually can just get rid of that private variable and modify and use the public one. Done.

Blackbaud-PaulCrowder (approved):

PR #661, opened 2022-10-11
chore(components/lists): filter strict mode support

Blackbaud-CoreyArcher:
[AB#2195445](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2195445)

### DIFF
--- a/libs/components/lists/src/lib/modules/filter/filter-button.component.spec.ts
+++ b/libs/components/lists/src/lib/modules/filter/filter-button.component.spec.ts
@@ -22,15 +22,15 @@ describe('Filter button', () => {
     fixture.detectChanges();
   });
 
-  function getButtonEl() {
+  function getButtonEl(): HTMLButtonElement {
     return nativeElement.querySelector('.sky-btn') as HTMLButtonElement;
   }
 
-  function verifyTextPresent() {
+  function verifyTextPresent(): void {
     expect(getButtonEl().innerText.trim()).toBe('Filter');
   }
 
-  function verifyTextNotPresent() {
+  function verifyTextNotPresent(): void {
     expect(getButtonEl().innerText.trim()).not.toBe('Filter');
   }
 
@@ -45,7 +45,7 @@ describe('Filter button', () => {
   it('should allow setting id', () => {
     component.buttonId = 'i-am-an-id-look-at-me';
     fixture.detectChanges();
-    expect(nativeElement.querySelector('.sky-btn').id).toBe(
+    expect(nativeElement.querySelector('.sky-btn')?.id).toBe(
       'i-am-an-id-look-at-me'
     );
   });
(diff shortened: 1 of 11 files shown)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/661/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/661/storybooks/ag-grid-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/661/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/661/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/661/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/661/playground/)

libs/components/lists/testing/src/paging/paging-fixture.spec.ts
```diff
@@ -26,11 +26,11 @@ const DATA_SKY_ID = 'test-paging';
   `,
 })
 class PagingTestComponent {
-  public currentPage: number = 1;
-  public itemCount: number = 8;
-  public maxPages: number = 3; // only show 3 pages in pager
-  public pageLabel: string;
```
Blackbaud-CoreyArcher, line 32: This got missed in the last PR

libs/components/lists/src/lib/modules/filter/filter-button.component.spec.ts
```diff
@@ -22,15 +22,15 @@ describe('Filter button', () => {
     fixture.detectChanges();
   });
 
-  function getButtonEl() {
+  function getButtonEl(): HTMLButtonElement {
```
Blackbaud-ErikaMcVey, line 25: is it possible for this to ever be undefined?

libs/components/lists/src/lib/modules/filter/fixtures/filter-button.component.fixture.ts
```diff
@@ -9,10 +9,10 @@ export class FilterButtonTestComponent {
...
   public showButtonText = false;
   public buttonClicked = false;
-  public buttonId: string;
-  public ariaExpanded: boolean;
-  public ariaControls: string;
-  public skyThemeSettings: SkyThemeSettings;
+  public buttonId: string | undefined;
+  public ariaExpanded: boolean | undefined;
+  public ariaControls: string | undefined;
+  public skyThemeSettings!: SkyThemeSettings;
```
Blackbaud-ErikaMcVey, line 15: could we set a default for this so we can avoid using `!`?

Blackbaud-CoreyArcher: I did this because Angular doesn't see it as being set in the constructor (due to the function call) but it will always have a value. I could have copied the code out of `useDefaultTheme()` so it wouldn't complain, but I think this is exactly the situation that `!` is intended to be used for.

libs/components/lists/src/lib/modules/filter/filter-button.component.spec.ts
Blackbaud-CoreyArcher: I don't think so? The fixture has a FilterButtonComponent child, which has a button child so searching for that class should always return a result.

Blackbaud-ErikaMcVey (approved):

PR #671, opened 2022-10-12
fix(components/forms): set radio group 'aria-owns' to satisfy accessibility rules

Blackbaud-SteveBrush:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/671/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/671/storybooks/ag-grid-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/671/storybooks/layout-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/671/storybooks/pages-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/671/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/671/storybooks/tabs-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/671/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/671/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/671/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/671/playground/)

Blackbaud-TrevorBurch (dismissed):

Blackbaud-TrevorBurch (dismissed):

libs/components/forms/src/lib/modules/radio/fixtures/radio.component.fixture.html
```diff
@@ -9,6 +9,8 @@ <h1 id="radio-group-label">Radio label</h1>
       [label]="label1"
       [value]="value1"
       [(ngModel)]="selectedValue"
+      (checkedChange)="onCheckedChange($event)"
+      (disabledChange)="onDisabledChange($event)"
```
Blackbaud-SteveBrush, line 13: ☝️ Added these for coverage.

libs/components/forms/src/lib/modules/radio/radio-group-id.service.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/**
+ * Tracks the element IDs for all radios within a radio group.
+ * @internal
+ */
+@Injectable()
+export class SkyRadioGroupIdService {
+  public get radioIds(): Observable<string[]> {
+    return this.#radioIdsObs;
```
Blackbaud-ErikaMcVey, line 12: should this be `#_radioIdsObs` since its used in a getter?

libs/components/forms/src/lib/modules/radio/radio-group.component.spec.ts
```diff
@@ -559,6 +559,33 @@ describe('Radio group component (reactive)', function () {
...
+
+    const radioGroupEl: HTMLDivElement =
+      fixture.nativeElement.querySelector('.sky-radio-group');
+
+    expect(radioGroupEl.getAttribute('aria-owns')).toMatch(
+      /sky-radio-sky-radio-[0-9]+-input sky-radio-sky-radio-[0-9]+-input sky-radio-sky-radio-[0-9]+-input/
+    );
+  });
+
+  it('should update aria-owns if radio IDs change', () => {
```
Blackbaud-ErikaMcVey: I think this test wording and the test could be improved. How are radio IDs changing? Adding new radios, removing old radios, both? And how exactly should the value change? 

If this were to fail 6 months from now it would take me a while to figure out what was supposed to be happening.

Blackbaud-ErikaMcVey (changes requested):

libs/components/forms/src/lib/modules/radio/radio-group-id.service.ts
```diff
@@ -0,0 +1,62 @@
+import { Injectable } from '@angular/core';
+
+import { BehaviorSubject, Observable } from 'rxjs';
+
+interface RegisterArgs {
```
Blackbaud-PaulCrowder: Since `SkyRadioGroupIdService` isn't part of the public API, I wouldn't worry about making an `args` interface. That's more for future-proofing public APIs for backwards compatibility.

libs/components/forms/src/lib/modules/radio/radio-group.component.spec.ts
```diff
@@ -559,6 +559,33 @@ describe('Radio group component (reactive)', function () {
...
     await expectAsync(fixture.nativeElement).toBeAccessible();
   });
+
+  it('should set aria-owns as a space-separated list of radio ids', () => {
+    fixture.detectChanges();
+
+    const radioGroupEl: HTMLDivElement =
+      fixture.nativeElement.querySelector('.sky-radio-group');
+
+    expect(radioGroupEl.getAttribute('aria-owns')).toMatch(
```
Blackbaud-PaulCrowder: Can this be a little more intentional instead of matching a pattern? e.g. inject or mock the `SkyRadioGroupIdService`, then validate that the `aria-owns` attribute value literally matches the IDs from the service.

libs/components/forms/src/lib/modules/radio/radio-group.component.spec.ts
```diff
@@ -559,6 +559,33 @@ describe('Radio group component (reactive)', function () {
...
+      fixture.nativeElement.querySelector('.sky-radio-group');
+
+    const originalAriaOwns = radioGroupEl.getAttribute('aria-owns');
+
+    fixture.componentInstance.options[0].id = 'foobar';
+    fixture.detectChanges();
+
+    const newAriaOwns = radioGroupEl.getAttribute('aria-owns');
+
+    expect(originalAriaOwns).not.toEqual(newAriaOwns);
```
Blackbaud-PaulCrowder: I'd be more explicit here, too. Check for the literal value you expect rather than whether it simply changed.

libs/components/forms/src/lib/modules/radio/radio-group-id.service.ts
Blackbaud-SteveBrush: For observable getters, we're using the `.asObservable()` pattern. For example: https://github.com/blackbaud/skyux/blob/main/libs/components/core/src/lib/modules/affix/affixer.ts#L47-L49. I'll update our internal wiki with an example.

libs/components/forms/src/lib/modules/radio/radio-group.component.spec.ts
Blackbaud-SteveBrush: Updated the wording.

libs/components/forms/src/lib/modules/radio/radio-group.component.ts
```diff
@@ -151,18 +163,28 @@ export class SkyRadioGroupComponent
   #_tabIndex: number | undefined;
 
   #changeDetector: ChangeDetectorRef;
+  #radioGroupIdSvc: SkyRadioGroupIdService;
   #ngControl: NgControl | undefined;
 
   constructor(
     changeDetector: ChangeDetectorRef,
+    @Host() radioGroupIdSvc: SkyRadioGroupIdService,
```
Blackbaud-PaulCrowder: Is there a particular reason we need `@Host` here? Is it possible that without it we'd somehow get an incorrect instance of `SkyRadioGroupIdService`? Since `SkyRadioGroupIdService` is provided by this class, it shouldn't ever look beyond the host component's injector anyway.

libs/components/forms/src/lib/modules/radio/radio.component.spec.ts
```diff
@@ -53,6 +53,20 @@ describe('Radio component', function () {
...
+      const onCheckedChangeSpy = spyOn(testComponent, 'onCheckedChange');
+      const radios: NodeListOf<HTMLInputElement> =
+        fixture.nativeElement.querySelectorAll('input');
+
+      // Select the second radio.
+      radios.item(1).click();
+      fixture.detectChanges();
+      await fixture.whenStable();
+
+      expect(onCheckedChangeSpy).toHaveBeenCalledWith(false);
```
Blackbaud-PaulCrowder: You can combine these two into `toHaveBeenCalledOnceWith()`.

Blackbaud-ErikaMcVey (dismissed):

libs/components/forms/src/lib/modules/radio/fixtures/radio.component.fixture.ts
```diff
@@ -23,13 +23,17 @@ export class SkyRadioTestComponent implements AfterViewInit {
...
   }
 
-  public onDisabledChange(value: boolean): void {}
+  public onClick(): void {
+    /* */
+  }
 
-  public onClick() {}
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public onDisabledChange(disabled: boolean): void {
```
Blackbaud-PaulCrowder: You can remove the `disabled` parameter to avoid the ESLint comment.

libs/components/forms/src/lib/modules/radio/fixtures/radio.component.fixture.ts
```diff
@@ -23,13 +23,17 @@ export class SkyRadioTestComponent implements AfterViewInit {
   @ViewChild(SkyRadioComponent)
   public checkboxComponent!: SkyRadioComponent;
 
-  public ngAfterViewInit() {
-    this.checkboxComponent.disabledChange.subscribe((value) => {
-      this.onDisabledChange(value);
-    });
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public onCheckedChange(checked: boolean): void {
```
Blackbaud-PaulCrowder: You can remove the `checked` parameter to avoid the ESLint comment.

libs/components/forms/src/lib/modules/radio/radio-group-id.service.ts
Blackbaud-PaulCrowder: It's not _exclusively_ used in a getter, so it can't use the `#_` prefix.

libs/components/forms/src/lib/modules/radio/fixtures/radio.component.fixture.html
Blackbaud-PaulCrowder: How does this help coverage? Should we be testing something we're not? If they are needed, you don't need to pass `$event` since the functions you're binding to don't do anything.

libs/components/forms/src/lib/modules/radio/radio.component.spec.ts
Blackbaud-SteveBrush: I didn't know about that one. Cool!

libs/components/forms/src/lib/modules/radio/fixtures/radio.component.fixture.ts
Blackbaud-SteveBrush: I removed the parameter based on your feedback. I got in the habit of including parameters in the callback function because Angular's strict template checking throws errors when the template implementation doesn't match the class's implementation (e.g. `onDisabledChange($event)` and `onDisabledChange(): void`). Since we're not running strict template checking on the fixtures, it'll be okay to remove the unused parameters.

libs/components/forms/src/lib/modules/radio/radio-group.component.ts
Blackbaud-SteveBrush: You're correct, it's redundant. Removed.

libs/components/forms/src/lib/modules/radio/radio-group.component.spec.ts
Blackbaud-SteveBrush: Good call.

libs/components/forms/src/lib/modules/radio/radio-group.component.spec.ts
Blackbaud-SteveBrush: Good idea. Since the component is generating the IDs using an auto-incrementing integer, I had to inject the `SkyIdService` into the component so I could spy on the ID generation process. Seems to work well. 👍

libs/components/forms/src/lib/modules/radio/radio-group-id.service.ts
Blackbaud-SteveBrush: I removed the interface and used two separate parameters instead.

Blackbaud-PaulCrowder (approved):

PR #689, opened 2022-10-14
refactor(components/tabs): tabs strict mode support

Blackbaud-TrevorBurch:
[AB#2195549](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2195549)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/689/storybook/)

Component Storybooks:

- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/689/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/689/storybooks/tabs-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/689/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/689/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/689/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/689/playground/)

libs/components/tabs/src/lib/modules/tabs/fixtures/tabset-active.component.fixture.html
```diff
@@ -9,7 +9,6 @@
     <sky-tab
       *ngIf="tab3Available"
       [tabHeading]="tab3Heading"
-      [tabHeaderCount]="tab3HeaderCount"
```
Blackbaud-TrevorBurch, line 12: This was removed as it wasn't being used (`tab3HeaderCount` didn't even exist in the backing TypeScript).

libs/components/tabs/src/lib/modules/tabs/fixtures/tabset-adapter.service.mock.ts
```diff
@@ -1,19 +1,40 @@
-import { Injectable } from '@angular/core';
+import { Injectable, OnDestroy } from '@angular/core';
+
+import { BehaviorSubject, Observable } from 'rxjs';
 
 import { SkyTabsetAdapterService } from '../tabset-adapter.service';
 
 @Injectable()
-export class MockTabsetAdapterService extends SkyTabsetAdapterService {
+export class MockTabsetAdapterService
```
Blackbaud-TrevorBurch, line 8: The changes to this file are because the previous functionality relied on `_overflowChange` and `currentOverflow` being accessible by the extending class. However, with the hash private variables that was no longer possible. I modified this class so that we can still use its basic functionality while still mocking the overflow changes.

libs/components/tabs/src/lib/modules/tabs/tab-button.component.ts
```diff
@@ -127,7 +127,7 @@ export class SkyTabButtonComponent implements AfterViewInit, OnDestroy {
     this.#tabsetService.focusedTabBtnIndex
       .pipe(distinctUntilChanged(), takeUntil(this.#ngUnsubscribe))
       .subscribe((focusedIndex) => {
-        if (focusedIndex === this.tabIndex) {
+        if (focusedIndex.toLocaleString() === this.tabIndex?.toLocaleString()) {
```
Blackbaud-TrevorBurch: This change to add `toLocaleString` was needed because `focusedIndex` was a string; however, `this.tabIndex` is of type `number | string | undefined` and so we can't do just a straight comparison as the types don't match. The `toLocaleString` allows us to do that comparison.

libs/components/tabs/src/lib/modules/tabs/tab.component.ts
```diff
@@ -119,19 +130,15 @@ export class SkyTabComponent implements OnChanges, OnDestroy {
   @Output()
   public close = new EventEmitter<void>(); // eslint-disable-line @angular-eslint/no-output-native
 
-  public get closeable(): boolean {
```
Blackbaud-TrevorBurch, line 123: This is only used internally and was better served as a function (we only call it one time from another typescript file) and so I changed it to be a function.

libs/components/tabs/src/lib/modules/tabs/tab.component.ts
```diff
@@ -61,14 +61,15 @@ export class SkyTabComponent implements OnChanges, OnDestroy {
    * This input only applies when the tabset's `tabStyle` is `"tabs"`.
    */
   @Input()
-  public set permalinkValue(value: string) {
-    this._permalinkValue = this.permalinkService.urlify(value);
+  public set permalinkValue(value: string | undefined) {
+    this.#permalinkValue = value;
+    this.#permalinkValueOrDefault =
```
Blackbaud-TrevorBurch: I went with a `OrDefault` variable here as the default relies on another variable that could change at a different time.

libs/components/tabs/src/lib/modules/tabs/tabset-nav-button.component.html
```diff
@@ -2,7 +2,13 @@
   *ngIf="isVisible"
   class="sky-btn sky-margin-inline-sm"
   [attr.aria-controls]="ariaControls"
-  [disabled]="disabled"
+  [disabled]="
+    disabled !== undefined
```
Blackbaud-TrevorBurch: The logic for this was in the getter previously; however, it was nasty. I don't love having this logic in the template though - so we can chat if needed.

libs/components/tabs/testing/src/tabs/tabset-fixture.spec.ts
```diff
@@ -164,11 +164,16 @@ describe('Tabset fixture', () => {
         .nativeElement
     ).toHaveCssClass('sky-btn-tab-selected');
 
-    let errorMessage: string;
+    let errorMessage: string | undefined;
 
     try {
       await tabset.clickTab(100);
     } catch (err) {
+      if (!(err instanceof Error)) {
```
Blackbaud-TrevorBurch: We have had to do this in other places recently. `err` is not typed as `Error` and so we have to add this check in order to check `err.message` below.

libs/components/tabs/src/lib/modules/tabs/fixtures/tabset-wizard.component.fixture.ts
```diff
@@ -5,21 +5,23 @@ import { Component } from '@angular/core';
...
 
   public selectedTab = 0;
 
-  public finishDisabled = undefined;
+  public finishDisabled: boolean | undefined = undefined;
 
   public passTabset = true;
 
-  public onSave: () => void;
+  public onSave: () => void = () => {
```
Blackbaud-PaulCrowder: Does this need an explicit type, or can you rely on type inference?

```
public onSave = () => { return; }
```

libs/components/tabs/src/lib/modules/tabs/tab-button.component.ts
Blackbaud-PaulCrowder: I'm not crazy about converting numbers to strings to compare them, especially when using `toLocaleString()` which introduces formatting (decimal separators, thousands separators, etc). What about this? This should preserve the previous behavior as long as `focusedIndex` is never `undefined`:

```
if (this.tabIndex !== undefined && focusedIndex === this.tabIndex) {
```

libs/components/tabs/src/lib/modules/tabs/tab.component.ts
```diff
@@ -87,28 +88,38 @@ export class SkyTabComponent implements OnChanges, OnDestroy {
    * @required
    */
   @Input()
-  public tabHeading: string;
+  public set tabHeading(value: string | undefined) {
+    this.#_tabHeading = value;
+
+    if (!this.#permalinkValue) {
```
Blackbaud-PaulCrowder: I'd expect this logic to be identical to the logic in the `permalinkValue` setter. I'd move that logic into a function that gets the value of `permalinkValue` and `tabHeading` and sets a `permalinkValueOrDefault` public property that is then referenced in the template. That will also let you go back to using the `#_permalinkValue` convention since it'll only be referenced in the `permalinkValue` accessors. Alternatively you could create a pure pipe that accepts `permalinkValue` and `tabHeading` and returns the value to use in the template. That would be a cleaner solution but would result in more code.

libs/components/tabs/src/lib/modules/tabs/tab.component.ts
```diff
@@ -140,26 +147,54 @@ export class SkyTabComponent implements OnChanges, OnDestroy {
...
+    tabsetService: SkyTabsetService
   ) {
+    this.#changeDetector = changeDetector;
+    this.#permalinkService = permalinkService;
+    this.#tabsetService = tabsetService;
     const id = nextId++;
     this.tabPanelId = `sky-tab-${id}`;
     this.tabButtonId = `${this.tabPanelId}-nav-btn`;
+
+    this.#_activeChange = new BehaviorSubject<void | undefined>(undefined);
```
Blackbaud-PaulCrowder: Don't access `#_` properties outside their corresponding get/set accessors.

libs/components/tabs/src/lib/modules/tabs/tabset-adapter.service.ts
```diff
@@ -1,34 +1,44 @@
...
 
-  private tabsOffsetLeft: number;
+  #tabsOffsetLeft: number | undefined;
 
-  private _overflowChange = new BehaviorSubject<boolean>(false);
+  #_overflowChange: BehaviorSubject<boolean>;
+  #_overflowChangeObs: Observable<boolean>;
+
+  constructor() {
+    this.#_overflowChange = new BehaviorSubject<boolean>(false);
```
Blackbaud-PaulCrowder: Don't reference `#_` outside their corresponding get/set accessors.

libs/components/tabs/src/lib/modules/tabs/tab.component.ts
```diff
@@ -30,27 +30,27 @@ export class SkyTabComponent implements OnChanges, OnDestroy {
...
   }
 
   public get active(): boolean {
-    return this._active || false;
+    return this.#_active;
   }
 
   public get activeChange(): Observable<void> {
-    return this._activeChange.asObservable();
+    return this.#_activeChangeObs;
```
Blackbaud-PaulCrowder: This seems backwards since there's also `this.#_activeChange` property. Since `activeChange` is only referenced internally, I don't think we need the `.asObservable()` version at all and can rely on the get accessor's `Observable` type to keep other code from calling non-`Observable` methods like `next()` on it.

libs/components/tabs/src/lib/modules/tabs/tabset-nav-button.component.html
Blackbaud-PaulCrowder: This is a good candidate for a pure pipe.

libs/components/tabs/src/lib/modules/tabs/tabset-permalink.service.ts
```diff
@@ -17,43 +17,56 @@ interface PermalinkParams {
...
-  ) {}
+    activatedRoute: ActivatedRoute,
+    location: Location,
+    router: Router
+  ) {
+    this.#activatedRoute = activatedRoute;
+    this.#location = location;
+    this.#router = router;
+
+    this.#_popStateChange = new Subject<void>();
```
Blackbaud-PaulCrowder: Don't reference `#_` outside their corresponding get/set accessors.

libs/components/tabs/src/lib/modules/tabs/tabset-permalink.service.ts
```diff
@@ -17,43 +17,56 @@ interface PermalinkParams {
 @Injectable()
 export class SkyTabsetPermalinkService implements OnDestroy {
   public get popStateChange(): Observable<void> {
-    return this._popStateChange.asObservable();
```
Blackbaud-PaulCrowder, line 20: This is another place where we don't need `asObservable()` since it's internal.

libs/components/tabs/src/lib/modules/tabs/tabset.service.ts
```diff
@@ -10,41 +10,62 @@ import { SkyTabsetActiveTabUnregisteredArgs } from './tabset-active-tab-unregist
...
-    tabIndex: SkyTabIndex;
+  #tabs: {
+    tabIndex: SkyTabIndex | undefined;
   }[] = [];
 
-  private tabCounter = 0;
+  #tabCounter = 0;
+
+  constructor() {
+    this.#_activeTabIndex = new BehaviorSubject<SkyTabIndex>(0);
```
Blackbaud-PaulCrowder: Don't reference `#_` outside their corresponding get/set accessors.

libs/components/tabs/src/lib/modules/tabs/tabset.service.ts
```diff
@@ -10,41 +10,62 @@ import { SkyTabsetActiveTabUnregisteredArgs } from './tabset-active-tab-unregist
  * @internal
  */
 @Injectable()
-export class SkyTabsetService {
+export class SkyTabsetService implements OnDestroy {
   public get activeTabUnregistered(): Observable<SkyTabsetActiveTabUnregisteredArgs> {
-    return this._activeTabUnregistered.asObservable();
```
Blackbaud-PaulCrowder, line 15: No need for `asObservable()` here or elsewhere in the code since it's internal.

libs/components/tabs/testing/src/tabs/tabset-fixture.spec.ts
```diff
@@ -164,11 +164,16 @@ describe('Tabset fixture', () => {
         .nativeElement
     ).toHaveCssClass('sky-btn-tab-selected');
 
-    let errorMessage: string;
+    let errorMessage: string | undefined;
 
     try {
```
Blackbaud-PaulCrowder: This whole section can be refactored with something like this. Same for the other instances.

```
await expectAsync(tabset.clickTab(100)).toBeRejectedWith('There is no tab at index 100.');
```

Blackbaud-PaulCrowder (commented):
I think I found all the places where this occurred, but if I didn't, be on the lookout for anywhere we're referencing `#_` properties outside their corresponding get/set accessors, as well as reevaluate whether we need `asObservable()` for an internal API.

libs/components/tabs/src/lib/modules/tabs/tab.component.ts
Blackbaud-TrevorBurch: I'm not sure what seemed backwards as I was following the convention that we've had for strict mode and that always relies on two properties. I got that from Steve - we can flex if needed. However, I do see what you are saying about these being internal so I changed it to only use the subject without having a field that uses `asObservable`.

libs/components/tabs/src/lib/modules/tabs/tabset-nav-button.component.html
Blackbaud-TrevorBurch: I think I did this right. Please let me know if not.

libs/components/tabs/src/lib/modules/tabs/tabset-nav-button-disabled.pipe.ts
```diff
@@ -0,0 +1,23 @@
...
+    buttonType?: SkyTabsetNavButtonType | string,
+    tabToSelect?: SkyTabComponent
+  ): boolean {
+    if (disabled !== undefined) {
+      return disabled;
+    } else if (buttonType === 'finish') {
+      return false;
+    }
+
+    return tabToSelect === undefined || !!tabToSelect.disabled;
```
Blackbaud-PaulCrowder: Checking a property value in a pure pipe could be problematic if the property value can change. Pure pipes are only called when their inputs change, so in the case where `tabToSelect.disabled` changes values but `tabToSelect` still points to the same object when the pipe was last called, this pipe won't be called. Instead of taking in `tabToSelect`, it should probably take in two `boolean` values, `tabToSelectExists` and `tabToSelectIsDisabled`. Then the template would look like this:

```
[disabled]="disabled | skyTabsetNavButtonDisabled: buttonType:!!tabToSelect:tabToSelect?.disabled"
```

libs/components/tabs/testing/src/tabs/tabset-fixture.spec.ts
Blackbaud-PaulCrowder: I don't see that this was changed anywhere.

libs/components/tabs/src/lib/modules/tabs/fixtures/tabset-wizard.component.fixture.ts
Blackbaud-PaulCrowder: This doesn't appear to have been addressed.

libs/components/tabs/src/lib/modules/tabs/tabset-adapter.service.ts
Blackbaud-PaulCrowder: This wasn't changed.

Blackbaud-PaulCrowder (approved):

PR #769, opened 2022-11-01
refactor(components/datetime): datepicker strict mode support

Blackbaud-TrevorBurch:
[AB#2195417](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2195417)
[AB#2353786](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2353786)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/769/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/769/storybooks/ag-grid-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/769/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/769/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/769/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/769/playground/)

apps/code-examples/src/app/code-examples/datetime/datepicker/basic/datepicker-demo.component.ts
```diff
@@ -9,13 +9,11 @@ import {
...
-export class DatepickerDemoComponent implements OnInit {
+export class DatepickerDemoComponent {
   public myForm: UntypedFormGroup;
 
-  constructor(private formBuilder: UntypedFormBuilder) {}
-
-  public ngOnInit(): void {
-    this.myForm = this.formBuilder.group({
+  constructor(formBuilder: UntypedFormBuilder) {
+    formBuilder.group({
```
johnhwhite: This should assign `this.myForm`, right?

Blackbaud-TrevorBurch: Yes. That was an accidental change. Fixed.

apps/playground/src/app/components/datetime/datepicker/datepicker.component.ts
```diff
@@ -18,33 +18,33 @@ import { delay, distinctUntilChanged } from 'rxjs/operators';
...
 })
-export class DatepickerComponent implements OnInit {
+export class DatepickerComponent {
   public dateFormat: string | undefined = undefined;
   public disabled = false;
   public minDate: Date | undefined;
   public maxDate: Date | undefined;
   public noValidate = false;
-  public reactiveForm: UntypedFormGroup | undefined;
+  public reactiveForm: UntypedFormGroup;
```
Blackbaud-SteveBrush, line 27: How difficult would it be to switch out the `Untyped*` types with the typed versions?

apps/playground/src/app/components/datetime/datepicker/datepicker.component.ts
```diff
@@ -18,33 +18,33 @@ import { delay, distinctUntilChanged } from 'rxjs/operators';
...
       selectedDate: new UntypedFormControl(
         new Date(1955, 10, 5),
         Validators.required
       ),
     });
+  }
+
+  public get reactiveDate(): AbstractControl {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    return this.reactiveForm.get('selectedDate')!;
```
Blackbaud-SteveBrush: Could you save the `selectedDate` as a private property of this class instead of using a non-null assertion?
```
#selectedDateControl = new FormControl(...);

this.reactiveForm = formBuilder.group({
  selectedDate: this.#selectedDate
});
```
I also wonder if using `FormGroup` with a type (and not `UntypedFormGroup`) would solve this. 🤔

libs/components/datetime/src/lib/modules/datepicker/date-formatter.ts
```diff
@@ -42,9 +42,9 @@ export class SkyDateFormatter {
     return momentValue.toDate();
   }
 
-  public dateIsValid(date: Date): boolean {
+  public dateIsValid(date: Date | undefined): boolean {
     return (
-      date &&
+      !!date &&
```
Blackbaud-SteveBrush: Is checking for truthy sufficient? Why convert to a `boolean` in this check?

libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar-inner.component.ts
```diff
@@ -151,6 +158,7 @@ export class SkyDatepickerCalendarInnerComponent
   }
 
   public compare(date1: Date, date2: Date): undefined | number {
+    /* istanbul ignore if */
```
Blackbaud-SteveBrush: It's rare that we need to ignore `if` checks for coverage. Why can't we test this? (See comment below.)

libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar-inner.component.ts
```diff
@@ -34,27 +34,33 @@ export class SkyDatepickerCalendarInnerComponent
...
+  public maxDate: Date | undefined;
 
   @Input()
-  public set selectedDate(value: Date) {
+  public set selectedDate(value: Date | undefined) {
     if (this.dateFormatter.dateIsValid(value)) {
-      this._selectedDate = value;
-      this.activeDate = value;
+      this.#_selectedDate = value;
+      this.activeDate = value!;
```
Blackbaud-SteveBrush: Can you remove the non-null assertion and instead modify the condition?
```
if (value && this.dateFormatter.dateIsValid(value)) {
```

Blackbaud-SteveBrush: Also, what happens if the user sets the input deliberately to `undefined`? Shouldn't it reset the selected and active dates?

libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar-inner.component.ts
```diff
@@ -406,18 +418,19 @@ export class SkyDatepickerCalendarInnerComponent
...
   protected isDisabled(date: Date): boolean {
-    const customDate = this.getCustomDate(date);
+    const customDate = this.#getCustomDate(date);
     return (
-      (this.minDate && this.compare(date, this.minDate) < 0) ||
-      (this.maxDate && this.compare(date, this.maxDate) > 0) ||
-      (customDate && customDate.disabled)
+      (!!this.minDate && date && this.compare(date, this.minDate)! < 0) ||
+      (!!this.maxDate && date && this.compare(date, this.maxDate)! > 0) ||
+      (!!customDate && !!customDate.disabled)
```
Blackbaud-SteveBrush: Instead of using non-null assertions, maybe we could rework the conditions a bit. This would also allow you to get coverage for the `/* istanbul ignore if */` statement above.
```
protected isDisabled(date: Date): boolean {
  const customDate = this.#getCustomDate(date);
  const minDateCompare = this.compare(date, this.minDate);
  const maxDateCompare = this.compare(date, this.maxDate);

  return (
    (minDateCompare !== undefined && minDateCompare < 0) ||
    (maxDateCompare !== undefined && maxDateCompare > 0) ||
    !!customDate?.disabled
  );
}
```

libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar-inner.component.ts
```diff
@@ -168,6 +176,9 @@ export class SkyDatepickerCalendarInnerComponent
     if (this.datepickerMode === 'year' && this.compareHandlerYear) {
       return this.compareHandlerYear(date1, date2);
     }
+
+    /* istanbul ignore next */
```
Blackbaud-SteveBrush, line 186: You might be able to remove this line, too, after the adjustments described below.

libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar-inner.component.ts
```diff
@@ -151,6 +158,7 @@ export class SkyDatepickerCalendarInnerComponent
   }
 
   public compare(date1: Date, date2: Date): undefined | number {
```
Blackbaud-SteveBrush: Can we switch this return signature to `number | undefined`?

libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar.component.ts
```diff
@@ -24,35 +22,37 @@ import { SkyDatepickerCustomDate } from './datepicker-custom-date';
...
   @Input()
-  public selectedDate: Date;
+  public selectedDate: Date | undefined;
 
   /** starting day of the week from 0-6 (0=Sunday, ..., 6=Saturday) */
   @Input()
-  public set startingDay(start: number) {
-    this._startingDay = start;
+  public set startingDay(start: number | undefined) {
+    this.#_startingDay = start || 0;
```
Blackbaud-SteveBrush: Might be better to use `??` since the `start` value could be `0` which evaluates as falsy.

libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar.component.ts
```diff
@@ -24,35 +22,37 @@ import { SkyDatepickerCustomDate } from './datepicker-custom-date';
...
 
-  public get startingDay() {
-    return this._startingDay || 0;
+  public get startingDay(): number {
+    return this.#_startingDay;
   }
 
   @Output()
-  public calendarDateRangeChange: EventEmitter<SkyDatepickerCalendarChange> = new EventEmitter<SkyDatepickerCalendarChange>();
+  public calendarDateRangeChange: EventEmitter<
```
Blackbaud-SteveBrush: You can simplify this:
```
public calendarDateRangeChange = new EventEmitter<
  SkyDatepickerCalendarChange | undefined
>();
```

libs/components/datetime/src/lib/modules/datepicker/monthpicker.component.html
```diff
@@ -29,10 +29,6 @@
           type="button"
           class="sky-btn sky-btn-default sky-btn-sm sky-datepicker-calendar-title"
           (click)="datepicker.toggleModeCalendar($event)"
-          [disabled]="datepicker.datepickerMode === maxMode"
-          [ngClass]="{
-            'sky-btn-disabled': datepicker.datepickerMode === maxMode
-          }"
```
Blackbaud-SteveBrush, line 35: What's the reasoning for removing the disabled states?

libs/components/datetime/src/lib/modules/datepicker/fuzzy-date.service.ts
```diff
@@ -33,38 +33,38 @@ interface SkyFuzzyDateRange {
   providedIn: 'root',
 })
 export class SkyFuzzyDateService implements OnDestroy {
-  private currentLocale: string;
+  #currentLocale = 'en-US';
```
Blackbaud-SteveBrush: This smells a bit, because now we're saving the default locale in fuzzy date service, but that value is owned by the locale provider. Instead, the default locale can be derived from the locale provider directly.
```
#currentLocale: string;

constructor(...) {
  this.#currentLocale = localeProvider.defaultLocale;
}
```

libs/components/datetime/src/lib/modules/datepicker/datepicker-input-fuzzy.directive.ts
```diff
@@ -166,125 +171,145 @@ export class SkyFuzzyDatepickerInputDirective
...
   }
 
-  private control: AbstractControl;
+  #control: AbstractControl | undefined;
 
-  private dateFormatter = new SkyDateFormatter();
+  #dateFormatter = new SkyDateFormatter();
 
-  private locale: string;
+  #locale = 'en-US';
```
Blackbaud-SteveBrush: Default locale should be derived from the locale provider's `defaultLocale` property.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input-fuzzy.directive.ts
```diff
@@ -166,125 +171,145 @@ export class SkyFuzzyDatepickerInputDirective
...
-    private resourcesService: SkyLibResourcesService,
-    @Optional() private datepickerComponent: SkyDatepickerComponent
+    changeDetector: ChangeDetectorRef,
+    configService: SkyDatepickerConfigService,
+    elementRef: ElementRef,
+    fuzzyDateService: SkyFuzzyDateService,
+    localeProvider: SkyAppLocaleProvider,
+    renderer: Renderer2,
+    resourcesService: SkyLibResourcesService,
+    @Optional() datepickerComponent: SkyDatepickerComponent
```
Blackbaud-SteveBrush: Can you put a question mark on this parameter?

libs/components/datetime/src/lib/modules/datepicker/datepicker-input-fuzzy.directive.ts
```diff
@@ -166,125 +171,145 @@ export class SkyFuzzyDatepickerInputDirective
...
+
+  #changeDetector: ChangeDetectorRef;
+  #configService: SkyDatepickerConfigService;
+  #elementRef: ElementRef;
+  #fuzzyDateService: SkyFuzzyDateService;
+  #localeProvider: SkyAppLocaleProvider;
+  #renderer: Renderer2;
+  #resourcesService: SkyLibResourcesService;
+  // NOTE: Though it is marked as `@Optional` we throw a custom error in `ngOnInit` if this value is not injected.
+  #datepickerComponent!: SkyDatepickerComponent;
```
Blackbaud-SteveBrush: Can you move the error to the constructor so that you don't need the non-null assertion?

libs/components/datetime/src/lib/modules/datepicker/datepicker-input-fuzzy.directive.ts
```diff
@@ -484,110 +509,115 @@ export class SkyFuzzyDatepickerInputDirective
...
   }
 
   /* istanbul ignore next */
-  private fuzzyDatesEqual(dateA: SkyFuzzyDate, dateB: SkyFuzzyDate): boolean {
+  #fuzzyDatesEqual(dateA?: SkyFuzzyDate, dateB?: SkyFuzzyDate): boolean {
     return (
-      dateA &&
-      dateB &&
+      !!dateA &&
+      !!dateB &&
```
Blackbaud-SteveBrush: Just for the sake of code readability, instead of converting these dates to booleans, I would check if they're undefined.
```
dateA !== undefined &&
dateB !== undefined &&
```

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -192,91 +192,108 @@ export class SkyDatepickerInputDirective
...
-    private resourcesService: SkyLibResourcesService,
-    @Optional() private datepickerComponent: SkyDatepickerComponent
+    adapter: SkyDatepickerAdapterService,
+    changeDetector: ChangeDetectorRef,
+    configService: SkyDatepickerConfigService,
+    elementRef: ElementRef,
+    localeProvider: SkyAppLocaleProvider,
+    renderer: Renderer2,
+    resourcesService: SkyLibResourcesService,
+    @Optional() datepickerComponent: SkyDatepickerComponent
```
Blackbaud-SteveBrush: Can you put a question mark on this parameter?

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -192,91 +192,108 @@ export class SkyDatepickerInputDirective
...
+  #_value: any;
+
+  #adapter: SkyDatepickerAdapterService;
+  #changeDetector: ChangeDetectorRef;
+  #configService: SkyDatepickerConfigService;
+  #elementRef: ElementRef;
+  #localeProvider: SkyAppLocaleProvider;
+  #renderer: Renderer2;
+  #resourcesService: SkyLibResourcesService;
+  #datepickerComponent!: SkyDatepickerComponent;
```
Blackbaud-SteveBrush: Instead of using a non-null assertion, maybe you could put the error check in the constructor?

libs/components/datetime/src/lib/modules/datepicker/daypicker-cell.component.ts
```diff
@@ -27,124 +27,137 @@ export class SkyDayPickerCellComponent implements OnInit, OnDestroy {
...
   @Input()
-  public activeDateHasChanged: boolean;
+  public activeDateHasChanged: boolean | undefined;
 
   /**
    * Specifies the date this picker cell will represent on the calendar.
    */
   @Input()
-  public date: SkyDatepickerDate;
+  public date!: SkyDatepickerDate;
```
Blackbaud-SteveBrush: Eh?

libs/components/datetime/src/lib/modules/datepicker/daypicker-cell.component.ts
```diff
@@ -27,124 +27,137 @@ export class SkyDayPickerCellComponent implements OnInit, OnDestroy {
...
+  ) {
+    this.#datepicker = datepicker;
+    this.#datepickerService = datepickerService;
+  }
 
   public ngOnInit(): void {
     this.hasTooltip =
-      this.date.keyDate &&
-      this.date.keyDateText &&
+      !!this.date &&
```
Blackbaud-SteveBrush: For code readability, instead of coercing undefined non-boolean values to boolean, you could coerce the entire condition like so:
```
this.hasTooltip = !!(
  this.date &&
  this.date.keyDate &&
  this.date.keyDateText &&
  this.date.keyDateText.length > 0 &&
  this.date.keyDateText[0].length > 0
);
```

libs/components/datetime/src/lib/modules/datepicker/daypicker-cell.component.ts
```diff
@@ -27,124 +27,137 @@ export class SkyDayPickerCellComponent implements OnInit, OnDestroy {
...
+      this.#hideTooltip();
+      this.#cancelPopover = false;
     }
   }
 
   public getKeyDateLabel(): string {
     if (this.hasTooltip) {
-      return this.date.keyDateText.join(', ');
+      // Safety check
+      /* istanbul ignore next */
```
Blackbaud-SteveBrush: Why can't this be tested?

libs/components/datetime/src/lib/modules/datepicker/daypicker-cell.component.spec.ts
```diff
@@ -124,60 +113,80 @@ describe('daypicker cell', () => {
...
+    }));
 
-    it('should not show tooltip', () => {
+    it('should show tooltip', fakeAsync(() => {
       fixture.detectChanges();
+      tick(500);
 
-      expect(showSpy).toHaveBeenCalled();
-    });
+      expect(controlerSpy).toHaveBeenCalled();
```
Blackbaud-SteveBrush: `controlerSpy` throughout.

libs/components/datetime/src/lib/modules/datepicker/daypicker.component.ts
```diff
@@ -34,57 +34,59 @@ interface SkyDateRange {
...
     /* istanbul ignore else */
     if (value) {
-      this.applyCustomDates(value, this.rows);
+      this.#applyCustomDates(value, this.rows);
     }
   }
 
   @Output()
-  public calendarDateRangeChange: EventEmitter<SkyDatepickerCalendarChange> = new EventEmitter<SkyDatepickerCalendarChange>();
+  public calendarDateRangeChange: EventEmitter<
```
Blackbaud-SteveBrush: Can be simplified to 
```
public calendarDateRangeChange = new EventEmitter<SkyDatepickerCalendarChange | undefined>();
```

libs/components/datetime/src/lib/modules/datepicker/daypicker.component.ts
```diff
@@ -269,12 +277,12 @@ export class SkyDayPickerComponent implements OnDestroy, OnInit {
     }
 
     return (
-      this.compareDays(rangeA.startDate, rangeB.startDate) === 0 &&
-      this.compareDays(rangeA.endDate, rangeB.endDate) === 0
+      this.#compareDays(rangeA!.startDate, rangeB!.startDate) === 0 &&
+      this.#compareDays(rangeA!.endDate, rangeB!.endDate) === 0
```
Blackbaud-SteveBrush: I would try to adjust the code to remove these non-null assertions. Possible?

libs/components/datetime/src/lib/modules/datepicker/daypicker.component.ts
```diff
@@ -257,9 +265,9 @@ export class SkyDayPickerComponent implements OnDestroy, OnInit {
     }
   }
 
-  private dateRangeRowsAreEqual(
-    rangeA: SkyDateRange,
-    rangeB: SkyDateRange
+  #dateRangeRowsAreEqual(
+    rangeA: SkyDateRange | undefined,
+    rangeB: SkyDateRange | undefined
```
Blackbaud-SteveBrush: `rangeA?: SkyDateRange` and `rangeB?: SkyDateRange`

libs/components/datetime/src/lib/modules/datepicker/datepicker.component.ts
```diff
@@ -238,37 +217,39 @@ export class SkyDatepickerComponent implements OnDestroy, OnInit {
...
+      this.#closePicker();
     } else {
-      this.openPicker();
+      this.#openPicker();
     }
   }
 
-  public onCalendarDateRangeChange(event: SkyDatepickerCalendarChange): void {
+  public onCalendarDateRangeChange(
+    event: SkyDatepickerCalendarChange | undefined
```
Blackbaud-SteveBrush: `event?: SkyDatepickerCalendarChange`

libs/components/datetime/src/lib/modules/datepicker/datepicker.component.ts
```diff
@@ -281,136 +262,136 @@ export class SkyDatepickerComponent implements OnDestroy, OnInit {
...
-  private addKeydownListner(): void {
-    this.overlayKeydownListner = fromEvent(window.document, 'keydown')
-      .pipe(takeUntil(this.ngUnsubscribe))
-      .subscribe((event: KeyboardEvent) => {
-        const key = event.key?.toLowerCase();
+  #addKeydownListner(): void {
+    this.#overlayKeydownListner = fromEvent(window.document, 'keydown')
+      .pipe(takeUntil(this.#ngUnsubscribe))
+      .subscribe((event) => {
+        const key = (event as KeyboardEvent).key?.toLowerCase();
```
Blackbaud-SteveBrush: The `fromEvent` operator takes generics:
```
fromEvent<KeyboardEvent>(`
```

libs/components/datetime/src/lib/modules/datepicker/datepicker.component.ts
```diff
@@ -281,136 +262,136 @@ export class SkyDatepickerComponent implements OnDestroy, OnInit {
...
+        .pipe(takeUntil(this.#calendarUnsubscribe))
+        .subscribe(() => {
+          /* istanbul ignore else */
+          if (this.isOpen) {
+            this.#closePicker();
+          }
+        });
 
-    overlay.attachTemplate(this.calendarTemplateRef);
+      this.#addKeydownListner();
```
Blackbaud-SteveBrush: `addKeydownListner` (spelling)

apps/playground/src/app/components/datetime/datepicker/datepicker.component.ts
Blackbaud-TrevorBurch: I would prefer to not attempt to tackle that in this PR. We haven't been doing changes in those types in strict mode stories and this close to 7 release time I don't know that I want to tackle that as scope creep here.

libs/components/datetime/src/lib/modules/datepicker/date-formatter.ts
Blackbaud-TrevorBurch: The double bang is needed here for typing. Without it - strict mode yells that `undefined` is not a valid return type.

libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar-inner.component.ts
Blackbaud-TrevorBurch: I made the change. I'm not going to modify how the original code worked in terms of resetting the dates in this PR. If we want to have the discussion that is fine - but that would change the original intention of the code as written and I don't want to do that in this PR.

libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar-inner.component.ts
Blackbaud-TrevorBurch: I had to leave this line. Strict mode yells without the fallback case at the end here; however, all cases are handled by the if-statements so this last line is never hit.

libs/components/datetime/src/lib/modules/datepicker/date-formatter.ts
Blackbaud-TrevorBurch: I coerced the entire thing instead of the `date` parameter

libs/components/datetime/src/lib/modules/datepicker/daypicker-cell.component.ts
Blackbaud-TrevorBurch: Because `getKeyDateLabel` is only ever called from something within an `ngIf=hasTooltip` which check from `keyDateText` not being `undefined` or an empty array. So we need the ternary here for strict mode; however, the `: ''` case will never be hit because of the wrapping code outside of it.

libs/components/datetime/src/lib/modules/datepicker/monthpicker.component.html
Blackbaud-TrevorBurch: So I added this back in this commit and fixed an issue where `maxMode` should have been `datepicker.maxMode`. However, this is somewhat dead code today as we alway have `maxMode` set to `year` and so this will never be hit. But in the spirit of not being destructive - I added it back.

libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar-inner.component.ts
Blackbaud-SteveBrush: During our last team discussion, we agreed we'd make "onceability changes, within reason" (listed as one of the items in our strict mode wiki). How much effort would it take to make the change? At the very least we need to create a bug in ADO.

libs/components/datetime/src/lib/modules/datepicker/daypicker-cell.component.ts
Blackbaud-SteveBrush: Couldn't you adjust the parent condition and simplify all of it into a single ternary and still get 100% coverage?
```
return this.hasTooltip && this.date?.keyDateText
  ? this.date.keyDateText.join(', ')
  : '';
```

libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar-inner.component.ts
Blackbaud-TrevorBurch: I made a change here to handle the `else` case in a better fashion. However, `activeDate` needs to be set to `new Date()`. I don't love this - but the existing code always set `activeDate` to `new Date()` if no selected date was given on initialization and to change things to let `activeDate` be `undefined` will take a lot of work as most of the other inner components are expecting there to always be an `activeDate`. I have put a TODO in the code to revisit this at some pont.

Blackbaud-SteveBrush (approved):

PR #778, opened 2022-11-02
refactor(components/datetime): date range picker strict mode support

Blackbaud-TrevorBurch:
[AB#2195413](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2195413)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/778/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/778/storybooks/ag-grid-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/778/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/778/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/778/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/778/playground/)

libs/components/datetime/src/lib/modules/date-range-picker/fixtures/date-range-picker.component.fixture.ts
```diff
@@ -17,41 +17,45 @@ import { SkyDateRangeCalculatorId } from '../types/date-range-calculator-id';
...
-  public initialValue: SkyDateRangeCalculation;
-  public label: string;
+  public initialValue: SkyDateRangeCalculation | undefined;
+  public label: string | undefined;
   public numValueChangeNotifications = 0;
-  public reactiveForm: UntypedFormGroup;
+  public reactiveForm: UntypedFormGroup | undefined;
   public startDateRequired = false;
-  public templateDisable: boolean = undefined;
+  public templateDisable: boolean | undefined = undefined;
```
Blackbaud-PaulCrowder: This doesn't need `= undefined` since it's declared as `boolean | undefined`.

libs/components/datetime/src/lib/modules/date-range-picker/fixtures/date-range-picker.component.fixture.ts
```diff
@@ -17,41 +17,45 @@ import { SkyDateRangeCalculatorId } from '../types/date-range-calculator-id';
...
-  constructor(private formBuilder: UntypedFormBuilder) {}
+  #formBuilder: UntypedFormBuilder;
+
+  constructor(formBuilder: UntypedFormBuilder) {
+    this.#formBuilder = formBuilder;
+  }
 
   public ngOnInit(): void {
-    this.reactiveForm = this.formBuilder.group({
+    this.reactiveForm = this.#formBuilder.group({
```
Blackbaud-PaulCrowder: I would move this into the constructor. Then you can get rid of the `| undefined` on `reactiveForm` as well as the private `#formBuilder` property.

Blackbaud-TrevorBurch: So the unit tests are currently written in a way that expects the ability to set an initial value between the construction and the first `detectChanges` being called. It would be possible to rewrite things but I think it would be more work than it would be worth for just a component fixture, thoughts?

Blackbaud-PaulCrowder: This sounds like a borderline onceability issue. The only reason it's not clear cut is because it's the "initial" value, so I can kind of rationalize requiring setting that before the component renders. However, that shouldn't keep us from being able to create `reactiveForm` in the constructor. We can create in the constructor and then set the value to the initial value in `ngOnInit()`.

Blackbaud-TrevorBurch: I took a look at this and think the couple of tests that were doing this weren't really necessary as other things cover what they were doing and they just complicated things. I decided to remove the tests and modified one other tests that could be done a slightly different way. Let me know if you have concerns.

Blackbaud-PaulCrowder (approved):

PR #811, opened 2022-11-10
chore(components/tiles): tiles strict mode support

Blackbaud-CoreyArcher:
[AB#2195565](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2195565)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/811/storybook/)

Component Storybooks:

- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/811/storybooks/tiles-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/811/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/811/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/811/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/811/playground/)

libs/components/tiles/src/lib/modules/tiles/tile/tile.component.ts
```diff
@@ -46,7 +46,7 @@ export class SkyTileComponent implements OnDestroy {
    * Specifies a human-readable name for the tile that is available for multiple purposes, such as accessibility and instrumentation. For example, the component uses the name to construct ARIA labels for tile controls to [support accessibility](https://developer.blackbaud.com/skyux/learn/accessibility).
    */
   @Input()
-  public tileName;
+  public tileName: any;
```
Blackbaud-SteveBrush: Is the intent of the code that the `tileName` be a string instead (see the fixture's setting it to `'test'` above)?

libs/components/tiles/src/lib/modules/tiles/tile-dashboard/fixtures/tile-context.fixture.ts
```diff
@@ -1,3 +1,3 @@
 export class TileTestContext {
-  public id: string;
+  public id?: string;
```
Blackbaud-SteveBrush: This is a class, so it should be `public id: string | undefined`.

libs/components/tiles/src/lib/modules/tiles/tile-dashboard/tile-dashboard.component.html
```diff
@@ -1,20 +1,22 @@
...
-<sky-tile-dashboard-column
-  class="sky-tile-dashboard-layout-single"
-  #singleColumn
->
-</sky-tile-dashboard-column>
-<span class="sky-screen-reader-only" [id]="moveInstructionsId">{{
-  'skyux_tile_move_instructions' | skyLibResources
-}}</span>
-<ng-content></ng-content>
+<ng-container *ngIf="config">
```
Blackbaud-SteveBrush, line 1: Good call.

libs/components/tiles/src/lib/modules/tiles/tile-dashboard/tile-dashboard.component.ts
```diff
@@ -76,44 +76,49 @@ export class SkyTileDashboardComponent implements AfterViewInit, OnDestroy {
...
-  private ngUnsubscribe = new Subject<void>();
-
-  private viewReady = false;
+  #configSet = false;
+  #dashboardService: SkyTileDashboardService;
+  #mediaQueryService: SkyMediaQueryService;
+  #ngUnsubscribe = new Subject<void>();
+  #resourcesService: SkyLibResourcesService | undefined;
+  #viewReady = false;
+  #_config!: SkyTileDashboardConfig;
```
Blackbaud-SteveBrush: We need to figure out a way to remove this non-null assertion because the config can technically be undefined at any moment. For example, what happens if the consumer wants to complete an asynchronous task before setting the config after `ngOnInit`? If any of this class's methods access the `config` value, they will throw an unhandled error. (From the looks of it, you're already adding `if` checks for the config anyway?)

libs/components/tiles/src/lib/modules/tiles/tile-dashboard/tile-dashboard.service.ts
```diff
@@ -45,28 +47,38 @@ export class SkyTileDashboardService {
...
-    private dragulaService: DragulaService,
-    private mediaQuery: SkyMediaQueryService,
-    private uiConfigService: SkyUIConfigService
+    dragulaService: DragulaService,
+    mediaQuery: SkyMediaQueryService,
+    uiConfigService: SkyUIConfigService,
+    @Optional() dynamicComponentService?: SkyDynamicComponentService
   ) {
-    this.bagId = 'sky-tile-dashboard-bag-' + ++bagIdIndex;
+    if (!dynamicComponentService) {
```
Blackbaud-SteveBrush, line 68: Was there something about strict mode that requires us to use `SkyDynamicComponentService`?

libs/components/tiles/src/lib/modules/tiles/tile-dashboard/fixtures/tile-context.fixture.ts
Blackbaud-CoreyArcher: Fixed

libs/components/tiles/src/lib/modules/tiles/tile-dashboard/tile-dashboard.service.ts
Blackbaud-CoreyArcher: I don't recall. I'm not seeing anything, so I think it was just something we saw that needed doing and did it.

libs/components/tiles/src/lib/modules/tiles/tile-dashboard/tile-dashboard.component.ts
Blackbaud-CoreyArcher: Making the config nullable seems to be ok. Made the change

libs/components/tiles/src/lib/modules/tiles/tile/tile.component.ts
Blackbaud-CoreyArcher: I didn't want to make assumptions, so since it was implictly `any` I just kept it that way. I would think it should be string though?

Blackbaud-SteveBrush: I think it's a safe assumption. A search for `tileName` in the monorepo only has it being used as a string.

libs/components/tiles/src/lib/modules/tiles/tile-dashboard/tile-dashboard.service.ts
```diff
@@ -45,28 +47,38 @@ export class SkyTileDashboardService {
...
+      console.warn(
+        'The SkyTileDashboardService was created without a reference to the SkyDynamicComponentService. The service will be unable to create tiles from the dashboard configuration. Constructing the SkyTileDashboardService without a SkyDynamicComponentService will be removed in a future breaking change.'
+      );
+    }
 
-    this.initDragula();
+    this.#dragulaService = dragulaService;
+    this.#dynamicComponentService = dynamicComponentService!;
```
Blackbaud-SteveBrush: Let's remove the non-null, here, and add an `if` check when we use the service, below, so that our source code reflects the optional nature of the property's signature.
```
/* istanbul ignore else */
if (tile && this.#dynamicComponentService) {
```

libs/components/tiles/src/lib/modules/tiles/tile/tile.component.ts
Blackbaud-CoreyArcher: Done

libs/components/tiles/src/lib/modules/tiles/tile-dashboard/tile-dashboard.service.ts
Blackbaud-CoreyArcher: Updated

Blackbaud-SteveBrush (approved):

PR #843, opened 2022-11-17
fix(components/modals): use `Provider[]` instead of `StaticProvider[]` for modal configuration

Blackbaud-SteveBrush:


Blackbaud-SteveBrush:
Closing since using `StaticProvider[]` is correct.

PR #994, opened 2023-02-15
feat(components/indicators): icon test harness

Blackbaud-SandhyaRajasabeson:
[AB#2195310](https://blackbaud.visualstudio.com/Products/_boards/board/t/SKY%20UX%20Program/Stories/?workitem=2195310)
[AB#2271323](https://blackbaud.visualstudio.com/Products/_boards/board/t/SKY%20UX%20Program/Stories/?workitem=2271323)
[AB#2400134](https://blackbaud.visualstudio.com/Products/_boards/board/t/SKY%20UX%20Program/Stories/?workitem=2400134)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/994/storybook/)

Component Storybooks:

- [a11y-storybook](https://blackbaud.github.io/skyux-pr-preview/994/storybooks/a11y-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/994/storybooks/ag-grid-storybook/)
- [core-storybook](https://blackbaud.github.io/skyux-pr-preview/994/storybooks/core-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/994/storybooks/indicators-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/994/storybooks/layout-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/994/storybooks/modals-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/994/storybooks/pages-storybook/)
- [popovers-storybook](https://blackbaud.github.io/skyux-pr-preview/994/storybooks/popovers-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/994/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/994/storybooks/tabs-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/994/storybooks/theme-storybook/)
- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/994/storybooks/tiles-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/994/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/994/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/994/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/994/playground/)

libs/components/indicators/testing/src/icon/icon-harness.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+    return { iconHarness, fixture, loader, pageLoader };
+  }
+
+  it('should setup the icon test harness', async () => {
+    const { iconHarness, fixture } = await setupTest();
+
+    await validateIconName(
+      iconHarness,
+      fixture,
+      fixture.componentInstance.iconName
```
Blackbaud-SteveBrush: These tests would be easier to follow if the icon's name was passed into the `validateIconName` function deliberately. I found myself scrolling up to your component fixture several times to see which of its properties the test was expecting.
```
await validateIconName(
  iconHarness,
  fixture,
  'house'
);
```

libs/components/indicators/testing/src/icon/icon-harness.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+  it('should return the correct icon name for all icon types and variants', async () => {
+    const { iconHarness, fixture } = await setupTest();
+    for (const type of iconTypes) {
+      fixture.componentInstance.iconType = type;
+      for (const variant of variants) {
+        fixture.componentInstance.variant = variant;
+        await validateIconName(
+          iconHarness,
+          fixture,
+          fixture.componentInstance.iconName
```
Blackbaud-SteveBrush, line 135: Can we change `fixture.componentInstance.iconName` to something deliberate, like `'house'`? 

Also, I don't understand the purpose of wrapping this test in two `for` loops. Is this test specifically written to prove that variant and type have no effect on the name being returned by the harness?

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,80 @@
...
+    const size = iconClass
+      ?.split(' ')
+      .find((classValue: string) =>
+        classValue.match(/^fa-(?=2xs|xs|sm|lg|2xl|[0-9]+x)/)
+      );
+    return size?.replace('fa-', '');
+  }
+
+  /** Gets if the icon is a variant */
+  public async getVariant(): Promise<string | undefined> {
```
Blackbaud-SteveBrush, line 101: ☝️ It would be easier to find a specific method in the future if the methods were listed alphabetically.

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,80 @@
...
+    return icon.hasClass(`fa-fw`);
+  }
+
+  /** Gets the icon type */
+  public async getIconType(): Promise<string> {
+    const iconClass = await (await this.#getIcon()).getAttribute('class');
+    const classes = iconClass
+      ?.split(' ')
+      .find((classValue: string) => /^sky-i-/.test(classValue));
+    if (classes) return 'skyux';
```
Blackbaud-SteveBrush: For the sake of code consistency, let's always wrap code blocks in curly braces.

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  /** Gets if the icon has fixed width */
+  public async isFixedWidth(): Promise<boolean> {
+    const icon = await this.#getIcon();
+    return icon.hasClass(`fa-fw`);
+  }
+
+  /** Gets the icon type */
+  public async getIconType(): Promise<string> {
+    const iconClass = await (await this.#getIcon()).getAttribute('class');
```
Blackbaud-PaulCrowder: Did you try switching to `getProperty('classList')` instead of splitting the attribute on a space? that will simplify the code a bit.

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,80 @@
...
+      .find((classValue: string) => /^sky-i-/.test(classValue));
+    if (classes) return 'skyux';
+    return 'fa';
+  }
+
+  /** Gets the icon size */
+  public async getIconSize(): Promise<string | undefined> {
+    const iconClass = await (await this.#getIcon()).getAttribute('class');
+    const size = iconClass
+      ?.split(' ')
```
Blackbaud-SteveBrush: You're using this logic in multiple places. Maybe you could simplify and get the `classList` property from a new private method?

```
async #getClassList(): Promise<DOMTokenList> {
  return (await this.#getIcon()).getProperty('classList');
}
```
You could then pull it into each method like this:
```
const iconClasses = await this.#getClassList();
```

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,80 @@
...
+  public static with(
+    filters: SkyIconHarnessFilters
+  ): HarnessPredicate<SkyIconHarness> {
+    return SkyIconHarness.getDataSkyIdPredicate(filters);
+  }
+
+  /** Gets the icon name */
+  public async getIconName(): Promise<string | undefined> {
+    const iconClass = await (await this.#getIcon()).getAttribute('class');
+    const regex = /^sky-i-|fa-(?!fw|2xs|xs|sm|lg|2xl|[0-9]+x)/;
```
Blackbaud-SteveBrush: When I'm dealing with complex regular expressions, I tend to do one of two things, either:
1. Create a top-level `const` that has a descriptive name (e.g. `const ICON_CLASS_REGEXP = /.../`).
2. Or, add a helpful comment above the regular expression so that future engineers can understand what it's doing at a glance.

I usually prefer the former since it makes the rest of the code much more readable, but obviously it's a judgment call for each type of usage.

libs/components/indicators/testing/src/icon/icon-harness.ts
Blackbaud-SandhyaRajasabeson: i tried that last week and wasn't able to get half of the functions at this link https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/entries to work locally, but i just thought of a way to do it and its working so brb gonna go implement that change

libs/components/indicators/testing/src/icon/icon-harness.ts
Blackbaud-SteveBrush: Better yet, this could be simplified to:
```
return classes ? 'skyux' : 'fa';
```

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,80 @@
...
+    const icon = await this.#getIcon();
+    return icon.hasClass(`fa-fw`);
+  }
+
+  /** Gets the icon type */
+  public async getIconType(): Promise<string> {
+    const iconClass = await (await this.#getIcon()).getAttribute('class');
+    const classes = iconClass
+      ?.split(' ')
+      .find((classValue: string) => /^sky-i-/.test(classValue));
```
Blackbaud-SteveBrush: The `find` method returns the found item, so yes, it does return a "truthy" value. For your use case, however, it's better to use `Array.some`. (I'm seeing this in a few places.)

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -26,15 +26,15 @@ export class SkyIconHarness extends SkyComponentHarness {
...
-    const regex = /^sky-i-|fa-(?!fw|2xs|xs|sm|lg|2xl|[0-9]+x)/;
-    let className = iconClass
-      ?.split(' ')
-      .find((classValue: string) => classValue.match(regex));
-
-    className = className?.replace(/-line|-solid$/, '');
-    className = className?.replace(/^sky-i-|fa-/, '');
-    return className;
+    const iconClass = await (await this.#getIcon()).getProperty('classList');
+    for (let i = 0; i < iconClass.length; i++) {
```
Blackbaud-PaulCrowder: One last thing to try: use `for...of` instead of a `for` loop here. It'll keep you from having to access the icon class by index multiple times inside the `for` loop.

```
const iconClasses = await (await this.#getIcon()).getProperty('classList');
for (const iconClass of iconClasses) {
  ...
}
```

Blackbaud-SandhyaRajasabeson: well damn, when i had iconClass hard cased to DOMTokenLlist it didnt like me using `for...of` loops bc it said DOMTokenList doesn't have an iterator, it works now tho without the type being set for `iconClass`

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  /** Gets if the icon has fixed width */
+  public async isFixedWidth(): Promise<boolean> {
+    const icon = await this.#getIcon();
+    return icon.hasClass(`fa-fw`);
+  }
+
+  /** Gets the icon type */
+  public async getIconType(): Promise<string> {
+    const iconClasses = await (await this.#getIcon()).getProperty('classList');
```
Blackbaud-PaulCrowder: OK, one last thing :) This line of code is gross enough that I think it justifies being put into its own function so it only has to be written once.

```
async #getIconClasses(): Promise<string[]> {
  const iconClasses = await (await this.#getIcon()).getProperty('classList');
  return Array.from(iconClasses);
}
```

libs/components/indicators/testing/src/icon/icon-harness.ts
Blackbaud-SteveBrush: Also, this regular expression might be a strong case to create a `data-` attribute, or a `sky-icon-foobar` class so that we don't need to exclude unknown FontAwesome utility classes to find the name. Let's talk offline as a team.

libs/components/indicators/testing/src/icon/icon-harness.spec.ts
Blackbaud-SandhyaRajasabeson: yup! i was also trying to future proof in case we add more `iconTypes` or `variants`. 
same for future proofing of the variable name, so if for some reason the `house` icon name goes away and this unit test fails, we only change it in one location. But i can change it for readability.

libs/components/indicators/testing/src/icon/icon-harness.ts
Blackbaud-SandhyaRajasabeson: im gonna bring this up in standup, but when i explicitely type the array/DomTokenList, the for loops all stop working bc it says there isnt an iterator, but if i don't type it it works at runtime???? it's very bizarre and frustrating

libs/components/indicators/testing/src/icon/icon-harness.ts
Blackbaud-SandhyaRajasabeson: would it be better to use the latter in this case just to have a longer explanation of what the regex is doing without someone in the future having to pull the regex into a tester to figure out what's going on

Blackbaud-SandhyaRajasabeson: from offline team talk: 
A new attribute is unneccessary for the following reasons - 
- we currently working on an older version of font awesome and have no plans to update to a newer version. So there is no chance of new classes being added that could break this regex. 
- We will also as a team not be adding any new classes that would follow the pattern of `fa-`

libs/components/indicators/src/lib/modules/icon/icon-stack.component.ts
```diff
@@ -13,7 +13,7 @@ import { SkyIconStackItem } from './icon-stack-item';
 export class SkyIconStackComponent {
   /**
    * The size of the icon using
-   * [Font Awesome sizes](https://fontawesome.com/how-to-use/on-the-web/styling/sizing-icons).
```
Blackbaud-SandhyaRajasabeson, line 16: links to the version of font awesome we use (v4.7) instead of the latest

libs/components/indicators/src/lib/modules/icon/icon.component.ts
```diff
@@ -30,13 +30,13 @@ export class SkyIconComponent {
...
    * The size of the icon using
-   * [Font Awesome sizes](https://fontawesome.com/how-to-use/on-the-web/styling/sizing-icons). Do not prefix the size with `fa-`.
+   * [Font Awesome sizes](https://fontawesome.com/v4/examples/). Do not prefix the size with `fa-`.
    */
   @Input()
   public size: string | undefined;
 
   /**
-   * Whether the icon has a fixed width.
+   * Ensures all icons of the same size take up the same width. Useful for stacking icons vertically.
```
Blackbaud-SandhyaRajasabeson: our current doc is not descriptive enough as to what fixed width actually does, todo: get confirmation from John Lyons on wording

apps/code-examples/src/app/code-examples/indicators/icon/basic/icon-demo.component.html
```diff
@@ -0,0 +1,26 @@
...
+<sky-icon
+  data-sky-id="icon-demo"
+  icon="calendar"
+  iconType="skyux"
+  variant="solid"
+  size="4x"
+  [fixedWidth]="true"
+></sky-icon>
+
+<div>
```
Blackbaud-SandhyaRajasabeson, line 14: added the below icons to show what fixedWidth does

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,108 @@
...
+   * Gets a `HarnessPredicate` that can be used to search for a
+   * `SkyIconHarness` that meets certain criteria.
+   */
+  public static with(
+    filters: SkyIconHarnessFilters
+  ): HarnessPredicate<SkyIconHarness> {
+    return SkyIconHarness.getDataSkyIdPredicate(filters);
+  }
+
+  async getIcon(): Promise<TestElement> {
```
Blackbaud-SteveBrush: Is this method public, private, private-private?

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Harness for interacting with an icon component in tests.
+ */
+export class SkyIconHarness extends SkyComponentHarness {
+  /**
+   * @internal
+   */
+  public static hostSelector = 'sky-icon';
+
+  //convert this into an actual function
```
Blackbaud-SteveBrush: Rogue comment.

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,108 @@
...
+  async getIcon(): Promise<TestElement> {
+    const icon = await this.locatorForOptional('.sky-icon')();
+    if (icon) {
+      return icon;
+    } else {
+      throw new Error('Icon does not exist');
+    }
+  }
+
+  async getIconClasses(): Promise<string[]> {
```
Blackbaud-SteveBrush: Is this method public, private, private-private?

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,108 @@
...
+  ): HarnessPredicate<SkyIconHarness> {
+    return SkyIconHarness.getDataSkyIdPredicate(filters);
+  }
+
+  async getIcon(): Promise<TestElement> {
+    const icon = await this.locatorForOptional('.sky-icon')();
+    if (icon) {
+      return icon;
+    } else {
+      throw new Error('Icon does not exist');
```
Blackbaud-SteveBrush: Is there a way you can make this error more meaningful if the consumer definitely puts `<sky-icon>` on the page but forgets to set the icon name?

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+
+  async getIconClasses(): Promise<string[]> {
+    const iconClasses = await (await this.getIcon()).getProperty('classList');
+    return Array.from(iconClasses);
+  }
+
+  /** Gets the icon name */
```
Blackbaud-SteveBrush: For consistency, let's make these comment blocks multi-line. Also, JSDocs comments should end in a period. Take a look at the alert harness for more JSDocs examples. https://github.com/blackbaud/skyux/blob/main/libs/components/indicators/testing/src/alert/alert-harness.ts
```
/**
 * Gets the icon name.
 */
```

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,108 @@
...
+  public async getIconName(): Promise<string | undefined> {
+    const iconClasses = await this.getIconClasses();
+    for (const iconClass of iconClasses) {
+      // match a class name that starts with `sky-i` or starts with `fa-` but does not follow with `fw` (fixed width) or lg, 2x, 3x, 4x, 5x
```
Blackbaud-SteveBrush: Good comment.

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,108 @@
...
+      if (/^sky-i-|^fa-(?!fw|lg|[2-5]+x)/.test(iconClass)) {
```
Blackbaud-SteveBrush, line 51: I'd match this Regular Expression's implementation with what you did below (e.g. `^(sky-i-|fa-)`), for code consistency's sake:
```
/^(sky-i-|fa-)(?!fw|lg|[2-5]+x)/
```

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,108 @@
...
+        return (
+          iconClass
+            .replace(ICON_CLASS_VARIANT_REGEXP, '')
+            // remove `sky-i` or `fa-` from the beginning of the icon class name
+            .replace(/^(sky-i-|fa-)/, '')
+        );
+      }
+    }
+    // this return will never be reached. If icon does not exist, the error in getIcon() will be thrown
```
Blackbaud-SteveBrush: Good comment, but do you mean `getIconClasses` instead of `getIcon`?

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,108 @@
...
+    /*istanbul ignore next*/
+    return undefined;
+  }
+
+  /** Gets the icon size */
+  public async getIconSize(): Promise<string | undefined> {
+    const iconClasses = await this.getIconClasses();
+    for (const iconClass of iconClasses) {
+      // match a class name that starts with `fa-` and  follows with lg, 2x, 3x, 4x, 5x
```
Blackbaud-SteveBrush: Good comment! Looks like you have a double space between `and  follows`.

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,108 @@
...
+  /** Gets if the icon is a variant */
+  public async getVariant(): Promise<string | undefined> {
+    if ((await this.getIconType()) === 'skyux') {
+      const iconClasses = await this.getIconClasses();
+      for (const iconClass of iconClasses) {
+        if (ICON_CLASS_VARIANT_REGEXP.test(iconClass)) {
+          return iconClass.substring(iconClass.lastIndexOf('-') + 1);
+        }
+      }
+      return undefined; // todo default as line
```
Blackbaud-SteveBrush: Todo?

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+    throw new Error(
+      'Variant cannot be determined because iconType is not skyux'
```
Blackbaud-SteveBrush: I'd be a tad more verbose with this error: 
```
A variant could not be determined because variants are only assigned to icons with a type of 'skyux'.
```

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,108 @@
...
+    );
+  }
+
+  /** Gets if the icon has fixed width */
```
Blackbaud-SteveBrush: JSDocs for methods that return boolean values should start with `Whether`.
```
Whether the icon is fixed width.
```

apps/code-examples/src/app/code-examples/indicators/icon/basic/icon-demo.component.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+describe('Basic icon', async () => {
+  async function setupTest(): Promise<SkyIconHarness> {
+    const fixture = TestBed.createComponent(IconDemoComponent);
+    const loader = TestbedHarnessEnvironment.loader(fixture);
+    const iconHarness = await loader.getHarness(
+      SkyIconHarness.with({
+        dataSkyId: 'icon-demo',
+      })
+    );
+    fixture.detectChanges();
```
Blackbaud-SteveBrush: Detecting changes should be governed by each `it` block (sometimes it's needed to run `expect` before the first change detection cycle). I would remove this line, and return the `fixture` from the `setupTest` function.
```
return { fixture, iconHarness };

// ...

it('should ...', async () => {
  const { fixture, iconHarness } = await setupTest();

  fixture.detectChanges();
});
```

apps/code-examples/src/app/code-examples/indicators/icon/icon-button/icon-button-demo.component.ts
```diff
@@ -0,0 +1,14 @@
...
+
+@Component({
+  selector: 'app-icon-demo',
+  templateUrl: './icon-button-demo.component.html',
+})
+export class IconDemoComponent {
+  public TextButtonClick(): void {
+    alert('Text with icon button clicked');
+  }
+  public IconOnlyButtonClick(): void {
```
Blackbaud-SteveBrush: ☝️ These methods should start with a lowercase letter.

apps/code-examples/src/app/code-examples/indicators/icon/icon-button/icon-button-demo.component.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+    dataSkyId?: string;
+  }): Promise<SkyIconHarness> {
+    const fixture = TestBed.createComponent(IconDemoComponent);
+    const loader = TestbedHarnessEnvironment.loader(fixture);
+    const iconHarness = await loader.getHarness(
+      SkyIconHarness.with({
+        dataSkyId: options?.dataSkyId,
+      })
+    );
+    fixture.detectChanges();
```
Blackbaud-SteveBrush: Same here (RE: return the `fixture` from `setupTest`).

apps/code-examples/src/app/code-examples/indicators/icon/basic/icon-demo.component.html
Blackbaud-SteveBrush: Nice demo.

libs/components/indicators/src/lib/modules/icon/icon.component.ts
Blackbaud-SteveBrush: Good idea. Thoughts @blackbaud-johnly?

libs/components/indicators/testing/src/icon/icon-harness.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+    const fixture = TestBed.createComponent(TestComponent);
+    const loader = TestbedHarnessEnvironment.loader(fixture);
+    const pageLoader = TestbedHarnessEnvironment.documentRootLoader(fixture);
+
+    const iconHarness: SkyIconHarness = options.dataSkyId
+      ? await loader.getHarness(
+          SkyIconHarness.with({ dataSkyId: options.dataSkyId })
+        )
+      : await loader.getHarness(SkyIconHarness);
+    fixture.detectChanges();
```
Blackbaud-SteveBrush: Triggering change detection should happen in the `it` blocks.

libs/components/indicators/testing/src/icon/icon-harness.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+
+    return { iconHarness, fixture, loader, pageLoader };
+  }
+
+  it('should return the correct icon name for all icon types and variants', async () => {
```
Blackbaud-SteveBrush, line 124: Good test description.

libs/components/indicators/testing/src/icon/icon-harness.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+        );
+      }
+    }
+  });
+
+  it('should throw error if icon name is not set', async () => {
+    const { iconHarness, fixture } = await setupTest();
+    fixture.componentInstance.iconName = undefined;
+    await expectAsync(iconHarness.getIconName()).toBeRejectedWithError(
+      'Icon does not exist'
```
Blackbaud-SteveBrush: The error for this specific scenario should be more helpful. The consumer has added the `<sky-icon>` to their template but has forgotten a `name` (or it's set to `undefined`). We need to find a way to make this error more helpful so the consumer knows what they can do to rectify it.

libs/components/indicators/testing/src/icon/icon-harness.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+  });
+
+  it('should return the default icon type', async () => {
+    const { iconHarness, fixture } = await setupTest();
+    fixture.componentInstance.iconType = undefined;
+    fixture.detectChanges();
+    await expectAsync(iconHarness.getIconType()).toBeResolvedTo('fa');
+  });
+
+  it('should return the right variant for skyux icons', async () => {
```
Blackbaud-SteveBrush: Do you mean "correct value"? I would avoid the word "right" since that word is used often in JavaScript for offset values.

libs/components/indicators/testing/src/icon/icon-harness.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+  });
+
+  it('should return default values for fixed width', async () => {
+    const { iconHarness, fixture } = await setupTest();
+    fixture.componentInstance.fixedWidth = undefined;
+    fixture.detectChanges();
+    await expectAsync(iconHarness.isFixedWidth()).toBeResolvedTo(false);
+  });
+
+  it('should return the right value for fixed width', async () => {
```
Blackbaud-SteveBrush: Ditto, "right".

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,108 @@
+import { HarnessPredicate, TestElement } from '@angular/cdk/testing';
+import { SkyComponentHarness } from '@skyux/core/testing';
+
+import { SkyIconHarnessFilters } from './icon-harness-filters';
+
+// match ending in `-line` or `-solid`
```
Blackbaud-SteveBrush: To be more clear: `matches a class name ending...`

Blackbaud-SteveBrush (commented):
There's a `NX` file in this commit that should be removed. 

Overall, looks good!

libs/components/indicators/src/lib/modules/icon/icon.component.ts
blackbaud-johnly: How about:

"Whether to enforce a fixed width based on icon size. By default, icons of a specified size share a consistent height, but their widths vary and can throw off vertical alignment. Use a fixed width when you stack icons vertically, such as in lists."

libs/components/indicators/testing/src/icon/icon-harness.ts
Blackbaud-SandhyaRajasabeson: `sky-i-fw` would never be a thing tho, `fa-` would be the only case where it is followed by things that we need to test for

libs/components/indicators/testing/src/icon/icon-harness.ts
Blackbaud-SandhyaRajasabeson: nah the error is in `#getIcon`. `#getIconClasses` calls `#getIcon`

libs/components/indicators/testing/src/icon/icon-harness.ts
Blackbaud-SandhyaRajasabeson: to do lmao, glad i put that there, i forgot to address that

libs/components/indicators/testing/src/icon/icon-harness.ts
Blackbaud-SteveBrush: `(sky-i-|fa-)` means `sky-i` or `fa-` (it's the same format you're using below.

libs/components/indicators/testing/src/icon/icon-harness.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+
+  async #getIconClasses(): Promise<string[]> {
+    const iconClasses = await (await this.#getIcon()).getProperty('classList');
+    return Array.from(iconClasses);
+  }
+
+  /**
+   * Gets the icon name
```
Blackbaud-SteveBrush: JSDocs comments need to end in a period. Same for the ones below.

Blackbaud-SteveBrush (dismissed):

Blackbaud-PaulCrowder (approved):

PR #1374, opened 2023-05-16
feat(components/http): add callerid query string to bbid requests

Blackbaud-TimMcVicker:
This adds a query string to bbid requests to the backend that represents the service (svcid) and SPA. We'll propagate this request on telemetry baggage across backend requests, which will eventually end up on usage-based billing events, so we can say "this amount of cost initiated from svcid=renxt". Backend .NET implementation is here: https://dev.azure.com/blackbaud/Products/_git/core-webservice-aspnetcore/pullrequest/311779


Blackbaud-TimMcVicker:
I've added some unit tests here but am still trying to figure out how to manually test it. The playground SPA in this repo doesn't appear to be setup for auth'd requests

PR #1421, opened 2023-06-01
feat(components/modals): add `isDirty` directive to modals

Blackbaud-ScottFreeman:
Adding new SkyModalIsDirtyDirective so developers can flag modals as dirty and prompt users to discard changes when they have unsaved work.


libs/components/modals/testing/src/modal/modal-harness.ts
```diff
@@ -83,4 +83,15 @@ export class SkyModalHarness extends SkyComponentHarness {
     const modal = this.#getModal();
     return (await modal).hasClass('sky-modal-full-page');
   }
+
+  /**
+   * Whether the modal has {@link SkyModalIsDirtyDirective.isDirty} set to dirty.
+   */
+  public async isDirty(): Promise<boolean> {
```
Blackbaud-PaulCrowder, line 90: I'm not sure we need this as it's not something that the user can perceive.

libs/components/modals/testing/src/modal/modal-harness.ts
```diff
@@ -83,4 +83,15 @@ export class SkyModalHarness extends SkyComponentHarness {
...
+    const modalHost = await this.host();
+    const isDirtyAttribute = await modalHost.getAttribute(
```
Blackbaud-PaulCrowder: If we do decide we need this, we should pick something other than `ng-reflect-is-dirty` to check as that is an implementation detail of Angular that is not part of their public API. We could add our own attribute to check instead.

libs/components/modals/src/lib/modules/modal/modal-is-dirty.directive.ts
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * Flags whether or not the modal is considered dirty.
+   * @required
+   */
+  @Input()
+  public isDirty = false;
+
+  readonly #ngUnsubscribe = new Subject<void>();
+
+  readonly #modalInstance: SkyModalInstance;
```
Blackbaud-PaulCrowder: You can use the `inject()` function here and eliminate the constructor.

```
readonly #modalInstance = inject(SkyModalInstance);
...
```

libs/components/modals/src/lib/modules/modal/modal-is-dirty.directive.ts
```diff
@@ -0,0 +1,99 @@
...
+  readonly #confirmService: SkyConfirmService;
```
Blackbaud-PaulCrowder: Our convention is to use `svc` as the suffix for service variable names.

```
#confirmSvc = inject(SkyConfirmService);
```

libs/components/modals/src/lib/modules/modal/modal.component.spec.ts
```diff
@@ -1133,4 +1135,118 @@ describe('Modal component', () => {
...
+
+      const keepWorkingButtonEl = getKeepWorkingButtonElement(confirmModalEl);
+      await expectAsync(keepWorkingButtonEl).toHaveLibResourceText(
+        'skyux_modal_dirty_default_keep_working_text'
+      );
+    }
+
+    let dirtyContextProvider: { providers: StaticProvider[] };
+
+    beforeEach(fakeAsync(() => {
```
Blackbaud-PaulCrowder: You don't need `fakeAsync()` here. Also verify that it's needed in other unit tests as using it unnecessarily makes it harder to refactor/modify tests.

libs/components/modals/src/lib/modules/modal/modal-is-dirty.directive.ts
Blackbaud-ScottFreeman: Sure, I can make these changes. I wasn't sure if that's the way you all were leaning because there benefits to injecting all dependencies into constructors. Java devs used `@Autowire` for a long time until they determined it was bad practice because it didn't allow variables to be marked as final (not an issue here), and it also made things harder to unit test because you had to fully buy into a Dependency Injection system since that was the only way to easily set those private fields. Now, it is best practice to inject all dependencies into the constructor in Java.

libs/components/modals/src/lib/modules/modal/modal.component.spec.ts
Blackbaud-ScottFreeman: Good catch. That was left behind from some refactoring I was doing. As for the other tests, you might be right in it not being needed. I thought it was necessary because of the `closeModal` function, but that only calls the function `tick()` if you pass in `true` for `async`. Otherwise, it ticks the ApplicationRef.

libs/components/modals/src/lib/modules/modal/modal-is-dirty.directive.ts
Blackbaud-PaulCrowder: I'm not worried about that, especially for components, because we would never construct them outside Angular's `TestBed`. For services you can provide the service in `TestBed` along with mocks (where needed) for injected tokens.

libs/components/modals/testing/src/modal/modal-harness.ts
Blackbaud-ScottFreeman: One of the goals with these types of features is to keep developers from different teams from rewriting the same code. My hope is for that to apply to tests as well. I added this so that a developer could easily unit test that they were setting `isDirty` and that it contained the expected value. Without this, I'm not sure how developers would be able to test this without seeing if a confirmation modal opened.

libs/components/modals/testing/src/modal/modal-harness.ts
Blackbaud-ScottFreeman: I meant to ask about this in hopes that maybe you all would have some ideas. I agree that this is definitely not ideal since it relies on an Angular implementation detail, but I wasn't sure how else to test the value of the directive's input. My hope was that if Angular changed things, we'd only have to update this one function for everyone.

It sounds like maybe you'd prefer if the directive added a new attribute to the host element that contained the `@Input`'s value. I think I can do that and then test for it here.

PR #1484, opened 2023-06-29
feat(components/modals): add errors input to modal footer component

Blackbaud-CoreyArcher:
[AB#2460030](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2460030)


libs/components/modals/src/lib/modules/modal/modal-footer.component.scss
Blackbaud-CoreyArcher, line 1: I had the thought that we would apply this style to all status indicators that consumers might already be using. Trevor suggested we scope it to our own implementation. Unsure what the right answer is here

libs/components/modals/src/lib/modules/modal/modal-footer.component.ts
Blackbaud-CoreyArcher, line 1: Feedback on docs comment appreciated

blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/1484/storybook/)

Component Storybooks:

- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/ag-grid-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/avatar-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/data-manager-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/errors-storybook/)
- [flyout-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/flyout-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/forms-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/layout-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/lookup-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/modals-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/phone-field-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/1484/storybooks/text-editor-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/1484/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/1484/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/1484/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/1484/playground/)

libs/components/modals/src/lib/modules/modal/modal-footer.component.scss
```diff
@@ -16,6 +16,11 @@
   .sky-btn + .sky-btn-link {
     margin-left: -2px;
   }
+
+  /* Should this be all status indicators or just ours? */
+  .sky-modal-footer-errors .sky-status-indicator {
+    margin-bottom: var(--sky-margin-stacked-lg);
```
Blackbaud-SteveBrush: Just confirming: Adam recommended this value for the margin?

libs/components/modals/src/lib/modules/modal/modal-footer.component.scss
Blackbaud-SteveBrush: I would scope it to our own implementation since we want to avoid altering existing consumer implementations. We can inform consumers of the correct thing to do in our weekly newsletter.

apps/playground/src/app/components/modal/modal-error.component.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.#modalInstance.save();
+  }
+
+  public cancel(): void {
+    this.#modalInstance.cancel();
+  }
+
+  #onClose(handler: SkyModalBeforeCloseHandler): void {
+    if (this.isError && handler.closeArgs.reason !== 'cancel') {
+      this.errors = ['Sample error', 'Sample error 2'];
```
Blackbaud-SteveBrush: Can we put a multiline error in here so we can see what it looks like in the playground?

Blackbaud-SteveBrush: I would also create a new modal visual test that includes 2-3 error messages (one modal), with one of the errors being multiline so we can catch any spacing changes in the future.

libs/components/modals/src/lib/modules/modal/modal-footer.component.ts
```diff
@@ -9,4 +9,11 @@ import { Component, ViewEncapsulation } from '@angular/core';
   styleUrls: ['./modal-footer.component.scss'],
   encapsulation: ViewEncapsulation.None,
 })
-export class SkyModalFooterComponent {}
+export class SkyModalFooterComponent {
+  /**
+   * A collection of error messages to be displayed to the user in the event of a form error.
+   * Values assigned here will be displayed following [SKY UX style guidelines](https://developer.blackbaud.com/skyux/design/guidelines/form-design#validation-and-error-handling).
```
Blackbaud-SteveBrush: I don't think you need the second sentence since it should be implied that if it's a first-class implementation it will follow SKY UX guidelines. 

The first sentence can be trimmed down a bit (describing "what" instead of "why"): "A list of error messages to be shown above the modal footer buttons."

Blackbaud-SteveBrush (commented):
I added enough errors to make the modal's height equal to the height of the window and I was unable to "escape" or take action on the modal. We need to account for this rare scenario, especially since it has more likelihood to occur on mobile devices.
Screenshot 2023-07-07 at 9 59 30 AM

libs/components/modals/src/lib/modules/modal/modal-footer.component.scss
Blackbaud-CoreyArcher: Yes, Adam and Paul suggested we use this variable

libs/components/modals/src/lib/modules/modal/modal-footer.component.ts
Blackbaud-CoreyArcher: updated

apps/playground/src/app/components/modal/modal-error.component.ts
Blackbaud-CoreyArcher: added

Blackbaud-CoreyArcher:
> I added enough errors to make the modal's height equal to the height of the window and I was unable to "escape" or take action on the modal. We need to account for this rare scenario, especially since it has more likelihood to occur on mobile devices. Screenshot 2023-07-07 at 9 59 30 AM

After discussing with the team, added some additional styling to cap these messages so they don't interefere with the overall modal layout.

libs/components/modals/src/lib/modules/modal/modal-footer.component.ts
```diff
@@ -9,4 +12,14 @@ import { Component, ViewEncapsulation } from '@angular/core';
...
   encapsulation: ViewEncapsulation.None,
 })
-export class SkyModalFooterComponent {}
+export class SkyModalFooterComponent {
+  public formErrors: SkyModalError[] | undefined;
+
+  #adapterSvc = inject(SkyModalComponentAdapterService);
+
+  constructor() {
+    this.#adapterSvc.formErrors.subscribe((errors) => {
```
Blackbaud-PaulCrowder: I'd prefer to see the service's `formErrors` property referenced directly in the template using the `async` pipe. That way it will be unsubscribed when the component is destroyed, avoids the "magic" of Angular change detection picking up the change to `formErrors` outside of an input changing, and eliminates a decent amount of code from the class.

Template:

```
  <div aria-live="polite">
    <div *ngIf="(errorsSvc.formErrors | async) as formErrors" class="sky-modal-footer-errors">
      <sky-status-indicator
        *ngFor="let formError of formErrors"
        descriptionType="error"
        indicatorType="danger"
      >
        {{ formError.message }}
      </sky-status-indicator>
    </div>
  </div>
```
Class:

```
  protected errorsSvc = inject(SkyModalErrorsService);
```

libs/components/modals/src/lib/modules/modal/modal-footer.component.ts
```diff
@@ -9,4 +12,14 @@ import { Component, ViewEncapsulation } from '@angular/core';
   styleUrls: ['./modal-footer.component.scss'],
   encapsulation: ViewEncapsulation.None,
 })
-export class SkyModalFooterComponent {}
+export class SkyModalFooterComponent {
+  public formErrors: SkyModalError[] | undefined;
+
+  #adapterSvc = inject(SkyModalComponentAdapterService);
```
Blackbaud-PaulCrowder: The purpose of adapter classes is to isolate DOM manipulation from the rest of the code. Adding a non-DOM-manipulating method or property to it means if we ever want to swap out all the DOM manipulation logic with another service, we'd have to duplicate the non-DOM-manipulation logic. I suggest creating a new `SkyModalErrorsService` with the new property and providing it in `SkyModalComponent`.

libs/components/modals/src/lib/modules/modal/modal-footer.component.scss
```diff
@@ -16,6 +16,16 @@
   .sky-btn + .sky-btn-link {
     margin-left: -2px;
   }
+
+  .sky-modal-footer-errors {
+    max-height: 100px;
+    overflow-y: auto;
+    margin-bottom: var(--sky-margin-stacked-lg);
```
Blackbaud-PaulCrowder: Instead of adding the margin here, can you remove the rule below and let the last status indicator serve as the bottom margin? Then you can use the `sky-margin-stacked-lg` CSS class on `sky-status-indicator`.

```
.sky-modal-footer-errors {
  max-height: 100px;
  overflow-y: auto;
}
```

In the template: 

```
<sky-status-indicator class="sky-margin-stacked-lg">
```

libs/components/modals/src/lib/modules/modal/modal-footer.component.ts
Blackbaud-CoreyArcher: updated

libs/components/modals/src/lib/modules/modal/modal-footer.component.ts
Blackbaud-CoreyArcher: added

libs/components/modals/src/lib/modules/modal/modal-footer.component.scss
Blackbaud-CoreyArcher: updated

Blackbaud-PaulCrowder (approved):

PR #1494, opened 2023-07-06
feat(sdk/testing): support axe-core 4.7

johnhwhite:
[AB#2600210](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2600210)

### DIFF
--- a/libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.html
+++ b/libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.html
@@ -4,15 +4,7 @@
   [formGroup]="editorForm"
 >
   <sky-lookup
-    [ariaLabel]="
-      skyComponentProperties?.ariaLabelledBy
-        ? undefined
-        : skyComponentProperties?.ariaLabel ||
-          colDef?.headerName ||
-          colDef?.headerTooltip ||
-          colDef?.field ||
-          colDef?.colId
-    "
+    [ariaLabel]="ariaLabel"
     [ariaLabelledBy]="skyComponentProperties?.ariaLabelledBy"
     [autocompleteAttribute]="skyComponentProperties?.autocompleteAttribute"
     [data]="skyComponentProperties?.data"
@@ -46,15 +38,7 @@
   [formGroup]="editorForm"
 >
   <sky-lookup
-    [ariaLabel]="
-      skyComponentProperties?.ariaLabelledBy
-        ? undefined
-        : skyComponentProperties?.ariaLabel ||
-          colDef?.headerName ||
-          colDef?.headerTooltip ||
-          colDef?.field ||
-          colDef?.colId
-    "
+    [ariaLabel]="ariaLabel"
     [ariaLabelledBy]="skyComponentProperties?.ariaLabelledBy"
     [autocompleteAttribute]="skyComponentProperties?.autocompleteAttribute"
     [data]="skyComponentProperties?.data"
--- a/libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.html
+++ b/libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.html
@@ -4,7 +4,15 @@
   [formGroup]="editorForm"
 >
   <sky-lookup
-    [ariaLabel]="skyComponentProperties?.ariaLabel"
+    [ariaLabel]="
+      skyComponentProperties?.ariaLabelledBy
+        ? undefined
+        : skyComponentProperties?.ariaLabel ||
+          colDef?.headerName ||
+          colDef?.headerTooltip ||
+          colDef?.field ||
+          colDef?.colId
+    "
     [ariaLabelledBy]="skyComponentProperties?.ariaLabelledBy"
     [autocompleteAttribute]="skyComponentProperties?.autocompleteAttribute"
     [data]="skyComponentProperties?.data"
@@ -23,12 +31,12 @@
     [searchTextMinimumCharacters]="
       skyComponentProperties?.searchTextMinimumCharacters
     "
-    (searchAsync)="skyComponentProperties?.searchAsync($event)"
+    (searchAsync)="skyComponentProperties?.searchAsync?.($event)"
     [selectMode]="skyComponentProperties?.selectMode"
     [showAddButton]="skyComponentProperties?.showAddButton"
     [showMoreConfig]="skyComponentProperties?.showMoreConfig"
     [wrapperClass]="skyComponentProperties?.wrapperClass"
-    (addClick)="skyComponentProperties?.addClick($event)"
+    (addClick)="skyComponentProperties?.addClick?.($event)"
   >
   </sky-lookup>
 </sky-input-box>
@@ -38,7 +46,15 @@
   [formGroup]="editorForm"
 >
   <sky-lookup
-    [ariaLabel]="skyComponentProperties?.ariaLabel"
+    [ariaLabel]="
+      skyComponentProperties?.ariaLabelledBy
+        ? undefined
+        : skyComponentProperties?.ariaLabel ||
+          colDef?.headerName ||
+          colDef?.headerTooltip ||
+          colDef?.field ||
+          colDef?.colId
+    "
     [ariaLabelledBy]="skyComponentProperties?.ariaLabelledBy"
     [autocompleteAttribute]="skyComponentProperties?.autocompleteAttribute"
     [data]="skyComponentProperties?.data"
@@ -61,7 +77,7 @@
     [showAddButton]="skyComponentProperties?.showAddButton"
     [showMoreConfig]="skyComponentProperties?.showMoreConfig"
     [wrapperClass]="skyComponentProperties?.wrapperClass"
-    (addClick)="skyComponentProperties?.addClick($event)"
+    (addClick)="skyComponentProperties?.addClick?.($event)"
   >
   </sky-lookup>
 </sky-input-box>
--- a/libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.ts
+++ b/libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.ts
@@ -8,7 +8,7 @@ import {
 import { UntypedFormControl, UntypedFormGroup } from '@angular/forms';
 
 import { ICellEditorAngularComp } from 'ag-grid-angular';
-import { ColDef, ColumnResizedEvent } from 'ag-grid-community';
+import { ColumnResizedEvent } from 'ag-grid-community';
 import { IPopupComponent } from 'ag-grid-community/dist/lib/interfaces/iPopupComponent';
 
 import { applySkyLookupPropertiesDefaults } from '../../apply-lookup-properties-defaults';
@@ -42,9 +42,7 @@ export class SkyAgGridCellEditorLookupComponent
   });
   public useAsyncSearch = false;
 
-  protected get colDef(): ColDef | undefined {
-    return this.#params?.colDef;
-  }
+  protected ariaLabel: string | undefined = undefined;
 
   #params: SkyCellEditorLookupParams | undefined;
   #triggerType: SkyAgGridCellEditorInitialAction | undefined;
@@ -64,6 +62,13 @@ export class SkyAgGridCellEditorLookupComponent
 
     this.#triggerType = SkyAgGridCellEditorUtils.getEditorInitialAction(params);
     const control = this.editorForm.get('selection');
+    this.ariaLabel = params.skyComponentProperties?.ariaLabelledBy
+      ? undefined
+      : params.skyComponentProperties?.ariaLabel ||
+        params.colDef.headerName ||
+        params.colDef.headerTooltip ||
+        params.colDef.field ||
+        params.colDef.colId;
 
     if (control) {
       switch (this.#triggerType) {
--- a/libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.ts
+++ b/libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.ts
@@ -8,7 +8,7 @@ import {
 import { UntypedFormControl, UntypedFormGroup } from '@angular/forms';
 
 import { ICellEditorAngularComp } from 'ag-grid-angular';
-import { ColumnResizedEvent } from 'ag-grid-community';
+import { ColDef, ColumnResizedEvent } from 'ag-grid-community';
 import { IPopupComponent } from 'ag-grid-community/dist/lib/interfaces/iPopupComponent';
 
 import { applySkyLookupPropertiesDefaults } from '../../apply-lookup-properties-defaults';
@@ -41,6 +41,10 @@ export class SkyAgGridCellEditorLookupComponent
     }),
   });
   public useAsyncSearch = false;
+
+  protected get colDef(): ColDef | undefined {
+    return this.#params?.colDef;
+  }
 
   #params: SkyCellEditorLookupParams | undefined;
   #triggerType: SkyAgGridCellEditorInitialAction | undefined;
--- a/libs/components/lookup/src/lib/modules/autocomplete/autocomplete-input.directive.ts
+++ b/libs/components/lookup/src/lib/modules/autocomplete/autocomplete-input.directive.ts
@@ -179,7 +179,7 @@ export class SkyAutocompleteInputDirective
     this.#textChangesObs = this.#textChanges.asObservable();
   }
 
-  public ngOnInit() {
+  public ngOnInit(): void {
     const element = this.#elementRef.nativeElement;
 
     this.#setAttributes(this.#elementRef);
@@ -283,17 +283,20 @@ export class SkyAutocompleteInputDirective
     }
   }
 
-  public setAriaOwns(overlayId: string | null): void {
+  /**
+   * Used to connect the input to the overlay.
+   */
+  public setAriaControls(overlayId: string | null): void {
     if (overlayId) {
       this.#renderer.setAttribute(
         this.#elementRef.nativeElement,
-        'aria-owns',
+        'aria-controls',
         overlayId
       );
     } else {
       this.#renderer.removeAttribute(
         this.#elementRef.nativeElement,
-        'aria-owns'
+        'aria-controls'
       );
     }
   }
--- a/libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.html
+++ b/libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.html
@@ -3,9 +3,9 @@
   aria-haspopup="listbox"
   class="sky-autocomplete"
   role="combobox"
-  [attr.aria-expanded]="isOpen"
+  [attr.aria-expanded]="isDropdownVisible | async"
   [attr.aria-labelledby]="ariaLabelledBy"
-  [attr.aria-controls]="isOpen ? resultsListId : null"
+  [attr.aria-controls]="(isDropdownVisible | async) ? resultsListId : null"
   [attr.id]="resultsWrapperId"
 >
   <ng-content></ng-content>
@@ -32,53 +32,47 @@
 
 <ng-template #resultControlsTemplateRef>
   <div
+    *ngIf="isDropdownVisible | async"
     class="sky-autocomplete-results"
-    [class.sky-screen-reader-only]="
-      !searchText || (showActionsArea && searchResults.length === 0)
-    "
     role="listbox"
     [attr.aria-labelledby]="ariaLabelledBy"
     [attr.id]="resultsListId"
   >
-    <ng-container
-      *ngIf="searchText && (!showActionsArea || searchResults.length > 0)"
-    >
-      <ng-container *ngFor="let result of searchResults; let i = index">
-        <div
-          *ngIf="
-            (!searchResultsLimit || i < searchResultsLimit) &&
-            !(enableShowMore && i >= 5)
-          "
-          class="sky-autocomplete-result"
-          role="option"
-          tabindex="0"
-          [attr.aria-selected]="isElementFocused(searchResultEl)"
-          [attr.data-descriptor-value]="result.data[descriptorProperty]"
-          [attr.id]="result.elementId"
-          [skyHighlight]="highlightText"
-          (mousedown)="onResultMouseDown(result.elementId, $event)"
-          (mousemove)="onResultMouseMove(i)"
-          #searchResultEl
-        >
-          <ng-container
-            *ngTemplateOutlet="
-              searchResultTemplate || defaultSearchResultTemplate;
-              context: { item: result.data }
-            "
-          >
-          </ng-container>
-        </div>
-      </ng-container>
+    <ng-container *ngFor="let result of searchResults; let i = index">
       <div
-        *ngIf="searchResults.length === 0"
-        class="sky-deemphasized sky-padding-squish-default"
+        *ngIf="
+          (!searchResultsLimit || i < searchResultsLimit) &&
+          !(enableShowMore && i >= 5)
+        "
+        class="sky-autocomplete-result"
+        role="option"
+        tabindex="0"
+        [attr.aria-selected]="isElementFocused(searchResultEl)"
+        [attr.data-descriptor-value]="result.data[descriptorProperty]"
+        [attr.id]="result.elementId"
+        [skyHighlight]="highlightText"
+        (mousedown)="onResultMouseDown(result.elementId, $event)"
+        (mousemove)="onResultMouseMove(i)"
+        #searchResultEl
       >
-        {{
-          noResultsFoundText ||
-            ('skyux_autocomplete_no_results' | skyLibResources)
-        }}
+        <ng-container
+          *ngTemplateOutlet="
+            searchResultTemplate || defaultSearchResultTemplate;
+            context: { item: result.data }
+          "
+        >
+        </ng-container>
       </div>
     </ng-container>
+    <div
+      *ngIf="searchResults.length === 0"
+      class="sky-deemphasized sky-padding-squish-default"
+    >
+      {{
+        noResultsFoundText ||
+          ('skyux_autocomplete_no_results' | skyLibResources)
+      }}
+    </div>
   </div>
   <div *ngIf="showActionsArea" class="sky-autocomplete-actions">
     <button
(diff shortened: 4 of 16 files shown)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/1494/storybook/)

Component Storybooks:

- [a11y-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/a11y-storybook/)
- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/colorpicker-storybook/)
- [core-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/core-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/errors-storybook/)
- [flyout-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/flyout-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/forms-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/lookup-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/modals-storybook/)
- [navbar-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/navbar-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/phone-field-storybook/)
- [popovers-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/popovers-storybook/)
- [progress-indicator-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/progress-indicator-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/split-view-storybook/)
- [storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/theme-storybook/)
- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/tiles-storybook/)
- [toast-storybook](https://blackbaud.github.io/skyux-pr-preview/1494/storybooks/toast-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/1494/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/1494/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/1494/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/1494/playground/)

libs/sdk/testing/package.json
```diff
@@ -20,7 +20,7 @@
     "@angular/core": "^15.2.9",
     "@angular/platform-browser": "^15.2.9",
     "@skyux/i18n": "0.0.0-PLACEHOLDER",
-    "axe-core": "^3.5.6 || ^4.6.3"
+    "axe-core": "^3.5.6 || ^4.6.3 || ^4.7.2"
```
Blackbaud-SteveBrush: I looked at the changelog and didn't see any breaking changes between `4.6` and `4.7`. At any rate, `^4.6.3` evaluates to the same SemVer range as `^4.7.2` so we can probably just remove `^4.6.3` and require folks to use the latest version of `4`.

libs/components/lookup/testing/src/autocomplete/autocomplete-input-harness.ts
```diff
@@ -63,9 +63,9 @@ export class SkyAutocompleteInputHarness extends ComponentHarness {
   }
 
   /**
-   * Returns the value of the input's `aria-owns` attribute.
+   * Returns the value of the input's `aria-controls` attribute.
    */
   public async getAriaOwns(): Promise<string | null> {
-    return (await this.host()).getAttribute('aria-owns');
+    return (await this.host()).getAttribute('aria-controls');
```
Blackbaud-SteveBrush, line 69: The name of this method conflicts with the type of value returned. Why did we have to use `aria-controls` instead of `aria-owns`?

libs/components/lookup/src/lib/modules/lookup/lookup.component.spec.ts
```diff
@@ -3776,7 +3782,9 @@ describe('Lookup component', function () {
 
         fixture.detectChanges();
         await fixture.whenStable();
-        await expectAsync(document.body).toBeAccessible(axeConfig);
+        await expectAsync(
+          fixture.nativeElement.ownerDocument.body
```
Blackbaud-SteveBrush: Can you provide some background on this change?

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -887,7 +887,7 @@ export class SkyAutocompleteComponent implements OnDestroy, AfterViewInit {
 
   #updateAriaOwns(): void {
     if (this.inputDirective) {
-      this.inputDirective.setAriaOwns(this.#overlay?.id || null);
+      this.inputDirective.setAriaControls(this.#overlay?.id || null);
```
Blackbaud-SteveBrush, line 925: The name of this method doesn't describe what it's doing. Is there a reason we switched to `aria-controls`?

libs/components/lookup/src/lib/modules/autocomplete/autocomplete-input.directive.ts
```diff
@@ -166,66 +159,44 @@ export class SkyAutocompleteInputDirective
...
 
   public ngOnDestroy(): void {
     this.#blur.complete();
     this.#textChanges.complete();
+  }
 
-    this.#ngUnsubscribe.next();
-    this.#ngUnsubscribe.complete();
+  @HostListener('blur')
+  public onBlurHandler(): void {
```
Blackbaud-SteveBrush: Is there an accessibility reason we switched from `observableFromEvent` to `HostListener`? If so, these can probably be `protected` instead of `public`.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete-input.directive.ts
```diff
@@ -108,7 +109,7 @@ export class SkyAutocompleteInputDirective
   }
 
   public get textChanges(): Observable<SkyAutocompleteInputTextChange> {
-    return this.#textChangesObs;
+    return this.#textChanges.asObservable();
```
Blackbaud-SteveBrush: These were set to `this.#textChangesObs;` in the constructor so that the value is created once, and not every time `get textChanges` is called.

libs/components/lookup/testing/src/autocomplete/autocomplete-input-harness.ts
johnhwhite: There was a [change in axe-core 4.6](https://github.com/dequelabs/axe-core/pull/3786) that made `aria-controls` required on `role="combobox"` elements and changes in 4.7 clarified rules around `combobox` and referenced elements.

Without switching from `aria-owns` to `aria-controls`, this is the error:

```
[Rule: 'aria-required-attr'] Required ARIA attributes must be provided - WCAG: wcag412
 Get help at: https://dequeuniversity.com/rules/axe/4.7/aria-required-attr?application=axeAPI

[CRITICAL] Fix any of the following:
  - Required ARIA attribute not present: aria-controls
Target: #sky-autocomplete-wrapper-91
HTML: <div _ngcontent-a-c80="" aria-autocomplete="list" aria-haspopup="listbox" role="combobox" class="sky-autocomplete" aria-expanded="true" aria-labelledby="my-autocomplete-label" id="sky-autocomplete-wrapper-91" aria-owns="sky-autocomplete-list-91">
  - [aria-required-attr] Required ARIA attribute not present: aria-controls
```

libs/components/lookup/src/lib/modules/lookup/lookup.component.spec.ts
johnhwhite: I made this change when I wasn't seeing the result I expected, it ended up being a different issue, and I've reverted this change.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
johnhwhite: Updated the method name. See above for the `aria-controls` explanation.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete-input.directive.ts
johnhwhite: I made these changes when I wasn't seeing the result I expected, it ended up being a different issue, and I've reverted these changes.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete-input.directive.ts
johnhwhite: Ahh ok, reverted.

libs/sdk/testing/package.json
johnhwhite: Nothing in the changelog suggests a breaking change, but the update does cause passing tests to potentially fail. I'll add it to the flex time agenda.

johnhwhite: We decided to leave the dependency at `4.6` but consumers using `4.7` should no longer see errors.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.html
```diff
@@ -5,7 +5,7 @@
   role="combobox"
   [attr.aria-expanded]="isOpen"
   [attr.aria-labelledby]="ariaLabelledBy"
-  [attr.aria-owns]="isOpen ? resultsListId : null"
+  [attr.aria-controls]="isOpen ? resultsListId : null"
```
Blackbaud-SteveBrush: Since you're only applying the `aria-controls` attribute if the results dropdown exists (i.e. `isOpen`), wouldn't that let you revert all the other changes in this file? (See my comment in your spec file as a followup.)

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.spec.ts
```diff
@@ -1094,27 +1094,57 @@ describe('Autocomplete component', () => {
...
+        },
+      };
+
+      fixture.detectChanges();
+
+      const inputElement: HTMLInputElement = getInputElement();
+
+      await fixture.whenStable();
+      fixture.detectChanges();
+      inputElement.focus();
```
Blackbaud-SteveBrush: This test doesn't fully show the search results until you put a value on the input (I set `teardown: { destroyAfterEach: false },` on the `TestBed` so I could view the component in the debug window). If you add `inputElement.value = 'r'` to this test the dropdown will show properly and `aria-controls` will be valid.

Because of this, you won't need the other changes in the autocomplete's template that were made to hide the dropdown.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
Blackbaud-SteveBrush: I still see `#updateAriaOwns`.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -969,22 +969,24 @@ export class SkyAutocompleteComponent implements OnDestroy, AfterViewInit {
...
-            /* istanbul ignore else */
-            if (this.#affixer) {
-              this.#affixer.reaffix();
-            }
-          });
-          break;
-      }
-    });
+    this.#messageStreamSub = this.messageStream
+      .pipe(takeUntil(this.#ngUnsubscribe))
```
Blackbaud-SteveBrush, line 1009: Good call.

libs/components/lookup/testing/src/autocomplete/autocomplete-input-harness.ts
Blackbaud-SteveBrush: Since this harness is internal, we can safely rename `getAriaOwns` to `getAriaControls`.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete-input.directive.ts
```diff
@@ -283,17 +283,28 @@ export class SkyAutocompleteInputDirective
     }
   }
 
+  /* istanbul ignore next */
+  /**
+   * @deprecated Use `setAriaControls` instead.
```
Blackbaud-SteveBrush: I would just remove this method altogether since it's not part of the public API.

libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.html
```diff
@@ -61,7 +66,7 @@
     [showAddButton]="skyComponentProperties?.showAddButton"
     [showMoreConfig]="skyComponentProperties?.showMoreConfig"
     [wrapperClass]="skyComponentProperties?.wrapperClass"
-    (addClick)="skyComponentProperties?.addClick($event)"
+    (addClick)="skyComponentProperties?.addClick?.($event)"
```
Blackbaud-SteveBrush, line 65: I'm confused by these final `?.(`. Are they needed?

libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.html
```diff
@@ -1,11 +1,17 @@
+<label class="sky-screen-reader-only" skyId #labelElement="skyId">{{
+  skyComponentProperties?.ariaLabel ||
```
Blackbaud-SteveBrush: Could you instead assign this value to the former `ariaLabel` on the lookup component, below?

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.html
johnhwhite: The element with `id="resultsListId"` previously only existed when `searchText && (!showActionsArea || searchResults.length > 0)`, which is not always the same as `isOpen`. In axe-core 4.7, there is a new reference check, and it flagged the `aria-controls` value as an invalid reference.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.spec.ts
johnhwhite: This test is intended to test the accessibility of the state where the input has focus and nothing has been entered yet, but show more is enabled. That was a case where the test would break in axe 4.7.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
johnhwhite: Good catch. Updated.

libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.html
johnhwhite: Yes, that field can be undefined and I'm not clear how this hasn't come up previously.

libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.html
johnhwhite: Updated to use the `ariaLabel` inputs.

libs/components/lookup/testing/src/autocomplete/autocomplete-input-harness.ts
johnhwhite: Good call.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete-input.directive.ts
johnhwhite: Removed.

libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.html
johnhwhite: When using `ariaLabel` and not `ariaLabelledby`, there was an issue that the combobox did not have a label. I've updated lookup to fix that issue.

libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.html
Blackbaud-SteveBrush: Interesting. Good catch.

libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.html
```diff
@@ -38,7 +46,15 @@
...
   <sky-lookup
-    [ariaLabel]="skyComponentProperties?.ariaLabel"
+    [ariaLabel]="
+      skyComponentProperties?.ariaLabelledBy
+        ? undefined
+        : skyComponentProperties?.ariaLabel ||
+          colDef?.headerName ||
+          colDef?.headerTooltip ||
+          colDef?.field ||
+          colDef?.colId
```
Blackbaud-SteveBrush: This grouping of possible labels appears twice in the template. To ensure both places get updated in the case of a change, let's make it a `protected` property of the class and write unit tests for it.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.html
Blackbaud-SteveBrush: Fair. Could you instead use the same condition to set `aria-controls` that is used to show the dropdown?
```
[attr.aria-controls]="
  searchText && (!showActionsArea || searchResults.length > 0)
    ? resultsListId
    : null
"
```
Better yet, put that logic in a public property so it's not copy/pasted throughout the template? Something like `isDropdownVisible`.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.spec.ts
Blackbaud-SteveBrush: The test should still show the dropdown with the "Show more" link, correct? It's not visible when `destroyAfterEach` is set to `false`. 🤔

libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.html
johnhwhite: Updated.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.html
johnhwhite: Refactored to use `isDropdownVisible`.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.spec.ts
johnhwhite: When I set a breakpoint in Chrome, the overlay is there prior to running the `toBeAccessible()` test, but gone when it finishes. Something about the test causes the overlay to dismiss. The test fails without the changes to the component, so I think it's a valid test that verifies the fix.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.html
```diff
@@ -32,7 +32,7 @@
 
 <ng-template #resultControlsTemplateRef>
   <div
-    *ngIf="searchText && (!showActionsArea || searchResults.length > 0)"
+    *ngIf="isDropdownVisible | async"
```
Blackbaud-SteveBrush, line 35: Love it. Much cleaner! 💪

libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.ts
```diff
@@ -60,6 +62,13 @@ export class SkyAgGridCellEditorLookupComponent
 
     this.#triggerType = SkyAgGridCellEditorUtils.getEditorInitialAction(params);
     const control = this.editorForm.get('selection');
+    this.ariaLabel = params.skyComponentProperties?.ariaLabelledBy
+      ? undefined
+      : params.skyComponentProperties?.ariaLabel ||
+        params.colDef.headerName ||
+        params.colDef.headerTooltip ||
+        params.colDef.field ||
+        params.colDef.colId;
```
Blackbaud-SteveBrush, line 78: Can any of those properties change during a change detection cycle? Do we need to respond to those changes?

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -469,10 +484,14 @@ export class SkyAutocompleteComponent implements OnDestroy, AfterViewInit {
 
   #_descriptorProperty = 'name';
 
+  #_enableShowMore = false;
+
   #_highlightText: string[] | undefined;
 
   #_inputDirective: SkyAutocompleteInputDirective | undefined;
 
+  #_isDropdownVisible = new BehaviorSubject<boolean>(false);
```
Blackbaud-SteveBrush: We reserve the `#_` syntax for get/set properties. I think we'd be safe to do the following:
```
protected isDropdownVisible: Observable<boolean>;

#isDropdownVisible = new BehaviorSubject<boolean>(false);
```

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -1033,4 +1062,13 @@ export class SkyAutocompleteComponent implements OnDestroy, AfterViewInit {
       });
     }
   }
+
+  #updateIsDropdownVisible(): void {
+    const isDropdownVisible =
+      !!this.searchText &&
+      (!this.showActionsArea || this.searchResults.length > 0);
```
Blackbaud-SteveBrush, line 1078: Love having the logic out of the template! 👍

libs/components/lookup/src/lib/modules/lookup/lookup.component.html
```diff
@@ -12,8 +12,15 @@
     }"
     #lookupWrapper
   >
+    <label
+      class="sky-screen-reader-only"
+      skyId
+      #labelElement="skyId"
+      [attr.aria-hidden]="!ariaLabel"
+      >{{ ariaLabel }}</label
```
Blackbaud-SteveBrush: We're already setting the `aria-label` on the `textarea` element. Is that not sufficient for `axe-core`?

libs/components/ag-grid/src/lib/modules/ag-grid/cell-editors/cell-editor-lookup/cell-editor-lookup.component.ts
johnhwhite: There is not an API to change a column definition. There is an event for "new columns loaded" but that doesn't make sense. Renderers have a `refresh` method but editors do not.

libs/components/lookup/src/lib/modules/lookup/lookup.component.html
Blackbaud-SteveBrush: Related? https://github.com/blackbaud/skyux/pull/1494#discussion_r1291273163

johnhwhite: Every `[role="combobox"]` [needs an accessible name](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role) (per MDN). An alternative would be to use `aria-labelled-by` to point to the `textarea` so it uses `textarea`'s label.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
johnhwhite: Updated.

libs/components/lookup/src/lib/modules/lookup/lookup.component.html
Blackbaud-SteveBrush: Yeah, that might be better. We need to also consider how the input box is handling the label element (in easy mode) and preferably connect the combobox with the label element it creates automatically. The code examples for the lookup component (in our docs) create a label element automatically if you need a reference point. This would also apply to the AG Grid pull request: https://github.com/blackbaud/skyux/pull/1548/files#r1296246221

johnhwhite: Updated `aria-labelledby` to fall back to the `controlId` and updated `controlId` to have a value when not using the host service.

Blackbaud-SteveBrush (dismissed):

Blackbaud-SteveBrush (approved):

blackbaud-sky-build-user:
Cherry-pick successful! :cherries: :tada: See https://github.com/blackbaud/skyux/pull/1607

PR #1557, opened 2023-08-08
ci: reduce the scope for e2e workflow

johnhwhite:
[AB#2639598](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2639598)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   NX_CLOUD_DISTRIBUTED_EXECUTION: true
+  IS_AUTO_RELEASE: ${{ startsWith( github.event.pull_request.labels[0].name, 'autorelease' ) }}
 
 jobs:
   install-deps:
@@ -17,30 +18,40 @@ jobs:
       node-version: ${{ steps.setup-node.outputs.node-version }}
     steps:
       - uses: actions/checkout@v3
+        if: ${{ env.IS_AUTO_RELEASE != 'true' }}
         with:
           # We need to fetch all branches and commits so that Nx affected has a base to compare against.
           fetch-depth: 0
       - uses: actions/setup-node@v3
+        if: ${{ env.IS_AUTO_RELEASE != 'true' }}
         id: setup-node
         with:
           node-version-file: '.nvmrc'
       # Rebase must happen before installing dependencies.
       - name: Rebase current branch
+        if: ${{ env.IS_AUTO_RELEASE != 'true' }}
         run: node ./scripts/rebase-pr.js
       - name: Cache node modules
+        if: ${{ env.IS_AUTO_RELEASE != 'true' }}
         id: cache
         uses: actions/cache@v3
         with:
           path: node_modules
           key: ${{ runner.os }}-node-${{ steps.setup-node.outputs.node-version }}-modules-${{ hashFiles('package-lock.json') }}
       - name: npm install
-        if: steps.cache.outputs.cache-hit != 'true'
+        if: ${{ env.IS_AUTO_RELEASE != 'true' && steps.cache.outputs.cache-hit != 'true' }}
         run: npm ci
       - name: Start Nx Cloud CI run
+        if: ${{ env.IS_AUTO_RELEASE != 'true' }}
         run: npx nx-cloud start-ci-run
+      - name: Skip
+        if: ${{ env.IS_AUTO_RELEASE == 'true' }}
+        run: |
+          echo "Skipping for release PR."
 
   agents:
     name: Bootup Nx Cloud agent
+    if: ${{ env.IS_AUTO_RELEASE != 'true' }}
     runs-on: ubuntu-latest
     needs: install-deps
     strategy:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -15,32 +15,47 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       node-version: ${{ steps.setup-node.outputs.node-version }}
+      is-auto-release: ${{ steps.is-auto-release.outputs.is-auto-release }}
     steps:
+      - name: Check auto release
+        id: is-auto-release
+        run: |
+          echo 'is-auto-release=${{ startsWith( github.event.pull_request.labels[0].name, 'autorelease' ) && 'true' || 'false' }}' >> $GITHUB_OUTPUT
```
Blackbaud-SteveBrush: Couldn't we put a condition at the job level?
```
jobs: 
  install-deps:
    if: ${{ startsWith(github.event.pull_request.labels[0].name, 'autorelease') }}
```

Blackbaud-SteveBrush: (You could still set that boolean as an output variable, like you're doing already.)

.github/workflows/e2e.yml
```diff
@@ -47,15 +56,42 @@ jobs:
...
           key: ${{ runner.os }}-node-${{ steps.setup-node.outputs.node-version }}-modules-${{ hashFiles('package-lock.json') }}
       - name: npm install
-        if: steps.cache.outputs.cache-hit != 'true'
+        if: ${{ steps.is-auto-release.outputs.is-auto-release != 'true' && steps.cache.outputs.cache-hit != 'true' }}
         run: npm ci
-      - name: Derive appropriate SHAs for base and head for `nx affected` commands
+      - name: Derive appropriate SHAs for base and head for `nx affected` commands in PR.
+        if: ${{ steps.is-auto-release.outputs.is-auto-release != 'true' && github.event_name == 'pull_request_target' }}
         uses: nrwl/nx-set-shas@v3
+      - name: Derive appropriate SHAs for base and head for `nx affected` commands since the last release.
```
Blackbaud-SteveBrush: My understanding of Nx is that it can still determine affected files on a push event (not just for pull requests). What part of that am I misunderstanding that requires this change?

.github/workflows/e2e.yml
```diff
@@ -47,15 +56,42 @@ jobs:
...
+          echo "Using NX_HEAD ${NX_HEAD}"
+          echo "NX_HEAD=${NX_HEAD}" >> $GITHUB_ENV
       - name: Set workflow parameters
         id: parameters
         shell: bash
         run: |
           set -exo pipefail
+
+          if [ "${{ steps.is-auto-release.outputs.is-auto-release }}" == "true" ]; then
+            # Release Please PR, quickly return "skip" parameters.
```
Blackbaud-SteveBrush: Is there a different way to word this? I'm unsure what "skip" parameters are. Maybe describe the "why" so that future engineers can piece this part out. Something like the following? "Abort building pull request preview applications for releases."

.github/workflows/e2e.yml
```diff
@@ -186,7 +218,7 @@ jobs:
           if-no-files-found: error
         if: ${{ fromJson(needs.install-deps.outputs.parameters).ghPagesRepo == 'skyux-pr-preview' }}
       - name: Skip
-        run: echo "Skip building ${{ matrix.app }} app. Only Storybook is published on branch builds."
+        run: echo "Skip."
```
Blackbaud-SteveBrush: I liked how this comment was formulated before since it gave a reason the step was being skipped. Is it possible to provide a reason?

.github/workflows/e2e.yml
```diff
@@ -204,22 +236,27 @@ jobs:
         with:
           fetch-depth: 0
           ref: ${{ github.head_ref || github.ref_name || github.event.ref }}
+        if: ${{ fromJson(needs.install-deps.outputs.parameters).ghPagesRepo != 'skip' }}
```
Blackbaud-SteveBrush: Possible to put this condition on the job itself?

.github/workflows/e2e.yml
```diff
@@ -247,13 +286,18 @@ jobs:
...
       - name: Publish preview and comment on PR
         run: |
           npx skyux-dev publish-storybook \
             --workingDirectory=${{ fromJson(needs.install-deps.outputs.parameters).ghPagesRepo }} \
             --pr=${{ needs.install-deps.outputs.workflow_dispatch_pr_number || github.event.number }}
         env:
           GITHUB_TOKEN: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
+        if: ${{ fromJson(needs.install-deps.outputs.parameters).ghPagesRepo != 'skip' }}
+      - name: Skip
+        run: echo "Skip."
```
Blackbaud-SteveBrush: Ditto.

nx.json
```diff
@@ -45,46 +45,63 @@
...
-      "{workspaceRoot}/jest.config.json",
-      "{workspaceRoot}/tsconfig.base.json"
+      "{workspaceRoot}/.eslintrc.json"
     ],
-    "production": [
-      "default",
+    "storybookInputs": [
+      "{projectRoot}/**/*",
+      "{workspaceRoot}/.babelrc",
+      "{workspaceRoot}/.storybook/**/*",
```
Blackbaud-SteveBrush, line 88: Should we include changes to the `e2e.yml` workflow?
```
"{workspaceRoot}/.github/workflows/e2e.yml",
```

nx.json
```diff
@@ -45,46 +45,63 @@
...
-        "{workspaceRoot}/.storybook/**/*",
-        "!{projectRoot}/.storybook/**/*"
-      ]
+      "inputs": ["storybookInputs", "^buildInputs"]
     }
   },
   "namedInputs": {
-    "default": ["{projectRoot}/**/*", "sharedGlobals"],
-    "sharedGlobals": [
+    "buildInputs": [
```
Blackbaud-SteveBrush, line 64: Do all of these inputs include changes to the `package.json`, `package-lock.json`, and `tsconfig.base.json` files, as well as the `/scripts` directory? All four of these areas can have an impact on `build`, `test`, `lint`, and `e2e`, so we should probably run the full suite if any of them change.

Blackbaud-SteveBrush (commented):
Love these changes! Just a few comments.

.github/workflows/ci.yml
johnhwhite: If the job is skipped, [everything that needs the job is also skipped](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds).

.github/workflows/e2e.yml
johnhwhite: Oh, nice. Ok, switching to that.

.github/workflows/e2e.yml
johnhwhite: Same as above. If the job is skipped, [everything that needs the job is also skipped](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds).

.github/workflows/e2e.yml
johnhwhite: Same as above. If the job is skipped, [everything that needs the job is also skipped](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds).

nx.json
johnhwhite: Right now I'd like to able to make small changes to the workflow without triggering all e2e tests to run.

nx.json
johnhwhite: NX handles `package.json` and `tsconfig.base.json`. I'll add `scripts`.

.github/workflows/e2e.yml
johnhwhite: We're using `skip` in the workflow generated from dev infra — if the array is empty, the step is never satisfied for the `verify` step, so `skip` is the keyword we're using to say do nothing.

I've restructured `verify` so it does not need publish (that was intended to add delay for e2e tests to finish but it wasn't necessary), which means `build-apps`, `build-storybook`, and `publish` jobs can all be job-level omitted (like you suggested above). That change simplifies this block.

.github/workflows/e2e.yml
johnhwhite: I've restructured `verify` so it does not need builds, so this can be job-level omitted and the comment can stay as-is.

nx.json
Blackbaud-SteveBrush: How would you test your changes to the workflow? Manually triggering them?

johnhwhite: Right. I anticipate needing to confirm that no tests were triggered in some cases. I can make small changes in e.g. an `integration-e2e` test if I need a test to run.

Blackbaud-SteveBrush (approved):

blackbaud-sky-build-user:
Cherry-pick [failed](https://github.com/blackbaud/skyux/actions/runs/5836383429)! :x: Please resolve conflicts and create a new PR.

PR #1563, opened 2023-08-10
fix(components/modals): add support for standalone modal components

Blackbaud-SteveBrush:


libs/components/core/src/lib/modules/dynamic-component/dynamic-component-options.ts
```diff
@@ -27,7 +31,7 @@ export interface SkyDynamicComponentOptions {
   viewContainerRef?: ViewContainerRef;
 
   /**
-   * Parent injector to use instead of the dynamic component service's injector
+   * The environment injector to use instead of the dynamic component service's injector.
    */
-  parentInjector?: Injector;
+  environmentInjector?: EnvironmentInjector;
```
Blackbaud-SteveBrush, line 36: We should add custom providers to the environment injector (i.e. the module injector) instead of the element injector because standalone components create a "standalone injector" as a child of its environment injector. See: https://angular.io/guide/standalone-components#standalone-injectors

johnhwhite (approved):

PR #1568, opened 2023-08-10
fix(components/i18n)!: move SkyLibResourcesService to SkyI18nModule providers

Blackbaud-SteveBrush:
BREAKING CHANGE: Library authors injecting `SkyLibResourcesService` into their component must import `SkyI18nModule` into that component's module providers.


Blackbaud-SteveBrush:
Before this change, if you had a `providedIn: 'root'` service inject `SkyLibResourcesService`, you would not see an error in the console. After this change, consumers will now get the following error:
```
NullInjectorError: No provider for SkyLibResourcesService!
```
It's worth noting that neither implementation would have produced a valid resource string without importing the `SkyI18nModule` into the `AppModule`.

Blackbaud-PaulCrowder (approved):

PR #1576, opened 2023-08-14
feat(components/ag-grid)!: move `SkyAgGridService` to `SkyAgGridModule` providers

Blackbaud-SteveBrush:
BREAKING CHANGE: The `SkyAgGridService` cannot be used in isolation; any component that injects `SkyAgGridService` must also import `SkyAgGridModule` into its module's providers.

### DIFF
--- a/libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-adapter.service.ts
+++ b/libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-adapter.service.ts
@@ -1,6 +1,9 @@
 import { Injectable } from '@angular/core';
 import { SkyCoreAdapterService } from '@skyux/core';
 
+/**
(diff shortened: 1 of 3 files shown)


libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-adapter.service.ts
```diff
@@ -1,6 +1,9 @@
 import { Injectable } from '@angular/core';
 import { SkyCoreAdapterService } from '@skyux/core';
 
+/**
+ * @internal
```
Blackbaud-SteveBrush, line 5: This file isn't exported, but I added the `@internal` just so we know at a glance that this service isn't part of the public API.

johnhwhite (approved):

PR #1592, opened 2023-08-16
refactor(components/core): make environment injector required when creating dynamic components

Blackbaud-SteveBrush:
Because we're switching the `SkyDynamicComponentService` from `providedIn: 'any'` to `providedIn: 'root'` there's no reason that consuming services/components shouldn't provide their environment injector when creating dynamic components. This will also safeguard future usage of the `SkyDynamicComponentService` since it will require the author to make a decision about which environment injector to use in its implementation.

### DIFF
--- a/apps/code-examples/src/app/features/progress-indicator.module.ts
+++ b/apps/code-examples/src/app/features/progress-indicator.module.ts
@@ -9,8 +9,8 @@ import { WaterfallIndicatorDocsComponent } from '../code-examples/progress-indic
 import { SkyProgressIndicatorPassiveDemoModule as SkyProgressIndicatorWaterfallDemoModule } from '../code-examples/progress-indicator/waterfall-indicator/basic/progress-indicator-waterfall-demo.module';
 import { WaterfallIndicatorDocsComponent as ProgressIndicatorWaterfallInlineHelpDemoComponent } from '../code-examples/progress-indicator/waterfall-indicator/inline-help/progress-indicator-waterfall-demo.component';
 import { SkyProgressIndicatorPassiveDemoModule as ProgressIndicatorWaterfallInlineHelpModule } from '../code-examples/progress-indicator/waterfall-indicator/inline-help/progress-indicator-waterfall-demo.module';
+import { WizardDemoComponent } from '../code-examples/progress-indicator/wizard/basic/wizard-demo.component';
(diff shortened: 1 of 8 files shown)


Blackbaud-SteveBrush:
Requires: https://github.com/blackbaud/skyux/pull/1582

apps/code-examples/src/app/features/progress-indicator.module.ts
```diff
@@ -9,8 +9,8 @@ import { WaterfallIndicatorDocsComponent } from '../code-examples/progress-indic
 import { SkyProgressIndicatorPassiveDemoModule as SkyProgressIndicatorWaterfallDemoModule } from '../code-examples/progress-indicator/waterfall-indicator/basic/progress-indicator-waterfall-demo.module';
 import { WaterfallIndicatorDocsComponent as ProgressIndicatorWaterfallInlineHelpDemoComponent } from '../code-examples/progress-indicator/waterfall-indicator/inline-help/progress-indicator-waterfall-demo.component';
 import { SkyProgressIndicatorPassiveDemoModule as ProgressIndicatorWaterfallInlineHelpModule } from '../code-examples/progress-indicator/waterfall-indicator/inline-help/progress-indicator-waterfall-demo.module';
+import { WizardDemoComponent } from '../code-examples/progress-indicator/wizard/basic/wizard-demo.component';
```
Blackbaud-SteveBrush, line 12: We were importing the wrong component.

libs/components/core/src/lib/modules/live-announcer/live-announcer.service.ts
```diff
@@ -11,7 +11,9 @@ import { SkyLiveAnnouncerArgs } from './types/live-announcer-args';
  * Allows for announcing messages to screen reader users through the use of a common `aria-live` element.
  * @internal
  */
-@Injectable({ providedIn: 'root' })
+@Injectable({
+  providedIn: 'root',
```
Blackbaud-SteveBrush, line 15: Just a formatting change for consistency with other services.

libs/components/core/src/lib/modules/overlay/overlay.service.ts
```diff
@@ -15,34 +16,18 @@ import { SkyOverlayComponent } from './overlay.component';
 
 /**
  * This service is used to create new overlays.
- * @dynamic
+ * @internal
```
Blackbaud-SteveBrush, line 19: We don't export this service, but let's add `@internal` to make it obvious it's not part of the public API.

johnhwhite (approved):

PR #1611, opened 2023-08-18
feat(components/ag-grid)!: `SkyCellType.Date` uses the same date formatting logic as the `SkyDatePipe`

Blackbaud-CoreyArcher:
[AB#1595291](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/1595291)

### DIFF
--- a/libs/components/datetime/src/lib/modules/date-pipe/date.service.ts
+++ b/libs/components/datetime/src/lib/modules/date-pipe/date.service.ts
@@ -2,13 +2,24 @@
 // behavior of using the `Intl` API for formatting dates rather than having to register every
 // supported locale.
 // https://github.com/angular/angular/blob/4.4.x/packages/common/src/pipes/date_pipe.ts
-import { SkyIntlDateFormatter } from '@skyux/i18n';
+import { Injectable, inject } from '@angular/core';
+import {
+  SkyAppLocaleInfo,
+  SkyAppLocaleProvider,
+  SkyIntlDateFormatter,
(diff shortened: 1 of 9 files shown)


libs/components/datetime/src/lib/modules/date-pipe/date.service.ts
```diff
@@ -20,12 +31,26 @@ export class SkyDateFormatUtility {
...
-  public static format(
-    locale: string,
+  #defaultFormat = 'short';
+  #defaultLocale = 'en-US';
+  #ngUnsubscribe = new Subject<void>();
+
+  constructor() {
+    inject(SkyAppLocaleProvider)
+      .getLocaleInfo()
+      .pipe(takeUntil(this.#ngUnsubscribe))
```
Blackbaud-TrevorBurch, line 41: `ngUnsubscribe` is never being resolved. We need an `ngOnDestroy` here to do that.

Blackbaud-TrevorBurch (changes requested):

libs/components/datetime/src/lib/modules/date-pipe/date.service.ts
Blackbaud-CoreyArcher: added

Blackbaud-CoreyArcher: added

Blackbaud-TrevorBurch (approved):

PR #1618, opened 2023-08-22
fix(components/lists): repeater item spacing in split view

Blackbaud-ErikaMcVey:


libs/components/lists/src/lib/modules/repeater/repeater-item.component.scss
```diff
@@ -163,6 +163,17 @@ sky-repeater-item {
 }
 
 .sky-theme-modern {
+  .sky-split-view-drawer {
+    sky-repeater-item .sky-repeater-item {
+      padding-left: var(--sky-padding-even-md);
+
+      &.sky-repeater-item-active {
+        // 15px padding minus 4px border
+        padding-left: 11px;
```
Blackbaud-SandhyaRajasabeson: instead of hardcoding this, can we get the 15-4px from the variables used for padding and border?
looking at the code that's `$sky-nav-selected-border-width` with 4px and the padding you've added above

Blackbaud-SandhyaRajasabeson (approved):

PR #1628, opened 2023-08-23
fix(components/tabs): vertical tab modern theme spacing has been updated to match design specifications

Blackbaud-TrevorBurch:
[AB#2612934](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2612934)

I chatted with @Blackbaud-AdamFunderburk about this and we came up with a few decisions

1. The padding on the tab buttons should be 15px to the left and not 18px
2. The 4px for the left border should be included in the 15px to the left (1 & 2 are based on what we are doing for split view to maintain consistency)
3. The left border for the active state should be 4px instead of 3px to match split view.
4. The top and bottom padding on a tab button should be 10px and not the 15px in the story. This is in the original spec and we could not find evidence as to where the 15px came from.
5. Tab groups should have 15px between them
6. Margins should be maintained on top when the tab or group is the first tab or group. The same applies for the bottom margin on last-of-kind elements
7. We are ok that this will affect sectioned form
8. Tabs should have 10px left margin along with the 15px padding on the tab button (just like is specified for tab groups)

BEGIN_COMMIT_OVERRIDE
fix(components/tabs)!: vertical tab modern theme spacing has been updated to match design specifications
END_COMMIT_OVERRIDE

### DIFF
--- a/libs/components/tabs/src/lib/modules/vertical-tabset/vertical-tab.component.scss
+++ b/libs/components/tabs/src/lib/modules/vertical-tabset/vertical-tab.component.scss
@@ -64,14 +64,14 @@
     color: $sky-text-color-deemphasized;
     cursor: pointer;
     padding: $sky-theme-modern-space-sm 0 $sky-theme-modern-space-sm
-      $sky-theme-modern-space-sm;
-    margin: $sky-theme-modern-margin-inline-sm 0
-      $sky-theme-modern-margin-inline-lg $sky-theme-modern-margin-inline-sm;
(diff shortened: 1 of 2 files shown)


Blackbaud-ErikaMcVey (approved):

Blackbaud-ErikaMcVey:
I put risk level 5 because you put 5 but since no action is required by the consumer and its just things shifting around a bit I'm not sure I truly consider it a breaking change

PR #1642, opened 2023-08-24
feat(components/tabs): make all tabs responsive containers

Blackbaud-ErikaMcVey:


libs/components/tabs/src/lib/modules/tabs/tab.component.ts
```diff
@@ -188,12 +186,11 @@ export class SkyTabComponent implements OnChanges, OnDestroy {
 
   @ViewChild('tabContentWrapper')
   public set tabContentWrapper(tabContentWrapper: ElementRef | undefined) {
-    this.#_tabElement = tabContentWrapper;
-    this.#observeTabResize();
-  }
-
-  public get tabContentWrapper(): ElementRef | undefined {
-    return this.#_tabElement;
+    if (tabContentWrapper) {
```
Blackbaud-TrevorBurch, line 190: I know that it shouldn't happen.... but should we have an else statement here that unobserves if somehow `tabContentWrapper` goes undefined?

Blackbaud-TrevorBurch (approved):

PR #1644, opened 2023-08-24
feat(components/datetime)!: timepicker hardcoded `aria-label` is removed and the label should now be provided through a wrapping input box

Blackbaud-TrevorBurch:
[AB#2346441](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2346441)

### DIFF
--- a/libs/components/datetime/src/lib/modules/timepicker/fixtures/timepicker-input-box-component.fixture.html
+++ b/libs/components/datetime/src/lib/modules/timepicker/fixtures/timepicker-input-box-component.fixture.html
@@ -1,6 +1,5 @@
 <div>
   <sky-input-box labelText="Time">
-    <label class="sky-control-label"> Input box test </label>
     <sky-timepicker #timePicker>
       <input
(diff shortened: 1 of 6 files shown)


libs/components/datetime/src/lib/modules/timepicker/fixtures/timepicker-input-box-component.fixture.html
```diff
@@ -1,5 +1,5 @@
 <div>
-  <sky-input-box>
+  <sky-input-box labelText="Time">
```
Blackbaud-ErikaMcVey, line 2: I know its just a test fixture but if you're using `labelText` then you don't need the label on the line below

Blackbaud-TrevorBurch: Fixed

Blackbaud-ErikaMcVey (approved):

PR #1701, opened 2023-09-09
feat(components/pages): provide schematic for updating `SkyPageComponent`

johnhwhite:
[AB#2667662](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2667662)

### DIFF
--- a/libs/components/packages/src/schematics/migrations/update-9/move-page-component/move-page-component.schematic.ts
+++ b/libs/components/packages/src/schematics/migrations/update-9/move-page-component/move-page-component.schematic.ts
@@ -21,7 +21,7 @@ export default function movePageComponent(): Rule {
 
       // Visit all TypeScript files in each project.
       tree.getDir(source).visit((path) => {
-        if (path.endsWith('.ts') && !path.includes('node_modules')) {
+        if (path.endsWith('.ts')) {
           // Parse the TypeScript file.
           const source = ts.createSourceFile(
             path,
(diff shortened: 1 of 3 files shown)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/1701/storybook/)

Component Storybooks:

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/1701/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/1701/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/1701/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/1701/playground/)

libs/components/packages/src/schematics/migrations/update-9/move-page-component/move-page-component.schematic.ts
```diff
@@ -0,0 +1,143 @@
...
+      });
+    });
+
+    // Add the dependency to @skyux/pages.
+    if (addDependency) {
+      const pages = getPackageJsonDependency(tree, '@skyux/pages');
+      if (!pages) {
+        const layout = getPackageJsonDependency(tree, '@skyux/layout');
+        if (layout) {
+          addPackageJsonDependency(tree, {
```
Blackbaud-ErikaMcVey, line 133: ok this is the only place I'm getting a little lost. `addDependency` was set to true if there were any imports from the pages library. Makes sense. They might already have that dependency, so first we get it and then only perform adding logic if they don't have it. Makes sense.

But then we get the layout package, and only add pages if they have layout? I suppose they would always have it since they were using that previously, and that gives the version number. Should we remove the layout dependency if they don't have any other imports from that library? Admittedly that sounds like it would be a bit tricky/overkill.

Blackbaud-ErikaMcVey (approved):

libs/components/packages/src/schematics/migrations/update-9/move-page-component/move-page-component.schematic.ts
johnhwhite: If they don't have layout it's not clear how we got here. It's using the version number from layout. It could have used core. Since pages depends on layout it's ok to leave it.

PR #1761, opened 2023-09-27
chore: update `SkyContentInfoProvider` to optionally use element ID for descriptor

Blackbaud-ErikaMcVey:


libs/components/core/src/lib/modules/content-info-provider/content-info-provider.spec.ts
```diff
@@ -18,31 +19,42 @@ describe('SkyContentInfoProvider', () => {
...
-    const contentDescriptorValue = 'value';
+    const contentDescriptor: SkyDescriptor = {
+      type: 'text',
+      value: 'value',
+    };
     const obs = contentInfoProvider.getInfo();
 
     obs?.subscribe((value) => {
-      expect(value.descriptor).toEqual(contentDescriptorValue);
+      expect(value.descriptor?.value).toEqual(contentDescriptor.value);
```
Blackbaud-PaulCrowder: Should this check both the properties on `contentDescriptor`?

```
expect(value.descriptor).toEqual(contentDescriptor);
```

libs/components/core/src/lib/modules/content-info-provider/descriptor-type.ts
```diff
@@ -0,0 +1,4 @@
+export enum SkyDescriptorType {
```
Blackbaud-PaulCrowder: Did you mean to use this for the descriptor interface's `type` property? If so I recommend using a string union instead of a string enum as those are easier to use in templates.

libs/components/core/src/lib/modules/content-info-provider/descriptor.ts
```diff
@@ -0,0 +1,4 @@
+export type SkyDescriptor = {
```
Blackbaud-PaulCrowder: I suggest using `SkyContentInfoDescriptor` so its obvious that it's for use with `SkyContentInfo`. Same with `SkyContentInfoDescriptorType`.

blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/1761/storybook/)

Component Storybooks:

- [a11y-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/a11y-storybook/)
- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/colorpicker-storybook/)
- [core-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/core-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/errors-storybook/)
- [flyout-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/flyout-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/forms-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/lookup-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/modals-storybook/)
- [navbar-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/navbar-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/phone-field-storybook/)
- [popovers-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/popovers-storybook/)
- [progress-indicator-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/progress-indicator-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/theme-storybook/)
- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/tiles-storybook/)
- [toast-storybook](https://blackbaud.github.io/skyux-pr-preview/1761/storybooks/toast-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/1761/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/1761/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/1761/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/1761/playground/)

libs/components/layout/src/lib/modules/box/box-header.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="sky-box-header">
-  <span skyTrim><ng-content></ng-content></span
+  <span id="sky-sr-label-box" skyTrim><ng-content></ng-content></span
```
Blackbaud-PaulCrowder: Hardcoding an ID here will cause an ID clash when there is more than one box header component on a page and break accessibility. Since you reference this same ID in the box component, use [`SkyIdService`](https://developer.blackbaud.com/skyux/components/id#class-skyidservice) in that component to generate an ID and provide it to the box header component using an injection token so the box header can then bind this element's `[attr.id]` attribute to it and pass it on to the content info service.

libs/components/layout/src/lib/modules/box/box.component.ts
```diff
@@ -38,4 +47,17 @@ export class SkyBoxComponent {
    */
   @Input()
   public ariaRole: string | undefined;
+
+  @ViewChild('titleRef', { read: ElementRef })
+  public set boxHeaderRef(value: ElementRef | undefined) {
+    if (value?.nativeElement?.textContent) {
```
Blackbaud-PaulCrowder: This is onceable; the text content of the element can change after `boxHeaderRef` is set. What happens if you always set the descriptor regardless of the contents of the header element? Wouldn't the screen reader just skip the element's empty content or a missing element?

libs/components/popovers/src/lib/modules/dropdown/dropdown.component.html
```diff
@@ -1,4 +1,10 @@
 <div class="sky-dropdown">
+  <div class="sky-screen-reader-only" id="sky-sr-label-context-menu">
```
Blackbaud-PaulCrowder: Same issue here with the hardcoded ID. Since you're not referencing it in your component, you can use the `skyId` directive instead of the service and then reference the ID elsewhere in the template.

```
<div class="sky-screen-reader-only" skyId #srLabelContextMenu="skyId">
...
<button [attr.aria-labelledby]="srLabelContextMenu.id">
```

libs/components/popovers/src/lib/modules/dropdown/dropdown.component.html
```diff
@@ -8,12 +14,20 @@
...
       (buttonType === 'context-menu'
-        ? (contentInfoObs | async)?.descriptor
+        ? (contentInfoObs | async)?.descriptor?.type === 'text'
           ? ('skyux_dropdown_context_menu_with_content_descriptor_default_label'
-            | skyLibResources : (contentInfoObs | async)?.descriptor)
+            | skyLibResources : (contentInfoObs | async)?.descriptor?.value)
           : ('skyux_dropdown_context_menu_default_label' | skyLibResources)
         : undefined)
     "
+    [attr.aria-labelledBy]="
```
Blackbaud-PaulCrowder: `aria-labelledby` should be all lower-case.

libs/components/popovers/src/lib/modules/dropdown/dropdown.component.spec.ts
```diff
@@ -1216,9 +1218,25 @@ describe('Dropdown component', function () {
...
-      contentInfoProvider.patchInfo({ descriptor: 'default label' });
+      contentInfoProvider.patchInfo({
+        descriptor: { type: 'elementId', value: 'sr-descriptor-label' },
+      });
+
+      detectChangesFakeAsync();
+      const button = getButtonElement();
+
+      expect(button?.getAttribute('aria-labelledBy')).toEqual(
+        'sky-sr-label-context-menu, sr-label-descriptor'
```
Blackbaud-PaulCrowder: This is expecting a comma, but IDs should only be separated by spaces.

libs/components/core/src/lib/modules/content-info-provider/descriptor-type.ts
Blackbaud-ErikaMcVey: I swear to god I deleted this file before I committed 🤯

libs/components/layout/src/lib/modules/box/box.component.ts
Blackbaud-ErikaMcVey: What happens is it doesn't ready anything for this empty element, but it does read the other non-empty element which says "context menu for". The problem is we always render `.sky-box-header-content` in box even if they don't specify any of the content that gets ng-contented in there--so if they don't include a header, as it is currently implemented their context menu would read "context menu for"

Unfortunately I think the answer is to move the "check" down a level. Box has to set the content info, but only box-header knows if a header element was actually provided--which if one was, I think we can safely assume there is content in it. I say unfortunately because I think that might mean another layer of viewchild so the parent can get the info from the child, but I haven't actually had a chance to think that through yet

libs/components/popovers/src/lib/modules/dropdown/dropdown.component.html
Blackbaud-ErikaMcVey: huh, surprised that still worked

libs/components/popovers/src/lib/modules/dropdown/dropdown.component.spec.ts
Blackbaud-ErikaMcVey: oops, forgot to update here--the test is currently failing

libs/components/popovers/src/lib/modules/dropdown/dropdown.component.html
Blackbaud-PaulCrowder: Native HTML attributes aren't case-sensitive, but our convention is to use all lower-case for element and attribute names.

libs/components/layout/src/lib/modules/box/box.component.ts
Blackbaud-PaulCrowder: We know if `sky-box-header` is present, though, and I think it's OK to assume that if that component is present, it has text content and that we should use its element ID for the content info. The consumer can always set the ARIA label directly on the context menu for the extreme edge case where `sky-box-header` is present but has no inner text.

Blackbaud-ErikaMcVey: We don't know if sky-box-header is present as it is currently written

libs/components/core/src/index.ts
```diff
@@ -17,6 +17,7 @@ export { SkyAffixer } from './lib/modules/affix/affixer';
 
 export { SkyContentInfoProvider } from './lib/modules/content-info-provider/content-info-provider';
 export { SkyContentInfo } from './lib/modules/content-info-provider/content-info';
+export { SkyContentInfoDescriptor as SkyDescriptor } from './lib/modules/content-info-provider/descriptor';
```
Blackbaud-PaulCrowder: The `as SkyDescriptor` isn't needed here. It's probably an artifact of renaming the class.

libs/components/popovers/src/lib/modules/shared/sky-popovers-resources.module.ts
```diff
@@ -23,6 +23,9 @@ const RESOURCES: { [locale: string]: SkyLibResources } = {
     skyux_dropdown_context_menu_with_content_descriptor_default_label: {
       message: 'Context menu for {0}',
     },
+    skyux_dropdown_context_menu_with_content_descriptor_el_default_label: {
+      message: 'Context menu for ',
```
Blackbaud-PaulCrowder, line 27: Did we decide to remove "for" from this description?

Blackbaud-ErikaMcVey: No we decided to leave it for now. At least that was what I announced at the end of flex and no one disagreed with me

Blackbaud-PaulCrowder (approved):

PR #1779, opened 2023-10-04
fix(components/toast): `SkyToastService` and `SkyToastLegacyService` add toast instances to the same pool

Blackbaud-SteveBrush:
`SkyToastLegacyService` and `SkyToastService` have their own copies of toast instances and toast streams. This fix makes sure both services contribute to the same pool of toast instances, and emit a single stream for consumers.

### DIFF
--- a/apps/code-examples/src/app/code-examples/angular-tree-component/angular-tree/advanced/demo.component.ts
+++ b/apps/code-examples/src/app/code-examples/angular-tree-component/angular-tree/advanced/demo.component.ts
@@ -33,7 +33,6 @@ import { AngularTreeDemoNode } from './node';
 @Component({
   standalone: true,
   selector: 'app-demo',
-  styleUrls: ['./demo.component.scss'],
   templateUrl: './demo.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
   imports: [
--- a/libs/components/toast/src/lib/modules/toast/toast.service.ts
+++ b/libs/components/toast/src/lib/modules/toast/toast.service.ts
@@ -27,7 +27,7 @@ import { SkyToastConfig } from './types/toast-config';
 export class SkyToastService implements OnDestroy {
   private static host: ComponentRef<SkyToasterComponent> | undefined;
   private static toasts: SkyToast[] = [];
-  private static toastStream = new BehaviorSubject<SkyToast[]>([]);
+  private static toastStream: BehaviorSubject<SkyToast[]>;
 
   /**
    * @internal
@@ -41,6 +41,7 @@ export class SkyToastService implements OnDestroy {
 
   constructor(dynamicComponentService: SkyDynamicComponentService) {
     this.#dynamicComponentService = dynamicComponentService;
+    SkyToastService.toastStream = new BehaviorSubject<SkyToast[]>([]);
   }
 
   public ngOnDestroy(): void {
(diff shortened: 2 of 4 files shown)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/1779/storybook/)

Component Storybooks:

- [flyout-storybook](https://blackbaud.github.io/skyux-pr-preview/1779/storybooks/flyout-storybook/)
- [toast-storybook](https://blackbaud.github.io/skyux-pr-preview/1779/storybooks/toast-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/1779/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/1779/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/1779/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/1779/playground/)

apps/code-examples/src/app/code-examples/angular-tree-component/angular-tree/advanced/demo.component.ts
```diff
@@ -33,7 +33,6 @@ import { AngularTreeDemoNode } from './node';
 @Component({
   standalone: true,
   selector: 'app-demo',
-  styleUrls: ['./demo.component.scss'],
```
Blackbaud-SteveBrush, line 36: This stylesheet is empty and can be removed from the code example.

libs/components/toast/src/lib/modules/toast/toast.service.ts
```diff
@@ -26,21 +26,22 @@ import { SkyToastConfig } from './types/toast-config';
 })
 export class SkyToastService implements OnDestroy {
   private static host: ComponentRef<SkyToasterComponent> | undefined;
+  private static toasts: SkyToast[] = [];
+  private static toastStream: BehaviorSubject<SkyToast[]>;
```
Blackbaud-SteveBrush, line 30: `toasts` and `toastStream` need to be static so that both `SkyToastService` and `SkyToastLegacyService` contribute to the same pool of toast instances.

johnhwhite (approved):

libs/components/toast/src/lib/modules/toast/toast.service.ts
```diff
@@ -49,7 +50,8 @@ export class SkyToastService implements OnDestroy {
       this.#removeHostComponent();
     }
 
-    this.#toastStream.complete();
+    SkyToastService.toasts = [];
+    SkyToastService.toastStream.complete();
```
Blackbaud-TrevorBurch, line 54: Couldn't this be problematic if two different services are spun up separately? Overwriting the static stream would mean anything in an existing stream would be wiped away - and completing the stream when one service is destroyed would cause the other to no longer work.

johnhwhite: The service is provided in root and the expectation is that there is only one instance.

Blackbaud-SteveBrush: Correct. Both the legacy and the normal service are provided in "root" (different from our other legacy services) so the only time this destroy hook would get called is if the app is destroyed.

Blackbaud-TrevorBurch: But with them both being provided in root - couldn't that mean there is a root instance of `SkyToastService` that is being used one place but another `SkyLegacyToastService` that is being provided somewhere else?

Blackbaud-TrevorBurch: Resolving as the chances and real use cases for this are very minimal.

PR #1783, opened 2023-10-04
feat(components/core): update content info provider to support element IDs, create screen reader label directive

Blackbaud-ErikaMcVey:
NOTE: the box changes will not be merged--I left them in for now in case anyone wanted to test the new functionality. You can see it in the playground for box. The labels don't get added in the dropdown playground since nothing else is provided aria labelledby IDs

- updates `SkyContentInfoProvider` to support 2 types of `descriptor`--an element ID to be used with `aria-labelledby` and a string
- updates all usages of `SkyContentInfoProvider` to use the new schema
- create `skySrLabel` directive, which takes the element it is placed on and puts it in a div within the document body. This is necessary because leaving the label element in the component updates the text content, which breaks consumer unit tests
- updates dropdown to use the new `SkyContentInfoProvider` element ID option, along with the sr label directive


libs/components/core/src/lib/modules/sr-label/sr-label.directive.ts
```diff
@@ -0,0 +1,75 @@
...
+    this.#insertSrLabel();
+  }
+
+  @Input()
+  public set destroyRef(ref: DestroyRef) {
+    ref.onDestroy(() => this.#removeSrLabel());
+  }
+
+  @Input()
+  public set createLabel(value: boolean | undefined) {
```
Blackbaud-ErikaMcVey: I had to go this `createLabel` route instead of putting an `*ngIf` on the label element because this will likely always be used in conjunction with `skyId`, and a template cannot reference an element that has any form of structural directive on it, because it interprets that as meaning the element might not be in the DOM to reference--even when we know it is there. You cannot `?.` or assert it is there or anything. It just won't let you reference it.

blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/1783/storybook/)

Component Storybooks:

- [a11y-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/a11y-storybook/)
- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/colorpicker-storybook/)
- [core-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/core-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/errors-storybook/)
- [flyout-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/flyout-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/forms-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/lookup-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/modals-storybook/)
- [navbar-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/navbar-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/phone-field-storybook/)
- [popovers-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/popovers-storybook/)
- [progress-indicator-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/progress-indicator-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/theme-storybook/)
- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/tiles-storybook/)
- [toast-storybook](https://blackbaud.github.io/skyux-pr-preview/1783/storybooks/toast-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/1783/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/1783/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/1783/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/1783/playground/)

libs/components/core/src/lib/modules/sr-label/sr-label.directive.ts
```diff
@@ -0,0 +1,75 @@
...
+} from '@angular/core';
+
+const SR_LABELS_CONTAINER_ID = 'sky-sr-labels-container';
+
+/**
+ * Adds the element to a screen reader only section of the body.
+ * This prevents components' DOM from including text only intended for screen readers.
+ */
+@Directive({
+  selector: '[skySrLabel]',
```
Blackbaud-SteveBrush: At first glance, I read this as "Senior Label". What do you think about expanding the abbreviation so it's more clear? `skyScreenReaderOnlyLabel`. If you must have the abbreviation, I would also capitalize the "R": `skySROnlyLabel` ("screen reader" is two words).

libs/components/core/src/lib/modules/sr-label/sr-label.directive.ts
```diff
@@ -0,0 +1,75 @@
...
+  standalone: true,
+})
+export class SkySrLabelDirective implements OnInit {
+  readonly #elementRef = inject(ElementRef);
```
Blackbaud-PaulCrowder: Private members usually go under public members but before the constructor.

libs/components/core/src/lib/modules/sr-label/sr-label.directive.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  @Input()
+  public set destroyRef(ref: DestroyRef) {
+    ref.onDestroy(() => this.#removeSrLabel());
+  }
+
+  @Input()
+  public set createLabel(value: boolean | undefined) {
+    this.#_createLabel = value ?? true;
```
Blackbaud-PaulCrowder: Having the default value here be `true` when `undefined` is specified is a bit confusing if it's supposed to act like `*ngIf`. Passing `undefined` to `*ngIf` is equivalent to `false`, and it's not uncommon to use a falsy statement and not a boolean for `*ngIf` (e.g. `<div *ngIf="someNonBooleanObject && someOtherNonBooleanObject" />`). This might be a case where we need to use a negative input name with a default value of `false` or `undefined`, like `srHidden`.

libs/components/layout/src/lib/modules/box/box-header.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="sky-box-header">
-  <span skyTrim><ng-content /></span
+  <span id="sky-sr-label-box" skyTrim><ng-content /></span
```
Blackbaud-PaulCrowder: This needs a dynamically-generated ID instead of a hard-coded ID.

libs/components/layout/src/lib/modules/box/box.component.html
```diff
@@ -9,7 +9,9 @@
   }"
 >
   <div class="sky-box-header-content">
-    <ng-content select="sky-box-header" />
+    <span #titleRef>
```
Blackbaud-PaulCrowder: Can this be an `ng-container` so we're not adding new elements to the DOM? I could see a consumer having a unit test that doesn't anticipate the child element.

libs/components/layout/src/lib/modules/box/box.component.ts
```diff
@@ -38,4 +47,17 @@ export class SkyBoxComponent {
    */
   @Input()
   public ariaRole: string | undefined;
+
+  @ViewChild('titleRef', { read: ElementRef })
```
Blackbaud-PaulCrowder: Make the selector and the property names match.

libs/components/layout/src/lib/modules/box/box.component.ts
```diff
@@ -38,4 +47,17 @@ export class SkyBoxComponent {
...
+  public set boxHeaderRef(value: ElementRef | undefined) {
+    console.log('hello');
+    console.log(value);
```
Blackbaud-PaulCrowder: `console.log()` calls should be removed.

libs/components/layout/src/lib/modules/toolbar/toolbar.component.ts
```diff
@@ -34,7 +34,13 @@ export class SkyToolbarComponent {
    */
   @Input()
   public set listDescriptor(value: string | undefined) {
-    this.#contentInfoProvider.patchInfo({ descriptor: value });
+    if (value) {
```
Blackbaud-PaulCrowder: Simplify to this?

```
this.#contentInfoProvider.patchInfo({
  descriptor: value ? { type: 'text', value } : undefined
});
```

libs/components/popovers/src/lib/modules/dropdown/dropdown.component.ts
```diff
@@ -210,6 +207,8 @@ export class SkyDropdownComponent implements OnInit, OnDestroy {
     return this.#_triggerButton;
   }
 
+  public destroyRef = inject(DestroyRef);
```
Blackbaud-PaulCrowder: This can be `protected` instead of `public` since it's only referenced in the template.

libs/components/angular-tree-component/src/lib/modules/angular-tree/angular-tree-node.component.ts
```diff
@@ -109,7 +109,9 @@ export class SkyAngularTreeNodeComponent implements AfterViewInit, OnInit {
 
   protected set nodeName(value: string) {
     this.#_nodeName = value;
-    this.#contentInfoProvider.patchInfo({ descriptor: value });
+    this.#contentInfoProvider.patchInfo({
+      descriptor: { type: 'text', value: value },
```
Blackbaud-PaulCrowder: You can shorten `{ value: value}` to `{ value }`.

libs/components/core/src/lib/modules/sr-label/sr-label.directive.ts
Blackbaud-ErikaMcVey: I'll change the name--I just think the default scenario would be always showing it, so I don't think we should have to specify to show it every time.

libs/components/layout/src/lib/modules/box/box-header.component.html
Blackbaud-ErikaMcVey: I explained this in the description and in stand up--I am going to remove all of the box changes and do it properly in a different PR. I left them in so that the reviewer could do manual testing if they wished, because otherwise you couldn't actually see the changes anywhere.

libs/components/core/src/lib/modules/sr-label/sr-label.directive.ts
Blackbaud-PaulCrowder: The change I proposed would mean you'd only specify it when it's hidden.

libs/components/core/src/lib/modules/content-info-provider/content-info.ts
```diff
@@ -1,3 +1,5 @@
+import { SkyContentInfoDescriptor } from './descriptor';
+
 export type SkyContentInfo = {
```
Blackbaud-SteveBrush, line 8: Some JSDocs would be helpful, especially for future engineers who might not understand what this is doing.

libs/components/core/src/lib/modules/content-info-provider/descriptor.ts
```diff
@@ -0,0 +1,4 @@
+export type SkyContentInfoDescriptor = {
+  type: 'text' | 'elementId';
```
Blackbaud-SteveBrush: JSDocs would be helpful here, too. Are `'text'` types always resource strings? Would it be more self-documenting to write this as `'resourceKey' | 'elementId'` so that people don't put plain English into the descriptor?

libs/components/core/src/lib/modules/sr-label/sr-label.directive.ts
```diff
@@ -0,0 +1,75 @@
...
+export class SkySrLabelDirective implements OnInit {
+  readonly #elementRef = inject(ElementRef);
+  #_createLabel = true;
+
+  public ngOnInit(): void {
+    this.#insertSrLabel();
+  }
+
+  @Input()
+  public set destroyRef(ref: DestroyRef) {
```
Blackbaud-SteveBrush: Do we need a destroy ref? Since this directive will know which element it created, it can remove that element in the `ngOnDestroy` callback. See my rough reproduction: https://stackblitz.com/edit/angular-phyizz?file=src%2Fapp%2Fscreen-reader-only.directive.ts,src%2Fapp%2Fdemo.component.ts,src%2Fapp%2Fapp.component.ts

libs/components/lookup/src/lib/modules/search/search.component.html
```diff
@@ -36,9 +36,10 @@
               ariaLabelledBy
                 ? undefined
                 : ariaLabel ||
-                  ((contentInfoObs | async)?.descriptor
+                  ((contentInfoObs | async)?.descriptor?.type === 'text'
                     ? ('skyux_search_aria_label_descriptor'
-                      | skyLibResources : (contentInfoObs | async)?.descriptor)
+                      | skyLibResources
+                        : (contentInfoObs | async)?.descriptor?.value)
```
Blackbaud-SteveBrush, line 42: We're repeating this logic a few times in this PR (and I'm assuming we'll need to repeat it for future implementations as well). Is there room to put this logic in a pipe or service? If it's not feasible, or doesn't add any value, we can move forward without changes.

libs/components/core/src/lib/modules/sr-label/sr-label.directive.ts
```diff
@@ -0,0 +1,75 @@
...
+    const srLabelsContainerEl = this.#getSrLabelsContainerEl();
+    if (this.createLabel) {
+      srLabelsContainerEl.append(this.#elementRef.nativeElement);
+    } else {
+      this.#removeSrLabel();
+    }
+  }
+
+  #getSrLabelsContainerEl(): HTMLElement {
+    let containerEl = document.getElementById(SR_LABELS_CONTAINER_ID);
```
Blackbaud-SteveBrush: Similar to flyout/modal/toast, we can store the container element as a private static variable so we don't have to query the DOM every time.
```
private static containerEl: HTMLDivElement | undefined;
```

libs/components/popovers/src/lib/modules/dropdown/dropdown.component.html
```diff
@@ -1,3 +1,18 @@
+<span
```
Blackbaud-SteveBrush: Can we bump this to the bottom of the file since it's a lower priority than the main section of the template?

libs/components/core/src/lib/modules/sr-label/sr-label.directive.ts
```diff
@@ -0,0 +1,75 @@
...
+    this.#insertSrLabel();
+  }
+
+  @Input()
+  public set destroyRef(ref: DestroyRef) {
+    ref.onDestroy(() => this.#removeSrLabel());
+  }
+
+  @Input()
+  public set createLabel(value: boolean | undefined) {
```
Blackbaud-SteveBrush: JSDocs would be helpful.

libs/components/core/src/lib/modules/sr-label/sr-label.directive.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  public get createLabel(): boolean {
+    return this.#_createLabel;
+  }
+
+  readonly #elementRef = inject(ElementRef);
+  #_createLabel = false;
+
+  #insertSrLabel(): void {
```
Blackbaud-SteveBrush: I would rename to `#updateScreenReaderLabel` (or `#updateLabel`) since it can both create and remove the label depending on a condition.

libs/components/core/src/lib/modules/sr-label/sr-label.directive.ts
```diff
@@ -0,0 +1,75 @@
...
+  Input,
+  OnInit,
+  inject,
+} from '@angular/core';
+
+const SR_LABELS_CONTAINER_ID = 'sky-sr-labels-container';
+
+/**
+ * Adds the element to a screen reader only section of the body.
+ * This prevents components' DOM from including text only intended for screen readers.
```
Blackbaud-SteveBrush: Need `@internal`?

libs/components/core/src/lib/modules/content-info-provider/content-info.ts
Blackbaud-SteveBrush, line 8: Need `@internal`?

libs/components/core/src/lib/modules/content-info-provider/descriptor.ts
```diff
@@ -0,0 +1,4 @@
+export type SkyContentInfoDescriptor = {
```
Blackbaud-SteveBrush: Need `@internal`?

libs/components/core/src/lib/modules/content-info-provider/descriptor.ts
Blackbaud-ErikaMcVey: No, text is never a resource key. It is a snippet of text that describes the content of the parent. If it needs to be localized the consumer will have done that already. I'll update JSDocs.

Blackbaud-SteveBrush (approved):

PR #1462, opened 2023-06-15
fix(components/config): ignore existing URL params when encoding config params

Blackbaud-SteveBrush:

PR #1800, opened 2023-10-09
feat(components/router): add `queryParams` to `skyHref` directive

Blackbaud-TomRamsey:
[AB#2704591](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2704591)


libs/components/router/src/lib/modules/href/href.directive.ts
```diff
@@ -44,6 +44,9 @@ export class SkyHrefDirective {
     return this.#_skyHref;
   }
 
+  @Input()
+  public queryParams?: Params | null;
```
Blackbaud-SteveBrush: I realize this is how Angular's router defines `queryParams`, but to maintain consistency with the SKY UX code base I would write it in the following format:
```
public queryParams: Params | undefined;
```
If `null` is required for the implementation, you could do:
```
public queryParams: Params | null | undefined;
```
(Whether or not we include the question mark will produce the same outcome, but we decided to omit it for class properties as a convention a while ago.)

libs/components/config/src/lib/params.spec.ts
```diff
@@ -319,6 +393,6 @@ describe('SkyAppRuntimeConfigParams', () => {
 
     expect(
       params.getLinkUrl('https://mysite.com?foobar=Robert+Hernandez')
-    ).toEqual('https://mysite.com?foobar=Robert+Hernandez&a1=%20');
+    ).toEqual('https://mysite.com?foobar=Robert%2BHernandez&a1=%20');
```
Blackbaud-SteveBrush: This should remain a `+` character; we have a consumer that returns encoded URLs from their backend (replacing spaces with "+") and we were double-encoding the params for them, breaking their service. Here's the original PR: https://github.com/blackbaud/skyux/pull/1462/files

libs/components/config/src/lib/params.ts
```diff
@@ -178,44 +179,67 @@ export class SkyAppRuntimeConfigParams {
 
   /**
    * Adds the current params to the supplied link URL.
+   * @param url The url to update
+   * @param queryParams Optional query parameters to include in the constructed url
    */
-  public getLinkUrl(url: string): string {
+  public getLinkUrl(url: string, queryParams?: Params | null): string {
```
Blackbaud-SteveBrush: Instead of using `Params` from `@angular/router` (and requiring a new peer dependency in `libs/components/config/package.json`, which would be considered a breaking change), I would just inline a static type:
```
queryParams?: Record<string, ...> | null
```

libs/components/config/src/lib/params.ts
```diff
@@ -178,44 +179,67 @@ export class SkyAppRuntimeConfigParams {
...
+    // Remove excluded params
+    for (const paramName of excludeParams.keys()) {
+      unifiedParams = unifiedParams.delete(paramName);
+    }
+
+    // Add provided parameters to the URL which are not on the exclusion list
+    // Respect pre-existing params which already have values
+    if (queryParams) {
+      for (const [paramName, decodedValue] of Object.entries(queryParams)) {
+        if (decodedValue) {
```
Blackbaud-SteveBrush, line 234: Should `queryParams` respect `excludeParams`? Seems like it should...

libs/components/router/src/lib/modules/href/href.directive.ts
Blackbaud-TomRamsey: I removed null and updated the syntax to align with your suggestion

libs/components/config/src/lib/params.ts
Blackbaud-TomRamsey: The Params equivalent is:

```
// eslint-disable-next-line @typescript-eslint/no-explicit-any
queryParams?: Record<string, any>
```

I've updated the PR.

libs/components/config/src/lib/params.spec.ts
Blackbaud-TomRamsey: We need to clarify the functional requirements of the function.  Currently the behavior of getLinkUrl is inconsistent - it will sometimes encode parms and it will sometimes _not_ encode them.

1) You are correct that we are now treating params provided on the inbound url as something we need to encode, in the same way we are encoding parameters we add.
2) Switching + to %2B isn't double encoding in of itself, but I agree if the caller assumes we aren't doing any encoding this will cause double encoding.
3) Would you expect the encoding behavior to be _different_ depending on how the parameters are ingested?

For example, what should the following do:
```
fit('should validate query string encoding', () => {
    const params = new SkyAppRuntimeConfigParams('?a1=str+1', allowed);
    expect(
      params.getLinkUrl('https://mysite.com?a2=str+2', { a3: 'str+3' })
    ).toEqual('____');
  });
```

4) SkyHref will immediately set the response to href of the native element.  We need to encode it, or we will generate malformed urls
5) Currently encoding is handled by UrlEncoder defined in params.ts which wraps encodeURIComponent.  Do you want to follow [angular's practice](https://github.com/angular/angular/blob/cb31dbc75ca4141d61cec3ba6e60505198208a0a/packages/common/http/src/params.ts#L96-L111) where we encode using encodeURIComponent and then put back + and any other subset of characters?

libs/components/config/src/lib/params.ts
Blackbaud-TomRamsey: I'm very much on the fence for this.  I viewed queryParams provided in as the most specific request for arguments to be provided to a link, so I generally felt they should trump the default sky config params or those already on the url.

Do you have a scenario in mind where the general rules for exclusion should trump someone if they explicitly have asked for a query param to be added?  If they didn't want it on there, it feels like they wouldn't be providing it explicitly to this single link.  The url is a bit different, since its already been tampered with by the call chain (adding things like envid).

libs/components/config/src/lib/params.spec.ts
Blackbaud-SteveBrush: Params that already exist in the URL should not be encoded, so I would expect this test to pass as-is: https://github.com/blackbaud/skyux/pull/1462/files#diff-07e3b4a4c4dadbafa1bd357b6708e7c9636d97376d5e54d4d90188f545541a26R313

Params that "we own" (added from `SkyuxConfigParams` or from the `queryParams` input) should be encoded, so I would expect your test to output: 
```
https://mysite.com?a2=str+2&a3=str%2B3&a1=str+1
```

Granted, this is not great behavior (ideally all params would be encoded correctly) but our consumers have made some strange decisions and we don't want to break them in a minor version. If you'd like, you could put a `// TODO: encode all params in a future major version` comment above the encoding logic so we remember to come back and consider a breaking change in a major version.

libs/components/config/src/lib/params.ts
Blackbaud-SteveBrush: I don't have a specific scenario in mind, and you raise a good argument. We can keep your logic as-is. 👍

libs/components/config/src/lib/params.spec.ts
Blackbaud-TomRamsey: I refactored the logic to walk a fine line.  We will respect the current url parameters if they do not require modification, deferring to the caller that they are encoded as expected.  If we need to modify those parameters because they should either be removed or have their values overridden, we will fold the existing parameters into the unified set and encode them properly.

Blackbaud-SteveBrush (approved):

PR #1808, opened 2023-10-11
sky-text-editor allows developer to choose what options to show for 'Create link'

JosieePerezz:
Currently, if a user wants to add a link in the text editor, they can choose to open the link in the "Current window" or a "New window". However, in some cases we don't want to allow for opening in the "Current window", as it messes up certain workflows. 
This PR adds in a `openLinksInNewWindowOnly` property to the text editor component so that a developer who is using the text editor component can specific whether a link should **only** open up in a new window. This way, any user who creates a link will have that link always open up in a new window, and will not disturb the workflow.

I also made some changes in the playground components to see the text editor component changes. However, I can remove them since they're not really needed.


libs/components/text-editor/src/lib/modules/text-editor/text-editor.component.ts
```diff
@@ -231,6 +231,15 @@ export class SkyTextEditorComponent
     return this.#_toolbarActions;
   }
 
+  @Input()
+  public set openLinksInNewWindowOnly(value: boolean) {
```
Blackbaud-PaulCrowder: Instead of a boolean property that controls one option, use a property like `linkWindowOptions` that takes an array of allowed options. Make the default value include both options, and when the input is set to `undefined`, set it back to the default. Follow the pattern used by `toolbarActions` which takes an array of strings that match the union type of all the valid options.

String union type:
```
export type SkyTextEditorLinkWindowOptionType = 'new' | 'existing';
```

Default value const (in `text-editor.component.ts` defined above the class):
```
const LINK_WINDOW_OPTIONS_DEFAULT: SkyTextEditorLinkWindowOptionsType[] = ['new', 'existing'];
```

Component input:
```
@Input()
public set linkOptions(value: SkyTextEditorLinkWindowOptionType[] | undefined) {
  this.#_linkOptions = value ?? LINK_WINDOW_OPTIONS_DEFAULT;
}

public get linkOptions(): SkyTextEditorLinkWindowOptionsType[] {
  return this.#_linkOptions;
}

#_linkOptions = LINK_WINDOW_OPTIONS_DEFAULT;
```

apps/playground/src/app/components/text-editor/text-editor/text-editor.component.ts
```diff
@@ -54,9 +54,11 @@ export class TextEditorComponent implements OnInit {
     'alignment',
     'indentation',
     'undo-redo',
-    'link',
+    'link'
```
Blackbaud-PaulCrowder: I'm seeing this and other formatting changes unrelated to your feature which makes me think you may not have the Prettier extension installed in your IDE. Since the repo contains a Prettier config, it should be as simple as installing the extension and then saving the file to auto-format it using our rules.

libs/components/text-editor/src/lib/modules/text-editor/url-modal/text-editor-url-modal.component.html
```diff
@@ -49,6 +49,14 @@
             </option>
           </select>
         </sky-input-box>
+        <ng-template #onlyOpenLinksInNewWindow>
```
Blackbaud-PaulCrowder: Instead of displaying a readonly field when there is only one link window option specified, use a `div` with a class of `sky-font-deemphasized` with the information about what option was chosen for the user (.e.g. "This link will open in a new window.").

Blackbaud-PaulCrowder (commented):
Some general feedback about using this new feature: even if you generate the HTML from this component with only allowing links to open in a new window, you should still ensure that all `<a>` elements in the resulting HTML have the appropriate `target` attribute for the context where the HTML is rendered. For instance, if you know that the HTML is being displayed in an iframe, you should always add `target="_blank"` or `target="_top"` to each `<a>` element before it's rendered, since the HTML may have been created before this feature was added or through an API that bypasses this component completely. Modifying it just before rendering it will also allow you to reuse HTML in different contexts where different window target requirements are needed.

blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/1808/storybook/)

Component Storybooks:

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/1808/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/1808/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/1808/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/1808/playground/)

libs/components/text-editor/src/lib/modules/text-editor/url-modal/text-editor-url-modal.component.html
JosieePerezz: Done!

apps/playground/src/app/components/text-editor/text-editor/text-editor.component.ts
JosieePerezz: Installed Prettier and ran it!

libs/components/text-editor/src/lib/modules/text-editor/text-editor.component.ts
JosieePerezz: Done!

libs/components/text-editor/src/lib/modules/text-editor/text-editor.component.ts
```diff
@@ -311,6 +324,8 @@ export class SkyTextEditorComponent
   #_mergeFields: SkyTextEditorMergeField[] = [];
   #_menus = MENU_DEFAULTS;
   #_toolbarActions: SkyTextEditorToolbarActionType[] = TOOLBAR_ACTION_DEFAULTS;
+  #_linkWindowOptions: SkyTextEditorLinkWindowOptionsType[] =
```
Blackbaud-PaulCrowder: Since `LINK_WINDOW_OPTIONS_DEFAULT` already has a type specified, `#_linkWindowOptions` can infer its type from `LINK_WINDOW_OPTIONS_DEFAULT` and doesn't need it specified here.

```
#_linkWindowOptions = LINK_WINDOW_OPTIONS_DEFAULT;
```

libs/components/text-editor/src/lib/modules/text-editor/url-modal/text-editor-url-modal.component.html
```diff
@@ -49,6 +53,14 @@
...
+          <div class="sky-font-deemphasized">
```
Blackbaud-PaulCrowder: Since this text serves as part of the description of the URL input, the `input` element should have an `aria-describedby` attribute that points to this element when it is displayed. You'll need to use `SkyIdService` to generate an ID for this `div` and bind it to `[id]` on this element and `[attr.aria-describedby]` on the URL input when applicable.

libs/components/text-editor/src/lib/modules/text-editor/url-modal/text-editor-url-modal.component.ts
```diff
@@ -73,6 +76,13 @@ export class SkyTextEditorUrlModalComponent {
   readonly #modalInstance = inject(SkyModalInstance);
 
   constructor() {
+    if (this.#modalContext.linkWindowOptions?.length === 1) {
```
Blackbaud-PaulCrowder: When referencing a property that's multiple levels deep, it's good to assign it to a variable first. This is slightly more efficient and results in less code once it's minified for production. You can also combine the two `if` statements into one and check the first item instead of using `includes()`.

```
const linkWindowOptions = this.#modalContext.linkWindowOptions;

if (linkWindowOptions?.length === 1 && linkWindowOptions[0] === 'new') {
}
```

libs/components/text-editor/src/lib/modules/text-editor/url-modal/text-editor-url-modal.component.html
```diff
@@ -26,7 +26,11 @@
             />
           </sky-input-box>
         </div>
-        <sky-input-box stacked="true" class="sky-form-group">
+        <sky-input-box
+          *ngIf="!openLinksInNewWindowOnly; else onlyOpenLinksInNewWindow"
```
Blackbaud-PaulCrowder: Since we're allowing consumers to mix and match `new` and `existing` as options, we should also add logic for when consumers only specify `existing`. If a consumer only specifies `existing`, I think it's safe to hide the element that lets them select the option and not show any help text since that's the default behavior of a link with no `target` attribute.

libs/components/text-editor/src/lib/modules/text-editor/toolbar/text-editor-toolbar.component.ts
```diff
@@ -158,7 +162,10 @@ export class SkyTextEditorToolbarComponent implements OnInit {
     const inputModal = this.#modalService.open(SkyTextEditorUrlModalComponent, [
       {
         provide: SkyUrlModalContext,
-        useValue: { urlResult: currentLink },
+        useValue: {
```
Blackbaud-PaulCrowder: You didn't write this initially, but the usage of an object literal where a class instance is expected is an anti-pattern. Would you mind updating this code to use the class instead of an object literal?

```
provide: SkyUrlModalContext,
useFactory: () => {
  const context = new SkyUrlModalContext();
  context.urlResult = currentLink;
  context.linkWindowOptions = this.linkWindowOptions;

  return context;
}
```

libs/components/text-editor/src/lib/modules/text-editor/text-editor.component.ts
JosieePerezz: Done!

libs/components/text-editor/src/lib/modules/text-editor/url-modal/text-editor-url-modal.component.html
JosieePerezz: Check out latest commit, I _think_ I did it correctly in generating the id for the div and setting the [`attr.aria-describedby]` attribute to the input element **only** if `openLinksInNewWindowOnly` is true.

libs/components/text-editor/src/lib/modules/text-editor/url-modal/text-editor-url-modal.component.ts
JosieePerezz: Assigned the variable first and then added an `if-else` to set the correct `target` depending whether the option was 'new' or 'existing'.

libs/components/text-editor/src/lib/modules/text-editor/url-modal/text-editor-url-modal.component.html
JosieePerezz: Added logic in. Thanks for the suggestion.

libs/components/text-editor/src/lib/modules/text-editor/toolbar/text-editor-toolbar.component.ts
JosieePerezz: Done!

libs/components/text-editor/src/lib/modules/text-editor/url-modal/text-editor-url-modal.component.html
```diff
@@ -49,6 +54,14 @@
             </option>
           </select>
         </sky-input-box>
+        <ng-template #linksOpenInSpecificWindow>
```
Blackbaud-PaulCrowder: Instead of displaying this element, you can use the new `hintText` input on input box to display the persistent inline help text. This would go on the input box for the URL input:

```
<sky-input-box 
  stacked="true"
  [hintText]="allowAllOpenLinkOptions 
    ? undefined 
    : ('skyux_text_editor_url_modal_open_in_new_window_label' | skyLibResources)
>
```

libs/components/text-editor/src/lib/modules/text-editor/url-modal/text-editor-url-modal.component.html
```diff
@@ -10,7 +10,12 @@
     >
       <sky-tab tabHeading="Web page">
         <div class="sky-form-group sky-text-editor-url-modal-first-field">
-          <sky-input-box stacked="true">
+          <sky-input-box
+            stacked="true"
+            [hintText]="openLinksInNewWindowOnly ?
+              ('skyux_text_editor_url_modal_open_in_new_window_label'
+                | skyLibResources)
+             : null">
```
Blackbaud-PaulCrowder: The SKY UX team convention is to use `undefined` rather than `null`. We don't usually make a distinction between "variable/parameter hasn't been initialized/specified" and "this variable/parameter's value was set to a nullish value", so we try to stick with `undefined` when possible for consistency.

libs/components/text-editor/src/lib/modules/text-editor/text-editor.component.spec.ts
```diff
@@ -463,13 +508,78 @@ describe('Text editor', () => {
...
+      {
+        desc: 'new window',
+        windowOption: 'new',
+        urlStrings: [
+          'href="https://google.com',
+          'rel="noopener noreferrer"',
+          'target="_blank"',
+        ],
+        containsHintTextExpect: (
+          inputBoxText = document.querySelector('.sky-modal .sky-input-box')
```
Blackbaud-PaulCrowder: An input box can have more text content than just the hint text. You can select one level deeper with `.sky-input-box-hint-text`. I'd also recommend adding a specific CSS class to the input box you're selecting (like `sky-text-editor-url-input`) so that adding another input box before this one in the future doesn't break this test. Something like this:

```
inputBoxText = document.querySelector('.sky-modal .sky-text-editor-url-input .sky-input-box-hint-text');
```

libs/components/text-editor/src/lib/modules/text-editor/text-editor.component.spec.ts
```diff
@@ -463,13 +508,78 @@ describe('Text editor', () => {
...
+            ?.textContent
+        ) =>
+          expect(inputBoxText).toContain(
```
Blackbaud-PaulCrowder: Since this test imports `expect` from `@skyux-sdk/testing`, you can use the extension method `toHaveText()` which will trim whitespace and then do an exact match. This will ensure the test fails if we inadvertently add more text to this element later.

libs/components/text-editor/src/lib/modules/text-editor/text-editor.component.spec.ts
```diff
@@ -463,13 +508,78 @@ describe('Text editor', () => {
...
+          expect(inputBoxText).not.toContain(
+            'This link will open in a new window.'
+          ),
+      },
+    ].forEach((testArgs) => {
+      it(`should show correct link window options when ${testArgs.desc} option is specified`, fakeAsync(() => {
+        testComponent.value = '<p>Click here</p>';
+        if (testArgs.windowOption === 'new') {
+          testComponent.linkWindowOptions = ['new'];
+          console.log('josie in new');
```
Blackbaud-PaulCrowder: Did you mean to remove this after debugging?

libs/components/text-editor/src/lib/modules/text-editor/text-editor.component.spec.ts
```diff
@@ -463,13 +508,78 @@ describe('Text editor', () => {
...
+        } else {
+          testComponent.linkWindowOptions = ['existing'];
+          console.log('josie in existing');
```
Blackbaud-PaulCrowder: Same as above.

libs/components/text-editor/src/lib/modules/text-editor/url-modal/text-editor-url-modal.component.html
JosieePerezz: Done. Changed from null to undefined.

libs/components/text-editor/src/lib/modules/text-editor/text-editor.component.spec.ts
JosieePerezz: Done.

libs/components/text-editor/src/lib/modules/text-editor/text-editor.component.spec.ts
JosieePerezz: Changed to using `toHaveText()`.

libs/components/text-editor/src/lib/modules/text-editor/text-editor.component.spec.ts
JosieePerezz: Oops, removed.

libs/components/text-editor/src/lib/modules/text-editor/text-editor.component.spec.ts
JosieePerezz: Removed.

Blackbaud-PaulCrowder (approved):

PR #1813, opened 2023-10-12
fix(components/lookup): autocomplete position with nested overflow elements

johnhwhite:
[AB#2698938](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2698938)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/1813/storybook/)

Component Storybooks:

- [a11y-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/a11y-storybook/)
- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/colorpicker-storybook/)
- [core-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/core-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/errors-storybook/)
- [flyout-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/flyout-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/forms-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/lookup-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/modals-storybook/)
- [navbar-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/navbar-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/phone-field-storybook/)
- [popovers-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/popovers-storybook/)
- [progress-indicator-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/progress-indicator-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/split-view-storybook/)
- [storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/theme-storybook/)
- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/tiles-storybook/)
- [toast-storybook](https://blackbaud.github.io/skyux-pr-preview/1813/storybooks/toast-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/1813/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/1813/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/1813/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/1813/playground/)

libs/components/core/src/lib/modules/affix/affixer.ts
```diff
@@ -190,6 +191,15 @@ export class SkyAffixer {
...
 
+    // In Safari on iOS, fixed position elements may scroll with the viewport.
+    if (offset && this.#config.position !== 'absolute') {
+      const { top, left } = getViewportOffset();
+      offset.top = (offset.top || 0) + top;
+      offset.left = (offset.left || 0) + left;
+      offset.bottom = (offset.bottom || 0) + top;
+      offset.right = (offset.right || 0) + left;
+    }
+
```
johnhwhite: This is 1/3 of the fix for mobile Safari, which was the original issue where we tried using absolute position.

libs/components/core/src/lib/modules/overlay/overlay.component.scss
```diff
@@ -15,6 +15,8 @@
 
     &-fixed {
       position: fixed;
+      overflow: hidden;
+      height: calc(var(--skyux-vh, 1vh) * 100);
```
johnhwhite: This is 2/3 of the fix for mobile Safari, which was the original issue where we tried using absolute position.

libs/components/core/src/lib/modules/overlay/overlay.component.ts
```diff
@@ -281,4 +289,28 @@ export class SkyOverlayComponent implements OnInit, OnDestroy {
...
+          ) {
+            this.#renderer.setStyle(
+              this.overlayRef?.nativeElement,
+              'top',
+              `${this.#doc.defaultView.visualViewport.offsetTop}px`
+            );
+          }
+        });
+    }
+  }
```
johnhwhite: This is 3/3 of the fix for mobile Safari, which was the original issue where we tried using absolute position.

libs/sdk/e2e-schematics/src/cypress-preset.ts
```diff
@@ -20,5 +21,31 @@ export function skyE2ePreset(
...
+        if (browser.name === 'firefox' && browser.isHeadless) {
+          launchOptions.args.push('--width=1280');
+          // Includes the toolbar.
+          launchOptions.args.push('--height=1350');
+        }
+
+        return launchOptions;
+      });
+    },
+    ...options,
```
johnhwhite: By setting the window larger, the browser does not have to scroll during the screenshot, which was an issue for tests using fixed positioning.

libs/components/theme/src/lib/styles/_modals.scss
```diff
@@ -1,4 +1,5 @@
 .sky-modal-body-open {
+  height: 100%;
```
Blackbaud-SteveBrush, line 2: This needed for screenshot testing, or?

libs/components/core/src/lib/modules/affix/affixer.spec.ts
```diff
@@ -14,13 +15,65 @@ import { SkyAffixer } from './affixer';
...
+            'affix-layer-fixed': fixedLayer3,
+            'affix-layer-overflow': overflowLayer3
+          }"
+        >
+          <div class="affixer-container">
+            <div #baseElement class="affixer-base">base</div>
+          </div>
+          <div #affixedElement class="affixed">affixed</div>
+        </div>
+      </div>
```
johnhwhite, line 42: The new issue was an affixed country field within an `overflow: hidden` element. These test assert the behavior.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.scss
```diff
@@ -2,7 +2,7 @@
 @use 'libs/components/theme/src/lib/styles/variables' as *;
 
 .sky-autocomplete-results-container {
-  position: absolute;
+  position: fixed;
```
johnhwhite, line 5: The new issue was an affixed country field within an `overflow: hidden` element. The autocomplete changes revert back to using `position: fixed`, with the Safari on iOS issues being mitigated through other fixes.

libs/components/core/src/lib/modules/affix/affixer.ts
```diff
@@ -190,6 +191,15 @@ export class SkyAffixer {
   #affix(): void {
     const offset = this.#getOffset();
 
+    // In Safari on iOS, fixed position elements may scroll with the viewport.
+    if (offset && this.#config.position !== 'absolute') {
```
Blackbaud-SteveBrush: The following would make the comment "read easier" with the`if` statement. What do you think?
```
if (offset && this.#config.position === 'fixed') {
```

libs/components/core/src/lib/modules/overlay/overlay.component.ts
```diff
@@ -281,4 +289,28 @@ export class SkyOverlayComponent implements OnInit, OnDestroy {
       this.#routerSubscription = undefined;
     }
   }
+
+  #addScrollListener(): void {
+    if (
+      this.#doc.defaultView?.visualViewport &&
```
Blackbaud-SteveBrush: `this.#doc.defaultView.visualViewport` is referenced four times in this method. What if you set it in the constructor so it's object properties are only accessed once?
```
readonly #visualViewport: VisualViewport | null | undefined;

constructor() {
  this.#visualViewport = this.#doc.defaultView?.visualViewport;
}
```

libs/components/core/src/lib/modules/overlay/overlay.component.ts
```diff
@@ -281,4 +289,28 @@ export class SkyOverlayComponent implements OnInit, OnDestroy {
       this.#routerSubscription = undefined;
     }
   }
+
+  #addScrollListener(): void {
```
Blackbaud-SteveBrush: Do we need to account for the visual viewport "resize" event as well?

libs/components/core/src/lib/modules/overlay/overlay.component.ts
```diff
@@ -118,10 +121,14 @@ export class SkyOverlayComponent implements OnInit, OnDestroy {
 
   #coreAdapter: SkyCoreAdapterService;
 
+  #doc = inject(DOCUMENT);
```
Blackbaud-SteveBrush: Injections should be `readonly`. 👍 
```
readonly #document = inject(DOCUMENT);
```

libs/components/core/src/lib/modules/affix/dom-utils.ts
```diff
@@ -45,6 +45,13 @@ export function getElementOffset(
   };
 }
 
+export function getViewportOffset(): { top: number; left: number } {
```
Blackbaud-SteveBrush: Let's rename this to `getVisualViewportOffset` to make it more clear which viewport we're referencing.

libs/components/core/src/lib/modules/overlay/overlay.component.ts
```diff
@@ -118,10 +121,14 @@ export class SkyOverlayComponent implements OnInit, OnDestroy {
...
+
   #injector: Injector;
 
   #ngUnsubscribe = new Subject<void>();
 
+  #renderer = inject(RendererFactory2).createRenderer(null, null);
```
Blackbaud-SteveBrush: Same here.

libs/components/core/src/lib/modules/overlay/overlay.service.ts
```diff
@@ -29,8 +34,12 @@ export class SkyOverlayService {
 
   #componentFactoryResolver: ComponentFactoryResolver;
 
+  #doc = inject(DOCUMENT);
```
Blackbaud-SteveBrush: Needs `readonly`.

libs/components/core/src/lib/modules/overlay/overlay.service.ts
```diff
@@ -158,4 +173,24 @@ export class SkyOverlayService {
       }
     }
   }
+
+  #maintainVhProperty(): void {
```
Blackbaud-SteveBrush: Is this stuff better suited for the `SkyOverlayAdapterService` since it's DOM stuff?

libs/components/core/src/lib/modules/overlay/overlay.service.ts
```diff
@@ -158,4 +173,24 @@ export class SkyOverlayService {
...
+    // Maintain a --skyux-vh CSS property that is equal to 1% of the viewport height.
```
Blackbaud-SteveBrush: Could you also provide some additional context around why this property needs to be 1% of the viewport height? Also, which viewport (layout vs visual)?

Blackbaud-SteveBrush (commented):
A few observations.

Overall, this is a much-improved user experience; the delay for the affixing behavior is fine since it's a known issue of `visualViewport` event handling. 

1) For autocomplete, there appears to be extra space between the input and the dropdown; I'm unsure if this is related to your changes, but it's something I noticed.
autocomplete-1

2) When pinch-zooming, the dropdown seems to left-align itself to the visual viewport, instead of the input (which is the desired behavior).
autocomplete-2

3) When scrolling the document, the repositioning changes work great. If, however, I place my thumb on an "overlay" surface (such as the autocomplete's dropdown, or the modal body background) and drag to scroll, the dropdown moves in the opposite direction that is intended, leading to some interesting results (I'll DM you a screen capture):
autocomplete-3
autocomplete-4

libs/components/theme/src/lib/styles/_modals.scss
johnhwhite: This class is added to `<body>` when the modal is open, and it stabilizes the behavior of the overlay affix on mobile.

libs/components/core/src/lib/modules/overlay/overlay.service.ts
johnhwhite: I moved away from this solution.

libs/components/core/src/lib/modules/affix/dom-utils.ts
johnhwhite: I moved away from this solution.

libs/components/core/src/lib/modules/overlay/overlay.component.ts
johnhwhite: Resize doesn't affect the top left position.

libs/components/core/src/lib/modules/overlay/overlay.component.ts
johnhwhite: I moved away from this solution.

johnhwhite:
I think I've addressed a lot of the issues.

The autocomplete positioning "issue" is that the affix base element is the outer `<sky-autocomplete>` element, and there is additional margin being added to the form group within `<sky-input-box>`, which the `<sky-autocomplete>` wraps.

Moving the affix to the input element doesn't look quite right either:

[🦔 ag-grid](https://percy.io/82a32315/skyux-ag-grid-storybook-e2e/builds/30788206?utm_campaign=82a32315&utm_content=skyux-ag-grid-storybook-e2e&utm_source=github_status_public)
[🦔 integration](https://percy.io/82a32315/skyux-integration-e2e/builds/30788483?utm_campaign=82a32315&utm_content=skyux-integration-e2e&utm_source=github_status_public)
[🦔 lookup](https://percy.io/82a32315/skyux-lookup-storybook-e2e/builds/30788372?utm_campaign=82a32315&utm_content=skyux-lookup-storybook-e2e&utm_source=github_status_public)

I've reverted to the previous behavior. I think we need to address that separately.

libs/components/core/src/lib/modules/affix/dom-utils.ts
```diff
@@ -78,30 +125,49 @@ export function getOverflowParents(child: HTMLElement): HTMLElement[] {
...
+  if (!parent || offset.top < 0 || offset.left < 0) {
+    return false;
+  }
 
-  return !(
-    parentOffset.top > offset.top ||
-    parentOffset.right < offset.right ||
-    parentOffset.bottom < offset.bottom ||
-    parentOffset.left > offset.left
+  if (parent.nodeName.toLowerCase() === 'body') {
```
Blackbaud-SteveBrush: You have a similar check above with the parent element which seems a bit less fragile than this one. Possible?
```
if (parent === window.document.body) {
```

libs/components/core/src/lib/modules/affix/affix.service.ts
```diff
@@ -1,24 +1,43 @@
...
   readonly #renderer = inject(RendererFactory2).createRenderer(undefined, null);
 
   readonly #viewportRuler = inject(ViewportRuler);
 
   readonly #zone = inject(NgZone);
 
+  constructor() {
+    this.#layoutViewport = this.#dynamicComponentService.createComponent(
+      SkyAffixLayoutViewportComponent
+    );
```
Blackbaud-SteveBrush: Using the dynamic component service to create this seems a bit overkill when all you need is a simple element added to the page (change detection and all the other features that get baked-in with a component won't be needed). You could instead use the existing `Renderer2` utility to generate the element for you. Here's some mock code to better explain what I'm talking about.

```
constructor() {
  this.#layoutViewport = this.#createLayoutViewportShim();
}
```

```
/**
 * Blah, blah explain the shim and why it's useful.
 * @see https://github.com/WICG/visual-viewport/blob/gh-pages/examples/fixed-to-viewport.html
 */
#createLayoutViewportShim(): HTMLDivElement {
  const div = this.#renderer.createElement('div');

  const styles = new Map<string, string>([
    ['display', 'block'],
    // etc.
  ]);

  for (const [key, value] of styles.entries()) {
    this.#renderer.setStyle(div, key, value);
  }

  this.#renderer.addClass(div, 'sky-affix-layout-viewport-shim');
  this.#renderer.appendChild(document.body, div);

  return div;
}
```

I'm using "shim" based on that link you provided, since it's supposed to add functionality to the DOM that doesn't exist yet, and it also communicates that we're using a "hack" to get the layout viewport since its API is not super useful at the moment.

libs/components/core/src/lib/modules/affix/affixer.ts
```diff
@@ -190,6 +195,12 @@ export class SkyAffixer {
   #affix(): void {
     const offset = this.#getOffset();
 
+    const offsetParentRect = this.#getOffsetParentRect();
+    offset.top = (offset.top || 0) - offsetParentRect.top;
+    offset.left = (offset.left || 0) - offsetParentRect.left;
+    offset.bottom = (offset.bottom || 0) - offsetParentRect.top;
+    offset.right = (offset.right || 0) - offsetParentRect.left;
```
Blackbaud-SteveBrush: ☝️ Should probably use `??` instead of `||` so it doesn't needlessly fail against a zero value (and then set it to zero).

libs/components/core/src/lib/modules/affix/dom-utils.ts
```diff
@@ -45,21 +41,72 @@ export function getElementOffset(
...
+  const viewportRect = viewportRuler.getViewportRect();
+
+  const visibleRect = {
+    top: Math.max(elementRect.top, 0),
+    left: Math.max(elementRect.left, 0),
+    bottom: Math.min(elementRect.bottom, viewportRect.height),
+    right: Math.min(elementRect.right, viewportRect.width),
+  };
+
+  if (element.nodeName.toLowerCase() === 'body') {
```
Blackbaud-SteveBrush: Same as the comment below; possible to do an equality check with the `body` element directly?

libs/components/core/src/lib/modules/affix/dom-utils.ts
```diff
@@ -45,21 +41,72 @@ export function getElementOffset(
   };
 }
 
+/**
+ * Returns an AffixRect that represents the outer dimensions of a given element.
+ */
+export function getOuterRect(element: HTMLElement): Required<AffixRect> {
+  const rect = element.getBoundingClientRect();
+  const computedStyle = window.getComputedStyle(element, undefined);
+  const marginTop = parseInt(computedStyle.marginTop, 10);
```
Blackbaud-SteveBrush: Do you care about decimal values? If so, `parseFloat` might be more appropriate.

Blackbaud-SteveBrush (commented):
Functionally, this is a much better solution; I left a few questions around implementation details, and the #2 screenshot from [this comment](https://github.com/blackbaud/skyux/pull/1813#pullrequestreview-1677283565) is still happening. Is it possible to left-align the dropdown with the input even when the page is zoomed or scrolled to the right a bit?

libs/components/core/src/lib/modules/affix/affix.service.ts
Blackbaud-SteveBrush: Your changes look good for this. One question: do we need to create a shim element for every instance of `SkyAffixer` or just have one (created by the service) that's used by all of them?

Blackbaud-SteveBrush (approved):

blackbaud-sky-build-user:
Cherry-pick [failed](https://github.com/blackbaud/skyux/actions/runs/6645920879)! :x: Please resolve conflicts and create a new PR.

PR #1814, opened 2023-10-13
feat(components/router): add `queryParams` to `skyHref` directive (#1800)

Blackbaud-SteveBrush:
Original contribution: https://github.com/blackbaud/skyux/pull/1800
by @Blackbaud-TomRamsey

[AB#2704591](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2704591)

### DIFF
--- a/libs/components/config/src/lib/params.spec.ts
+++ b/libs/components/config/src/lib/params.spec.ts
@@ -278,6 +278,47 @@ describe('SkyAppRuntimeConfigParams', () => {
     expect(params.getAllKeys()).toEqual([]);
   });
 
+  it('should add provided params to a url link', () => {
+    const params: SkyAppRuntimeConfigParams = new SkyAppRuntimeConfigParams(
+      '',
+      allowed
+    );
+    expect(params.getLinkUrl('https://mysite.com', { q1: 5 })).toEqual(
+      'https://mysite.com?q1=5'
+    );
+  });
+
+  it('should ignore null, undefined, or empty provided params to a url link', () => {
+    const params: SkyAppRuntimeConfigParams = new SkyAppRuntimeConfigParams(
+      '',
+      allowed
+    );
+    expect(params.getLinkUrl('https://mysite.com', undefined)).toEqual(
+      'https://mysite.com'
+    );
+    expect(params.getLinkUrl('https://mysite.com', {})).toEqual(
+      'https://mysite.com'
+    );
+  });
+
+  it('should add provided params to a url link with a query string', () => {
+    const params: SkyAppRuntimeConfigParams = new SkyAppRuntimeConfigParams(
+      '',
+      {
+        q3: { value: 3 },
+        q4: { value: 4, excludeFromLinks: true },
+      }
+    );
+    expect(
+      params.getLinkUrl('https://mysite.com?q1=1&q2=2', {
+        q1: 5,
+        q5: null,
+        q6: undefined,
+        q7: '',
+      })
+    ).toEqual('https://mysite.com?q1=5&q2=2&q3=3');
+  });
+
   it("should exclude certain parameters from being added to a link's querystring", () => {
     const params = new SkyAppRuntimeConfigParams('https://mysite.com?a=1&b=2', {
       a: true,
@@ -296,10 +337,40 @@ describe('SkyAppRuntimeConfigParams', () => {
       },
     });
 
-    // Parameter 'f' should remain in the link because it already exists in the URL.
-    expect(params.getLinkUrl('https://mysite.com?c=3&f=6#foobar')).toEqual(
-      'https://mysite.com?c=3&f=6&a=1&b=2#foobar'
+    expect(params.getLinkUrl('https://mysite.com?c=3&f=6&g=9#foobar')).toEqual(
+      'https://mysite.com?c=3&g=9&a=1&b=2#foobar'
     );
+  });
+
+  it('should combine app config params, provided params, and url params', () => {
+    const params = new SkyAppRuntimeConfigParams(
+      'https://mysite.com?a=1&b=2&c=3',
+      {
+        b: {
+          excludeFromRequests: true,
+        },
+        c: {
+          value: 42,
+        },
+        d: {
+          value: 43,
+        },
+        m: {
+          excludeFromRequests: true,
+        },
+        n: {
+          excludeFromLinks: true,
+        },
+      }
+    );
+
+    expect(
+      params.getLinkUrl('https://mysite.com?a=10&c=13&f=6&m=14#foobar', {
+        a: 5,
+        m: 10,
+        n: 11,
+      })
+    ).toEqual('https://mysite.com?a=5&c=13&f=6&m=10&n=11&d=43#foobar');
   });
 
   it('should support query params with multiple values', () => {
@@ -320,5 +391,17 @@ describe('SkyAppRuntimeConfigParams', () => {
     expect(
       params.getLinkUrl('https://mysite.com?foobar=Robert+Hernandez')
     ).toEqual('https://mysite.com?foobar=Robert+Hernandez&a1=%20');
+  });
+
+  it('should validate query string encoding', () => {
+    const params = new SkyAppRuntimeConfigParams('?a1=str+1', {
+      a4: {
+        value: 'str+4',
+      },
+    });
+
+    expect(
+      params.getLinkUrl('https://mysite.com?a2=str+2', { a3: 'str+3' })
+    ).toEqual('https://mysite.com?a2=str+2&a3=str%2B3&a4=str%2B4');
   });
 });
--- a/libs/components/config/src/lib/params.spec.ts
+++ b/libs/components/config/src/lib/params.spec.ts
@@ -315,8 +315,9 @@ describe('SkyAppRuntimeConfigParams', () => {
         q5: null,
         q6: undefined,
         q7: '',
+        q8: false,
       })
-    ).toEqual('https://mysite.com?q1=5&q2=2&q3=3');
+    ).toEqual('https://mysite.com?q1=5&q2=2&q7=&q8=false&q3=3');
   });
 
   it("should exclude certain parameters from being added to a link's querystring", () => {
--- a/libs/components/config/src/lib/params.spec.ts
+++ b/libs/components/config/src/lib/params.spec.ts
@@ -236,7 +236,7 @@ describe('SkyAppRuntimeConfigParams', () => {
   it('should add provided params to a url link', () => {
     const params = new SkyAppRuntimeConfigParams('', allowed);
     expect(
-      params.getLinkUrl('https://mysite.com', { queryParams: { q1: 5 } })
+      params.getLinkUrl('https://mysite.com', { queryParams: { q1: '5' } })
     ).toEqual('https://mysite.com?q1=5');
   });
 
@@ -248,9 +248,9 @@ describe('SkyAppRuntimeConfigParams', () => {
     expect(
       params.getLinkUrl('https://mysite.com?q1=1&q2=2', {
         queryParams: {
-          q1: 5,
+          q1: '5',
           q7: '',
-          q8: false,
+          q8: 'false',
         },
       })
     ).toEqual('https://mysite.com?q1=5&q2=2&q7=&q8=false&q3=3');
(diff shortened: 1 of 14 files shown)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/1814/storybook/)

Component Storybooks:

- [a11y-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/a11y-storybook/)
- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/colorpicker-storybook/)
- [core-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/core-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/errors-storybook/)
- [flyout-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/flyout-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/forms-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/lookup-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/modals-storybook/)
- [navbar-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/navbar-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/phone-field-storybook/)
- [popovers-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/popovers-storybook/)
- [progress-indicator-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/progress-indicator-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/split-view-storybook/)
- [storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/theme-storybook/)
- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/tiles-storybook/)
- [toast-storybook](https://blackbaud.github.io/skyux-pr-preview/1814/storybooks/toast-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/1814/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/1814/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/1814/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/1814/playground/)

libs/components/config/src/lib/params.spec.ts
```diff
@@ -296,12 +338,42 @@ describe('SkyAppRuntimeConfigParams', () => {
       },
     });
 
-    // Parameter 'f' should remain in the link because it already exists in the URL.
-    expect(params.getLinkUrl('https://mysite.com?c=3&f=6#foobar')).toEqual(
```
Blackbaud-SteveBrush, line 300: This is the only existing test that was changed (deliberately).

libs/components/config/src/lib/params.ts
```diff
@@ -178,44 +178,103 @@ export class SkyAppRuntimeConfigParams {
...
+      const value = obj[key];
+
+      if (value === undefined || value === null) {
+        delete obj[key];
+      }
+    }
+
+    return new Map(
+      Object.entries(obj) as [string, string | number | boolean][]
+    );
```
Blackbaud-SteveBrush: I added this private method to remove `undefined` and `null` values from any URL parameters coming from our `skyHref` directive. Angular's `routerLink` directive accepts `any` and then handles the value internally to create the URL parameters. However, we use `HttpParams` (from `@angular/common/http`) to construct our URL params and it doesn't accept `undefined` or `null` values.

libs/components/router/src/lib/modules/href/href-query-params.ts
```diff
@@ -1 +0,0 @@
-export type SkyHrefQueryParams = { [k: string]: any };
```
Blackbaud-SteveBrush, line 1: This file is unused and never exported to the public API so I deleted it. I'm guessing it was created a while ago with the hopes of supporting query params, but it was dropped from the original pull request.

libs/components/router/src/lib/modules/href/href.directive.spec.ts
```diff
@@ -331,6 +331,34 @@ describe('SkyHref Directive', () => {
...
+
+    fixture.detectChanges();
+    tick();
+
+    const element = el.querySelector('a[data-sky-id="query-params"]');
+
+    expect(element?.getAttribute('href')).toEqual(
+      'https://success/example/page?a=foo&b=&c=0&d=false&envid=abc123#foobar'
+    );
+  }));
```
Blackbaud-SteveBrush, line 358: I added this unit test to make sure we're covering more use-cases.

libs/components/router/src/lib/modules/href/href.directive.ts
```diff
@@ -44,6 +45,21 @@ export class SkyHrefDirective {
...
   }
 
+  /**
+   * A collection of query URL parameters.
+   */
+  @Input()
+  public set queryParams(value: SkyHrefQueryParams | undefined) {
+    if (value !== this.#_queryParams) {
+      this.#_queryParams = value;
+      this.#applyChanges(this.#getChanges());
```
Blackbaud-SteveBrush, line 55: I added a setter to this input so that we could update the href when query parameters change.

libs/components/config/src/lib/params.ts
```diff
@@ -178,44 +178,103 @@ export class SkyAppRuntimeConfigParams {
...
+   * Removes undefined and null values from the given object and converts to a map.
+   */
+  #convertToCleanMap(
+    obj: Record<string, string | number | boolean | null | undefined>
+  ): Map<string, string | number | boolean> {
+    for (const key in obj) {
+      const value = obj[key];
+
+      if (value === undefined || value === null) {
+        delete obj[key];
```
Blackbaud-PaulCrowder: Deleting a property on this object is going to alter the original object which might have adverse downstream effects to the caller. If you really need to use a `Map`, I'd do that step first, then loop through the `Map` and delete entries that are `null` or `undefined`. Otherwise you could destructure it into a new object:

```
obj = {...obj};
```

libs/components/config/src/lib/params.ts
```diff
@@ -178,44 +178,103 @@ export class SkyAppRuntimeConfigParams {
...
+      }
+    }
+
+    return new Map(
```
Blackbaud-PaulCrowder: Is there still benefit to creating a `Map` here if you use the destructuring technique above?

libs/components/config/src/lib/params.spec.ts
```diff
@@ -278,6 +278,48 @@ describe('SkyAppRuntimeConfigParams', () => {
     expect(params.getAllKeys()).toEqual([]);
   });
 
+  it('should add provided params to a url link', () => {
+    const params: SkyAppRuntimeConfigParams = new SkyAppRuntimeConfigParams(
```
Blackbaud-PaulCrowder: Use type inference here and in the other tests with a similar structure.

libs/components/config/src/lib/params.ts
```diff
@@ -178,44 +178,103 @@ export class SkyAppRuntimeConfigParams {
 
   /**
    * Adds the current params to the supplied link URL.
+   * @param url The url to update.
+   * @param queryParams Optional query parameters to include in the constructed url.
    */
-  public getLinkUrl(url: string): string {
+  public getLinkUrl(
+    url: string,
+    queryParams?: Record<string, string | number | boolean | null | undefined>
```
Blackbaud-PaulCrowder: Should this use the `SkyHrefQueryParams` interface that was added in this PR?

libs/components/config/src/lib/params.ts
```diff
@@ -178,44 +178,103 @@ export class SkyAppRuntimeConfigParams {
...
 
-    return joinedParams.length === 0
-      ? url
-      : `${beforeFragment}${delimiter}${joinedParams}` +
-          (fragment ? `#${fragment}` : '');
+  /**
+   * Removes undefined and null values from the given object and converts to a map.
+   */
+  #convertToCleanMap(
+    obj: Record<string, string | number | boolean | null | undefined>
```
Blackbaud-PaulCrowder: Should this use the `SkyHrefQueryParams` interface that was added in this PR?

libs/components/config/src/lib/params.ts
Blackbaud-SteveBrush: The `SkyHrefQueryParams` type is added to `@skyux/router` but `@skyux/config` does not have a peer dependency on `@skyux/router` so it can't be used. I was trying to avoid adding too much to the API, but based on this feedback I created two new public types in `@skyux/config`:
```
// Similar to SkyHrefQueryParams but without undefined and null.
export type SkyConfigQueryParams = Record<string, string | number | boolean>;

export type SkyConfigGetLinkUrlOptions = {
  queryParams: SkyConfigQueryParams;
};
```
Having a `SkyConfigGetLinkUrlOptions` options object is probably the better route anyway since we might want to add `queryParamsHandling` etc. in the future.

libs/components/config/src/lib/params.ts
Blackbaud-SteveBrush: See above.

libs/components/config/src/lib/params.ts
Blackbaud-SteveBrush: Probably not; I just like the API of the `Map` better than the standard object. I'll revert it.

libs/components/config/src/lib/params.ts
Blackbaud-SteveBrush: Oh, good catch. I'll clone the object before messing with it.

libs/components/config/src/lib/params.spec.ts
Blackbaud-SteveBrush: Sounds good. Prettier did some adjusting so the file will appear like many changes were made but none are consequential; it is what it is.

libs/components/config/src/lib/params.ts
Blackbaud-SteveBrush: I converted this a standard function above the class and made its types more generic.

libs/components/router/src/lib/modules/href/href.directive.ts
```diff
@@ -11,14 +11,37 @@ import {
...
 type HrefChanges = { href: string; hidden: boolean };
 
+/**
+ * Transforms SkyHrefQueryParams into SkyConfigQueryParams.
+ */
+function toConfigQueryParams(params: SkyHrefQueryParams): SkyConfigQueryParams {
+  // Clone the object to avoid modifying the original.
+  const clone = { ...params };
+
+  for (const key in clone) {
```
Blackbaud-TomRamsey: I would recommend against filtering the query params here because it will create forked behavior based on who calls getLinkUrl.

Blackbaud-SteveBrush: Good point. Reverting.

libs/components/config/src/lib/params.ts
Blackbaud-SteveBrush: Ended up removing this function altogether.

Blackbaud-PaulCrowder (approved):

PR #1912, opened 2023-12-11
refactor(components/forms): create `sky-errors` component to reuse with other inputs

johnhwhite:
[AB#2697800](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2697800)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/1912/storybook/)

Component Storybooks:

- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/colorpicker-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/errors-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/forms-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/lookup-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/phone-field-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/1912/storybooks/theme-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/1912/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/1912/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/1912/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/1912/playground/)

libs/components/forms/src/index.ts
```diff
@@ -35,6 +37,7 @@ export { SkyCharacterCounterInputDirective as λ2 } from './lib/modules/characte
 export { SkyCheckboxLabelComponent as λ4 } from './lib/modules/checkbox/checkbox-label.component';
 export { SkyCheckboxRequiredValidatorDirective as λ5 } from './lib/modules/checkbox/checkbox-required-validator.directive';
 export { SkyCheckboxComponent as λ3 } from './lib/modules/checkbox/checkbox.component';
+export { SkyErrorComponent as λ21 } from './lib/modules/error/error.component';
```
Blackbaud-SteveBrush: We already export a `SkyErrorComponent` from the `@skyux/errors` package. Perhaps these should be named `SkyFormErrorComponent` and `SkyFormErrorsComponent` instead

libs/components/forms/src/assets/locales/resources_en_US.json
```diff
@@ -7,6 +7,42 @@
     "_description": "Screen reader only label for character count over limit symbol",
     "message": "You are over the character limit."
   },
+  "skyux_error_character_count": {
```
Blackbaud-SteveBrush: See my comment below. Might be good to prefix these with `skyux_form_error_`.

libs/components/forms/src/lib/modules/error/error.component.ts
```diff
@@ -5,16 +5,18 @@ import { SkyStatusIndicatorModule } from '@skyux/indicators';
...
-  template: `<sky-status-indicator
-    class="sky-error-indicator sky-input-box-error"
-    descriptionType="error"
-    indicatorType="danger"
-  >
-    <ng-content />
-  </sky-status-indicator>`,
+  template: `
+    <sky-status-indicator
+      class="sky-error-indicator sky-error"
```
Blackbaud-SteveBrush: `sky-form-error-indicator sky-form-error`

libs/components/forms/src/lib/modules/error/errors.component.html
Blackbaud-SteveBrush, line 1: Files should be renamed to `form-errors/form-errors.component.html`

libs/components/forms/src/lib/modules/error/errors.component.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * @internal
+ */
+@Component({
+  selector: 'sky-errors',
+  standalone: true,
+  imports: [
+    CommonModule,
+    SkyI18nModule,
```
Blackbaud-SteveBrush: The `SkyI18nModule` should be exported by `SkyFormsResourcesModule` so is redundant.

libs/components/forms/testing/src/input-box/input-box-harness.ts
```diff
@@ -71,8 +71,7 @@ export class SkyInputBoxHarness extends SkyComponentHarness {
   public async getCustomErrors(): Promise<SkyStatusIndicatorHarness[]> {
     const errors = await this.locatorForAll(
       new HarnessPredicate(SkyStatusIndicatorHarness, {
-        selector:
-          'sky-status-indicator:not(sky-input-box-errors sky-status-indicator)',
+        selector: 'sky-status-indicator:not(sky-errors sky-status-indicator)',
```
Blackbaud-SteveBrush: The story mentions creating a new test harness for the `sky-form-errors` component to make it easier for consuming components to test. Probably would be a good idea to include in this pull request to solidify your API.

libs/components/forms/src/lib/modules/input-box/input-box.component.html
```diff
@@ -170,16 +170,10 @@
...
-        controlDir &&
-        controlDir.invalid &&
-        (controlDir.dirty || controlDir.touched) &&
-        controlDir.errors
-      "
-      [errors]="controlDir.errors"
+    <sky-errors
+      [errors]="controlDir?.errors"
       [labelText]="labelText"
+      [showErrors]="controlDir?.touched || controlDir?.dirty"
```
Blackbaud-SteveBrush: What about `controlDir.invalid`?

libs/components/forms/src/lib/modules/error/errors.component.ts
```diff
@@ -0,0 +1,56 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class SkyErrorsComponent {
+  @HostBinding('attr.aria-live')
+  public readonly ariaLive = 'polite';
+
+  /**
+   * The validation errors from the form control.
+   */
+  @Input({ transform: (value: unknown) => value || undefined })
```
Blackbaud-SteveBrush: Why is this `transform` needed?

libs/components/forms/src/lib/modules/error/errors.component.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Input label text to display in the error messages.
+   */
+  @Input()
+  public labelText: string | undefined;
+
+  /**
+   * Indicates whether to show error messages, which might only be true if the
+   * form control is touched or dirty.
```
Blackbaud-SteveBrush: I would also indicate in the comment that this input is required to make `aria-live` work (since the `sky-form-errors` component needs to exist on the page always and `*ngIf` would disrupt the announcer.

libs/components/forms/src/lib/modules/error/errors.component.ts
```diff
@@ -0,0 +1,56 @@
...
+    SkyErrorComponent,
+    SkyFormsResourcesModule,
+  ],
+  templateUrl: './errors.component.html',
+  styleUrls: ['./errors.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class SkyErrorsComponent {
+  @HostBinding('attr.aria-live')
+  public readonly ariaLive = 'polite';
```
Blackbaud-SteveBrush: Take a look at this page: https://hidde.blog/how-to-make-inline-error-messages-accessible/#heading-3 and this code example: https://codepen.io/hidde/pen/eRGwer.

They're using `aria-describedby` to link the error region with the input field. Do we also need to do that?

Blackbaud-SteveBrush: I was having difficulty getting the screen reader to read out the errors on the input box code example, here: https://blackbaud.github.io/skyux-pr-preview/1912/code-examples/#/forms/input-box/basic

Blackbaud-SteveBrush: Does `protected` work here?

libs/components/forms/src/lib/modules/error/errors.component.ts
johnhwhite: `null` threw an error.

libs/components/forms/src/lib/modules/input-box/input-box.component.html
johnhwhite: The previous logic for input box used `touched` or `dirty` to prevent errors on fields before they have been acted on.

libs/components/forms/src/lib/modules/error/errors.component.ts
johnhwhite: Yes, good call, updated.

libs/components/forms/src/lib/modules/error/errors.component.ts
johnhwhite: With this approach, the `<sky-form-errors>` element is present regardless, and the content is wrapped in `*ngIf` that uses this field, which is a departure from the input-box implementation where the whole error element was wrapped in an `*ngIf`.

libs/components/forms/src/lib/modules/error/errors.component.ts
johnhwhite: Yes, good call, updated.

libs/components/forms/src/assets/locales/resources_en_US.json
```diff
@@ -7,6 +7,42 @@
...
+    "_description": "Error message for a field with an invalid email value",
+    "message": "Enter an email address with a valid format."
+  },
+  "skyux_form_error_maxlength": {
+    "_description": "Error message for a field with a value that is too long",
+    "message": "Limit {0} to {1} character(s)."
+  },
+  "skyux_form_error_minlength": {
+    "_description": "Error message for a field with a value that is not long enough",
+    "message": "{0} must be at least {1} character(s)."
```
Blackbaud-SandhyaRajasabeson, line 28: are screenreaders able to parse `character(s)` in a non awkward way?

libs/components/forms/src/lib/modules/form-error/form-errors.component.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  let component: SkyFormErrorsComponent;
+
+  beforeEach(async () => {
+    fixture = TestBed.createComponent(SkyFormErrorsComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+    await fixture.whenStable();
+  });
+
+  it('should be toggle announcements', async () => {
```
Blackbaud-SandhyaRajasabeson: nit: grammar

libs/components/forms/src/lib/modules/form-error/form-errors.component.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+    expect(component).toBeTruthy();
+    expect(fixture.nativeElement.outerHTML).toContain('aria-live="polite"');
+    expect(fixture.nativeElement.outerHTML).toContain('aria-relevant="all"');
+    component.announceErrors = false;
+    fixture.detectChanges();
```
Blackbaud-SandhyaRajasabeson: another nit: can we add spaces so the different steps in the unit test are more separated, it's a bit hard to follow rn

libs/components/forms/src/index.ts
johnhwhite: Good call. Updated.

libs/components/forms/src/assets/locales/resources_en_US.json
johnhwhite: Updated.

libs/components/forms/src/lib/modules/error/error.component.ts
johnhwhite: Updated.

libs/components/forms/src/lib/modules/error/errors.component.html
johnhwhite: Good call. Updated.

libs/components/forms/src/assets/locales/resources_en_US.json
johnhwhite: Our resources don't currently have a better way to handle pluralization. This was moved over from [skyux_input_box_error_maxlength](https://github.com/blackbaud/skyux/blob/main/libs/components/forms/src/assets/locales/resources_en_US.json#L128).

libs/components/forms/src/lib/modules/form-error/form-errors.component.spec.ts
johnhwhite: Good call. Updated.

libs/components/forms/src/lib/modules/form-error/form-errors.component.spec.ts
johnhwhite: Good call. Updated.

libs/components/forms/src/assets/locales/resources_en_US.json
Blackbaud-SandhyaRajasabeson: goootcha, ima throw a flex topic on the board just for myself but resolve this thread

libs/components/forms/src/lib/modules/error/errors.component.ts
johnhwhite: Switched to `role="alert"`, which should make the errors assertive.

johnhwhite: I've added `errorId` to the `aria-describedby` — we also apply it to `aria-errormessage`, which seems more appropriate but also does not seem to have much browser support.

libs/components/forms/testing/src/input-box/input-box-harness.ts
johnhwhite: We moved this piece to AB#2766321.

libs/components/forms/src/index.ts
```diff
@@ -3,6 +3,8 @@ export { SkyCharacterCounterModule } from './lib/modules/character-counter/chara
 export { SkyCheckboxChange } from './lib/modules/checkbox/checkbox-change';
 export { SkyCheckboxModule } from './lib/modules/checkbox/checkbox.module';
 
+export { SkyFormErrorsComponent } from './lib/modules/form-error/form-errors.component';
```
Blackbaud-SteveBrush: This should be a lambda export. Let's create a `SkyFormErrorsModule` that exports the `SkyFormErrorsComponent` so the API is consistent with all other components (until we figure out the best way to roll out standalone components to our consumers).
```
/**
 * @internal
 */
@NgModule({
  imports: [SkyFormErrorsComponent],
  exports: [SkyFormErrorsComponent]
})
export class SkyFormErrorsModule {}
```

libs/components/forms/src/lib/modules/form-error/form-error.component.ts
```diff
@@ -0,0 +1,32 @@
+import { Component, HostBinding } from '@angular/core';
+import { SkyStatusIndicatorModule } from '@skyux/indicators';
+
+/**
+ * @internal
+ */
+@Component({
+  selector: 'sky-form-error',
```
Blackbaud-SteveBrush, line 8: Mind slapping the `ChangeDetectionStrategy.OnPush` on here?

libs/components/forms/src/lib/modules/form-error/form-error.component.ts
```diff
@@ -0,0 +1,32 @@
...
+      :host {
+        display: block;
+        margin-top: var(--sky-margin-inline-xs);
+      }
+    `,
+  ],
+})
+export class SkyFormErrorComponent {
+  @HostBinding('class')
+  protected cssClass = 'sky-form-error-indicator';
```
Blackbaud-SteveBrush: We're using the same class for the host element and the `sky-status-indicator` component (above). Needed?

libs/components/forms/src/lib/modules/form-error/form-error.component.ts
```diff
@@ -0,0 +1,32 @@
...
+      indicatorType="danger"
+    >
+      <ng-content />
+    </sky-status-indicator>
+  `,
+  styles: [
+    `
+      :host {
+        display: block;
+        margin-top: var(--sky-margin-inline-xs);
```
Blackbaud-SteveBrush: This CSS property was designed to be used for "inline" elements. I'd use `--sky-margin-stacked-xs` like we do for the hint text: https://github.com/blackbaud/skyux/blob/main/libs/components/forms/src/lib/modules/input-box/input-box.component.scss#L123

libs/components/forms/src/lib/modules/form-error/form-errors.component.ts
```diff
@@ -0,0 +1,68 @@
...
+  @HostBinding('attr.aria-relevant')
+  protected ariaRelevant: 'all' | undefined = 'all';
+
+  @HostBinding('attr.role')
+  protected role: 'alert' | undefined = 'alert';
+
+  /**
+   * The validation errors from the form control.
+   */
+  @Input({ transform: (value: unknown) => value || undefined })
```
Blackbaud-SteveBrush: Instead of using a transform, could you expand the type to `errors: ValidationErrors | null | undefined`? I feel like that would be a cleaner API and wouldn't require much explanation.

libs/components/forms/src/lib/modules/form-error/form-errors.component.ts
```diff
@@ -0,0 +1,68 @@
...
+  templateUrl: './form-errors.component.html',
+  styleUrls: ['./form-errors.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class SkyFormErrorsComponent {
+  @HostBinding('attr.aria-relevant')
+  protected ariaRelevant: 'all' | undefined = 'all';
+
+  @HostBinding('attr.role')
+  protected role: 'alert' | undefined = 'alert';
```
Blackbaud-SteveBrush: We talked offline, but we can go with `aria-live="assertive" aria-atomic="true"` (since `role="alert"` has problems with `aria-describedby` on Safari).
```
@HostBinding('attr.aria-atomic')
protected ariaAtomic = true;

@HostBinding('attr.aria-live')
protected ariaLive = 'assertive';
```

libs/components/forms/src/lib/modules/input-box/input-box.component.html
```diff
@@ -170,16 +170,11 @@
...
-        controlDir.errors
-      "
-      [errors]="controlDir.errors"
+    <sky-form-errors
+      [errors]="controlDir?.errors"
       [labelText]="labelText"
+      [showErrors]="controlDir?.touched || controlDir?.dirty"
+      announceErrors="false"
     />
     <ng-content select=".sky-error-label,.sky-error-indicator" />
```
Blackbaud-SteveBrush, line 178: Instead of creating more complicating logic around when to set `announceErrors` to `true`/`false`, could you put the `<ng-content select=".sky-error-label,.sky-error-indicator" />` element in the `sky-form-errors` component template so that `sky-form-errors` _always_ announces the messages? This will clean up the `sky-form-errors` API greatly, and reduce confusion when implementing it.

**form-errors.component.html**
```
<ng-container *ngIf="labelText && showErrors && errors">
...
</ng-content>

<ng-container *ngIf="showErrors">
  <ng-content select=".sky-error-label,.sky-error-indicator" />
</ng-container>
```

libs/components/forms/src/lib/modules/input-box/input-box.component.ts
```diff
@@ -241,6 +241,13 @@ export class SkyInputBoxComponent
         this.hintText,
       );
 
+      this.#adapterService.updateDescribedBy(
+        this.inputRef,
+        this.errorId,
+        'errorId', // Passing a string here to force the aria-describedby attribute to include the errorId.
```
Blackbaud-SteveBrush: I understand why you had to add this string, but I'm curious if we need to use it in the way it's intended: the third argument is basically a fail-safe to make sure an empty container doesn't get added to `aria-describedby`. In the case of errors, it's possible that a control could have errors, but the `sky-form-errors` element is empty, so we wouldn't want to add the ID to the `aria-describedby` in that case.

Blackbaud-SteveBrush: We talked offline, but it might be worth wrapping both the errors and the hint text elements in a single DIV with an ID. That way, we're only worrying about one `aria-describedby` element instead of two.
```
<input [aria-describedby]="fieldDescriptors.id" />

<div skyId #fieldDescriptors>
  <sky-form-errors />
  <hint-text />
</div>
```

Blackbaud-SteveBrush: Also, I noticed that this component's `hintText` setter [emits to the `ariaDescribedBy` observable](https://github.com/blackbaud/skyux/blob/main/libs/components/forms/src/lib/modules/input-box/input-box.component.ts#L140) (which is passed through to the [`SkyInputBoxHostService.ariaDescribedBy`](https://github.com/blackbaud/skyux/blob/main/libs/components/forms/src/lib/modules/input-box/input-box-host.service.ts#L23-L25) and [subscribed to by the country field](https://github.com/blackbaud/skyux/blob/main/libs/components/lookup/src/lib/modules/country-field/country-field.component.ts#L552). 

We need to figure out an elegant way to emit the `ariaDescribedBy` observable when errors change as well.

apps/playground/src/app/components/forms/input-box/input-box.component.html
```diff
@@ -672,11 +671,18 @@
...
           name="address"
           autocomplete="street-address"
+          required
         >
 Value</textarea
         >
       </sky-input-box>
     </div>
+
+    <sky-form-errors
```
Blackbaud-SteveBrush, line 681: We talked offline about putting a `hintText` example in the `apps/code-examples` SPA to make sure the hint text and errors work well together.

Blackbaud-SteveBrush: Also, might be good to create a visual test of both hint text and errors on the same input box to ensure the margins never change between PRs.

libs/components/forms/src/lib/modules/error/errors.component.ts
johnhwhite: Added `| null` to the type, removed the transform.

libs/components/forms/src/lib/modules/error/errors.component.ts
johnhwhite: Like we talked about, we're going to hold off on `aria-describedby` for now.

libs/components/forms/src/index.ts
johnhwhite: Good call, updated.

libs/components/forms/src/lib/modules/form-error/form-error.component.ts
johnhwhite: Good call, updated.

libs/components/forms/src/lib/modules/form-error/form-error.component.ts
johnhwhite: Cleaned this up.

libs/components/forms/src/lib/modules/form-error/form-error.component.ts
johnhwhite: Good call, updated.

libs/components/forms/src/lib/modules/form-error/form-errors.component.ts
johnhwhite: Updated the type.

libs/components/forms/src/lib/modules/form-error/form-errors.component.ts
johnhwhite: 👍

libs/components/forms/src/lib/modules/input-box/input-box.component.html
johnhwhite: Added `<ng-content />` and now always announcing errors.

libs/components/forms/src/lib/modules/input-box/input-box.component.ts
johnhwhite: Like we talked about, not touching `aria-describedby` at this point.

apps/playground/src/app/components/forms/input-box/input-box.component.html
johnhwhite: Added the example.

That visual test already exists.

Blackbaud-SteveBrush (approved):

PR #1981, opened 2024-02-01
fix(components/forms): refactor form errors

Blackbaud-ErikaMcVey:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/1981/storybook/)

Component Storybooks:

- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/colorpicker-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/errors-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/forms-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/lookup-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/phone-field-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/1981/storybooks/theme-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/1981/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/1981/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/1981/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/1981/playground/)

libs/components/forms/testing/src/input-box/input-box-harness.ts
```diff
@@ -9,6 +9,7 @@ import { SkyStatusIndicatorHarness } from '@skyux/indicators/testing';
 import { SkyPopoverHarness } from '@skyux/popovers/testing';
 
 import { SkyCharacterCounterIndicatorHarness } from '../character-counter/character-counter-indicator-harness';
+import { SkyFormErrorsHarness } from '../public-api';
```
Blackbaud-SteveBrush: It's better to avoid importing from the "public API" in an internal file (only the consumer should interact with that file). Also, importing from "barrel files" can [cause unforeseen problems](https://dev.to/tassiofront/barrel-files-and-why-you-should-stop-using-them-now-bc4). let's put the full path instead.

libs/components/forms/testing/src/form-error/form-errors-harness.ts
```diff
@@ -34,85 +20,28 @@ export class SkyFormErrorsHarness extends SkyComponentHarness {
...
-  /*
-   * Gets if the URL error has triggered.
+  /**
+   * Whether an error with the given name has fired.
    */
-  public async hasUrlError(): Promise<boolean> {
-    return (await this.#getFormErrorsClasses()).includes('sky-form-error-url');
+  public async hasError(errorName: string): Promise<boolean> {
+    const formErrors = await this.getFormErrors();
+    return !!formErrors.find((error) => {
```
Blackbaud-SteveBrush: Instead of `!!formErrors.find`, you can do `formErrors.some`, which returns a boolean.

libs/components/forms/testing/src/form-error/form-errors-harness.ts
```diff
@@ -34,85 +20,28 @@ export class SkyFormErrorsHarness extends SkyComponentHarness {
...
-   * Gets if the date error has triggered.
-   */
-  public async hasDateError(): Promise<boolean> {
-    return (await this.#getFormErrorsClasses()).includes('sky-form-error-date');
-  }
-
-  /*
-   * Gets if the email error has triggered.
+  /**
+   * Gets an array of all errors fired.
```
Blackbaud-SteveBrush: I'd use the word "list" instead of "array" in case the inner workings change.

libs/components/forms/src/lib/modules/input-box/input-box.module.ts
```diff
@@ -14,11 +15,16 @@ import { SkyInputBoxComponent } from './input-box.component';
   imports: [
     CommonModule,
     SkyCharacterCounterModule,
+    SkyFormErrorComponent,
```
Blackbaud-SteveBrush: Let's _import_ the `SkyFormErrorsModule`, but _export_ the `SkyFormErrorComponent`, to follow existing patterns.

libs/components/forms/src/lib/modules/input-box/input-box.component.ts
```diff
@@ -47,6 +49,10 @@ import { SkyInputBoxPopulateArgs } from './input-box-populate-args';
     SkyContentInfoProvider,
     SkyInputBoxAdapterService,
     SkyInputBoxHostService,
+    {
+      provide: FORM_ERRORS,
+      useValue: true,
```
Blackbaud-SteveBrush, line 54: The type used to declare `FORM_ERRORS` is `InjectionToken<string>` but a boolean is being used here. If this is supposed to be a boolean, maybe it should be named `InjectionToken<boolean>('SKY_FORM_ERRORS_ENABLED')` instead?

libs/components/forms/src/lib/modules/input-box/input-box.component.html
```diff
@@ -176,6 +176,7 @@
     [showErrors]="controlDir?.touched || controlDir?.dirty"
   >
     <ng-content select=".sky-error-label,.sky-error-indicator" />
+    <ng-content select="sky-form-error" />
```
Blackbaud-SteveBrush: Does `<ng-content select="sky-form-error,.sky-error-label,.sky-error-indicator" />` work?

libs/components/forms/src/lib/modules/form-error/form-errors.component.html
```diff
@@ -15,41 +15,31 @@
...
+  <sky-form-error *ngIf="errors['skyPhoneField']" errorName="phone">
     {{ 'skyux_form_error_phone' | skyLibResources }}
   </sky-form-error>
 
-  <sky-form-error *ngIf="errors['skyTime']" class="sky-form-error-time">
+  <sky-form-error *ngIf="errors['skyTime']" errorName="time">
     {{ 'skyux_form_error_time' | skyLibResources }}
   </sky-form-error>
 
-  <sky-form-error *ngIf="errors['skyUrl']" class="sky-form-error-url">
```
Blackbaud-SteveBrush, line 52: I'm bringing this question up since we're making `sky-form-error` part of the public API: Would it be better to create a new input named `errorText` that only accepts a string, instead of using `<ng-content />` to prevent the consumer from putting whatever they want in the error? Similar to how we're handling the label text for other easy mode components:
```
<sky-form-error
  errorName="foobar"
  errorText="Something bad happened."
/>
```

libs/components/forms/src/lib/modules/form-error/form-errors-token.ts
```diff
@@ -0,0 +1,3 @@
+import { InjectionToken } from '@angular/core';
+
+export const FORM_ERRORS = new InjectionToken<string>('FORM_ERRORS');
```
Blackbaud-SteveBrush: 1. This should be prefixed with `SKY_` so that it doesn't collide with any other token.
2. Wouldn't hurt to put `@internal` above it so future developers don't think this is part of the public API.
3. (See other feedback about `FORM_ERRORS`, below.)

libs/components/forms/src/lib/modules/form-error/form-error.component.ts
```diff
@@ -28,6 +39,26 @@ import { SkyStatusIndicatorModule } from '@skyux/indicators';
...
-  protected readonly cssClass = 'sky-form-error-indicator';
+  protected cssClass = '';
+
+  #_errorName = '';
+
+  protected readonly formErrors = inject(FORM_ERRORS, { optional: true });
+  readonly #changeDetector = inject(ChangeDetectorRef);
+
+  #updateClasses(): void {
+    this.cssClass = `sky-form-error-${this.errorName} sky-form-error-indicator`;
```
Blackbaud-SteveBrush: Are you adding a CSS class purely so you can look it up for the test harness? If so, you might run into issues if the error name is not a safe value to use for a CSS name. For example, the following is a valid error type:

```
<sky-form-error
  *ngIf="favoriteColor.errors?.['Opaque Color']"
  errorName="Opaque Color"
>
```
Instead, we could use a `data-*` attribute to save the error name as-is so we don't have to worry about malformed CSS classes.

libs/components/forms/src/lib/modules/form-error/form-error.component.ts
```diff
@@ -28,6 +39,26 @@ import { SkyStatusIndicatorModule } from '@skyux/indicators';
...
+    return this.#_errorName;
+  }
+
   @HostBinding('class')
-  protected readonly cssClass = 'sky-form-error-indicator';
+  protected cssClass = '';
+
+  #_errorName = '';
+
+  protected readonly formErrors = inject(FORM_ERRORS, { optional: true });
```
Blackbaud-SteveBrush: What do you think about throwing an error in the constructor if this is undefined so that consumers aren't confused when the error doesn't display in an unsupported context? Something like...
```
throw new Error('The `sky-form-error` component is not supported in the provided context.')
```

libs/components/forms/src/lib/modules/form-error/form-error.component.ts
```diff
@@ -28,6 +39,26 @@ import { SkyStatusIndicatorModule } from '@skyux/indicators';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class SkyFormErrorComponent {
+  @Input()
```
Blackbaud-SteveBrush: This is required, yes? If so, we could use `@Input({ required: true })` so that the consumer gets a build error if the input is not set in their template.

Blackbaud-SteveBrush: Also, needs a JSDocs comment.

libs/components/forms/src/lib/modules/form-error/form-error.component.ts
```diff
@@ -1,15 +1,26 @@
...
+  HostBinding,
+  Input,
+  inject,
+} from '@angular/core';
 import { SkyStatusIndicatorModule } from '@skyux/indicators';
 
+import { FORM_ERRORS } from './form-errors-token';
+
 /**
  * @internal
```
Blackbaud-SteveBrush: This isn't internal anymore, correct? We'll need to remove the `@internal` flag and add JSDocs to the class and to its inputs.

libs/components/forms/src/index.ts
```diff
@@ -29,6 +29,9 @@ export { SkySelectionBoxGridAlignItemsType } from './lib/modules/selection-box/t
 export { SkyToggleSwitchModule } from './lib/modules/toggle-switch/toggle-switch.module';
 export { SkyToggleSwitchChange } from './lib/modules/toggle-switch/types/toggle-switch-change';
 
+export { SkyFormErrorComponent } from './lib/modules/form-error/form-error.component';
```
Blackbaud-SteveBrush: This component needs to be exported with a "lambda" name so that consumers can't pull it into their stuff directly. I would export the `SkyFormErrorsModule` if you have to export anything.

```
@NgModule({
  imports: [SkyFormErrorComponent, SkyFormErrorsComponent],
  exports: [SkyFormErrorComponent, SkyFormErrorsComponent]
})
export class SkyFormErrorsModule {}
```

apps/code-examples/src/app/code-examples/forms/input-box/basic/demo.component.spec.ts
```diff
@@ -71,7 +88,7 @@ describe('Basic input box demo', () => {
   });
 
   describe('favorite color field', () => {
-    it('should not allow bird to be selected', async () => {
+    it('should not allow blur to be selected', async () => {
```
Blackbaud-SteveBrush: Can we name the non-color value something more self-explanatory (especially since this is a code example). Something like "invalid color". Or, just go back to "bird" since "blur" has a specific meaning when developing forms and could be confusing.

libs/components/forms/testing/src/form-error/form-error-harness.ts
```diff
@@ -19,15 +19,10 @@ export class SkyFormErrorHarness extends SkyComponentHarness {
     return SkyFormErrorHarness.getDataSkyIdPredicate(filters);
   }
 
-  async #getFormErrorClasses(): Promise<string[]> {
-    const formErrorClasses = await (await this.host()).getProperty('classList');
-    return Array.from(formErrorClasses);
-  }
-
   /*
```
Blackbaud-SteveBrush: Can you double-check the JSDocs comments in this file? I think the format needs to be the following in order for docs-tools to pick it up (I could be wrong):
```
/**
 *
 */
```

libs/components/forms/src/lib/modules/form-error/form-errors.component.html
Blackbaud-ErikaMcVey: good idea

libs/components/forms/src/lib/modules/form-error/form-error.component.ts
Blackbaud-ErikaMcVey: actually, we don't even need that class at all anymore--looks like the harness refactor made it unnecessary

Blackbaud-SteveBrush (approved):

blackbaud-sky-build-user:
Cherry-pick successful! :cherries: :tada: See https://github.com/blackbaud/skyux/pull/1988

PR #2198, opened 2024-04-01
feat: add startAtDate to SkyDatepicker to specify a date when the calendar initially opens

Blackbaud-JamesZhang:
### Use case
Currently, the datepicker allows disabling dates [by passing in min or max values](https://developer.blackbaud.com/skyux/components/datepicker?docs-active-tab=design#class-skydatepickerconfigservice).  However, the datepicker always loads to the current date with no way to configure it.

Given a case where the first valid date is months (or even years) in the past/future, because the datepicker always loads to the current date, the user needs to scroll through months of pages to find a valid date.

### Proposed changes
This PR adds a startAtDate property which allows the consumer to specify a specific date for the calendar to show when it initially opens.  Its functionality is nearly identical with [Angular Material Datepicker's startAt property](https://material.angular.io/components/datepicker/overview#setting-the-calendar-starting-view):

> The month, year, or range of years that the calendar opens to is determined by first checking if any date is currently selected, if so it will open to the month or year containing that date. Otherwise it will open to the month or year containing today's date. This behavior can be overridden by using the startAt property of <mat-datepicker>. In this case the calendar will open to the month or year containing the startAt date.

### Associated work item
[Feature 2861635](https://dev.azure.com/blackbaud/Products/_workitems/edit/2861635)
[Related Slack thread](https://blackbaud.slack.com/archives/C04U0PCBHNV/p1710447538257769)

### DIFF
--- a/libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar.component.ts
+++ b/libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar.component.ts
@@ -33,6 +33,9 @@ export class SkyDatepickerCalendarComponent {
   @Input()
   public maxDate: Date | undefined;
 
+  @Input()
+  public startAtDate: Date | undefined;
+
   /** currently selected date */
   @Input()
   public selectedDate: Date | undefined;
--- a/libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar.component.ts
+++ b/libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar.component.ts
@@ -113,8 +113,8 @@ export class SkyDatepickerCalendarComponent {
       this.selectedDate = value;
       this.datepicker?.select(value, false);
     } else {
-      this.selectedDate = new Date();
-      this.datepicker?.select(new Date(), false);
+      this.selectedDate = this.startAtDate ?? new Date();
+      this.datepicker?.select(this.startAtDate ?? new Date(), false);
     }
   }
 }
--- a/libs/components/datetime/src/lib/modules/datepicker/datepicker-config.service.ts
+++ b/libs/components/datetime/src/lib/modules/datepicker/datepicker-config.service.ts
@@ -20,6 +20,10 @@ export class SkyDatepickerConfigService {
    */
   public minDate: Date | undefined;
 
+  /**
+   * The date to open the calendar to initially if there isn't a date currently selected.
+   * @default the current date
+   */
   public startAtDate: Date | undefined;
 
   /**
--- a/libs/components/datetime/src/lib/modules/datepicker/datepicker-config.service.ts
+++ b/libs/components/datetime/src/lib/modules/datepicker/datepicker-config.service.ts
@@ -21,7 +21,7 @@ export class SkyDatepickerConfigService {
   public minDate: Date | undefined;
 
   /**
-   * The date to open the calendar to initially if there isn't a date currently selected.
+   * The date to open the calendar to initially.
    * @default the current date
    */
   public startAtDate: Date | undefined;
--- a/libs/components/datetime/src/lib/modules/datepicker/datepicker-config.service.ts
+++ b/libs/components/datetime/src/lib/modules/datepicker/datepicker-config.service.ts
@@ -20,6 +20,8 @@ export class SkyDatepickerConfigService {
    */
   public minDate: Date | undefined;
 
+  public startAtDate: Date | undefined;
+
   /**
    * The starting day of the week in the calendar,
    * where `0` sets the starting day to Sunday.
--- a/libs/components/datetime/src/lib/modules/datepicker/datepicker-input-fuzzy.directive.ts
+++ b/libs/components/datetime/src/lib/modules/datepicker/datepicker-input-fuzzy.directive.ts
@@ -159,7 +159,7 @@ export class SkyFuzzyDatepickerInputDirective
   }
 
   /**
-   * The fuzzy date to open the calendar to initially if there isn't a date currently selected. Place this attribute on the `input` element to set a date to open the calendar to.
+   * The fuzzy date to open the calendar to initially.
    * This property accepts a `SkyFuzzyDate` value that includes numeric month, day, and year values.
    * For example: `{ month: 1, day: 1, year: 2007 }`.
    * @default The current date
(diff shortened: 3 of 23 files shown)


libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -132,6 +132,19 @@ export class SkyDatepickerInputDirective
...
 
+  @Input()
+  public set startAtDate(value: Date | undefined) {
+    this.#_startAtDate = value;
+    this.#datepickerComponent.startAtDate = this.startAtDate;
+
+    this.#onValidatorChange();
+  }
+
+  // TODO: Refactor to not have getter logic
```
Blackbaud-JamesZhang, line 145: This refactor comment was copied from the existing `minDate` implementation above.  

@Blackbaud-TrevorBurch I found that it was originally added in [this PR](https://github.com/blackbaud/skyux/pull/769/files#diff-a172c578d81c8045a2191a112f9169b22eb83e3e9496968ed46304f1f824f847).  Let me know if we would rather use an updated implementation rather than copying from the existing one.

Blackbaud-JamesZhang:
I've only implemented the `startAtDate` logic for the standard Datepicker for now.  Given the team approves this approach, does the team also want to add this functionality to the fuzzy Datepicker as well?

blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/2198/storybook/)

Component Storybooks:

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/2198/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/2198/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/2198/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/2198/playground/)

libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar.component.ts
```diff
@@ -110,8 +113,8 @@ export class SkyDatepickerCalendarComponent {
       this.selectedDate = value;
       this.datepicker?.select(value, false);
     } else {
-      this.selectedDate = new Date();
-      this.datepicker?.select(new Date(), false);
+      this.selectedDate = this.startAtDate ?? new Date();
```
Blackbaud-JamesZhang, line 116: Do we want to validate the date is valid using `this.#formatter.dateIsValid`?  Here or in any other places.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -132,6 +132,19 @@ export class SkyDatepickerInputDirective
     return this.#_minDate || this.#configService.minDate;
   }
 
+  @Input()
```
Blackbaud-SteveBrush, line 139: Mind adding JSDocs to this input?

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -132,6 +132,19 @@ export class SkyDatepickerInputDirective
...
+  public set startAtDate(value: Date | undefined) {
+    this.#_startAtDate = value;
+    this.#datepickerComponent.startAtDate = this.startAtDate;
+
+    this.#onValidatorChange();
```
Blackbaud-SteveBrush: Would the `startAtDate` value ever affect the validation of the datepicker? You might be okay to leave this out.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
Blackbaud-SteveBrush: We're good to copy/paste for now.

libs/components/datetime/src/lib/modules/datepicker/datepicker-calendar.component.ts
Blackbaud-SteveBrush: I don't think so, since `startAtDate` doesn't require validation or affect other properties' validations.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
Blackbaud-JamesZhang: Removed the onValidatorChange

libs/components/datetime/src/lib/modules/datepicker/datepicker-input.directive.ts
```diff
@@ -132,6 +132,21 @@ export class SkyDatepickerInputDirective
     return this.#_minDate || this.#configService.minDate;
   }
 
+  /**
+   * The date to open the calendar to initially if there isn't a date currently selected. Place this attribute on the `input` element to override the default in `SkyDatepickerConfigService`.
```
Blackbaud-SteveBrush: I think we're okay to keep it simple with: `The date to open the calendar to initially.`

libs/components/datetime/src/lib/modules/datepicker/datepicker-config.service.ts
```diff
@@ -20,6 +20,12 @@ export class SkyDatepickerConfigService {
    */
   public minDate: Date | undefined;
 
+  /**
+   * The date to open the calendar to initially if there isn't a date currently selected.
```
Blackbaud-SteveBrush: I think we're okay to keep it simple with: `The date to open the calendar to initially.` The word "initially" already communicates that it can be overridden with the datepicker's selected value after the fact.

libs/components/datetime/src/lib/modules/datepicker/datepicker-input-fuzzy.directive.ts
```diff
@@ -158,6 +158,23 @@ export class SkyFuzzyDatepickerInputDirective
     return this.#_minDate;
   }
 
+  /**
+   * The fuzzy date to open the calendar to initially if there isn't a date currently selected. Place this attribute on the `input` element to set a date to open the calendar to.
```
Blackbaud-SteveBrush: I think we're okay to keep it simple with:
```
The fuzzy date to open the calendar to initially.
This property accepts a `SkyFuzzyDate` value that includes numeric month, day, and year values.
For example: `{ month: 1, day: 1, year: 2007 }`.
@default The current date
```

Blackbaud-SteveBrush (approved):

PR #2237, opened 2024-04-12
feat(components/ag-grid): compact option

johnhwhite:
[AB#2881723](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2881723)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/2237/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/2237/storybooks/ag-grid-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/2237/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/2237/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/2237/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/2237/playground/)

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
```diff
@@ -41,6 +56,9 @@ export class SkyAgGridWrapperComponent
   @HostBinding('class.sky-ag-grid-layout-normal')
   public isNormalLayout = false;
 
+  @Input({ transform: booleanAttribute })
```
Blackbaud-ErikaMcVey, line 64: this needs JSDocs. Also, from what I can tell only modern supports a compact theme. If they're in default and set this to true, would nothing happen? Or would they get a modern themed compact grid?

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
```diff
@@ -171,38 +186,23 @@ export class SkyAgGridWrapperComponent
   }
 
   public ngOnInit(): void {
-    this.#themeSvc?.settingsChange
+    combineLatest<[boolean, SkyThemeSettingsChange | undefined]>([
+      this.#hasEditableClass.pipe(distinctUntilChanged()),
+      this.#themeSvc?.settingsChange ?? of(undefined),
```
Blackbaud-ErikaMcVey, line 203: should the theme service even be optional? Is there a scenario where it wouldn't be provided?

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
```diff
@@ -215,7 +215,7 @@ export class SkyAgGridWrapperComponent
         agGridClassObserver.disconnect();
       });
     }
-    setTimeout(callback);
+    callback();
```
Blackbaud-ErikaMcVey, line 234: what changed that we don't need the `setTimeout` anymore?

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
```diff
@@ -287,34 +287,41 @@ export class SkyAgGridWrapperComponent
...
+  ): void {
+    const agTheme = agGridTheme(hasEditableClass, themeSettings, this.compact);
+    const previousValue = this.#wrapperClasses.getValue();
+    const previousTheme = previousValue.find((c) => c.startsWith('ag-theme-'));
     let value = [
-      ...this.#wrapperClasses
-        .getValue()
-        .filter((c) => !c.startsWith('ag-theme-')),
-      `ag-theme-sky-${variation}-${agTheme}`,
+      ...previousValue.filter((c) => !c.startsWith('ag-theme-')),
```
Blackbaud-ErikaMcVey, line 314: is it possible that there is more than one theme class? If not, can we just use the `previousTheme` we already found in this list?

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
```diff
@@ -287,34 +287,41 @@ export class SkyAgGridWrapperComponent
...
     const textSelectionClass = 'sky-ag-grid-text-selection';
-    if (this.#getTextSelection()) {
+    if (this.#getTextSelection(hasEditableClass)) {
       value.push(textSelectionClass);
     } else {
       value = value.filter((c) => c !== textSelectionClass);
     }
     this.#wrapperClasses.next([...new Set(value)]);
+    if (this.agGrid?.api && !this.agGrid.api.isDestroyed()) {
+      if (this.agGrid?.api?.getGridOption('domLayout') !== 'autoHeight') {
```
Blackbaud-ErikaMcVey: I thiiiiink typescript is smart enough that you don't need all the `?` before api, since the first if already guarantees it exists

also, why do we care if the `domLayout` is `autoHieght` here? I thought `autoHeight` set the height of the grid to the number of rows x the height of the rows. If the theme changed to compact wouldn't we still want to reset the heights then?

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
```diff
@@ -287,34 +287,41 @@ export class SkyAgGridWrapperComponent
...
+        this.agGrid?.api?.resetRowHeights();
+      }
+      this.agGrid?.api?.refreshHeader();
+      if (
+        previousTheme?.endsWith('-compact') !==
```
Blackbaud-ErikaMcVey: might be nice to make this a function similar to `agGridThemeIsCompact`, maybe `agGridHasCompactClass`?

libs/components/ag-grid/src/lib/styles/ag-grid-theme.ts
```diff
@@ -0,0 +1,31 @@
+import { SkyThemeSettings } from '@skyux/theme';
+
+type SkyAgGridThemeType =
+  `ag-theme-sky-${'data-entry-grid' | 'data-grid'}-${'default' | 'modern-light' | 'modern-dark' | 'modern-light-compact' | 'modern-dark-compact'}`;
+
+export function agGridThemeIsCompact(
```
Blackbaud-ErikaMcVey, line 6: any reason you didn't want to add these functions to the ag grid service? Was it just getting bloated? I'm not opposed to this, just don't see us export helper functions individually too often

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid.service.spec.ts
```diff
@@ -235,10 +229,7 @@ describe('SkyAgGridService', () => {
     });
 
     it('should unsubscribe from the theme service when destroyed', () => {
-      const overrideOptions = { gridOptions: {} };
-      let gridOptions = agGridService.getGridOptions(overrideOptions);
-
-      expect(gridOptions.rowHeight).toBe(38);
+      expect(agGridService.getHeaderHeight()).toBe(37);
```
Blackbaud-ErikaMcVey, line 232: I realize this was an existing test already but why are we checking rowHeight in "should unsubscribe from the theme service when destroyed"?

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
```diff
@@ -171,38 +186,23 @@ export class SkyAgGridWrapperComponent
...
-          // On subsequent theme changes, we need to call the api to re-render the grid.
-          this.#currentTheme = settings.currentSettings;
-          this.agGrid.api.setHeaderHeight(
-            this.#agGridService.getHeaderHeight(),
-          );
-          this.agGrid.api.resetRowHeights();
-          this.agGrid.api.refreshCells();
-        }
+      .subscribe(([hasEditableClass, settings]) => {
+        this.#updateGridTheme(hasEditableClass, settings?.currentSettings);
```
Blackbaud-ErikaMcVey: so it looks like we update the theme if editable or the theme changes, and when we update theme we check if it is compact. Does it trigger an update when the compact input changes, or would that change not take effect until the editable state or themeSettings changed? I could see that being something that consumers give the user the option to toggle between?

libs/components/ag-grid/src/lib/styles/ag-grid-theme.ts
johnhwhite: I went with this file because I wanted to minimize the number of places with this logic without adding these functions to the public API. On the service, the methods would need to be public.

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
johnhwhite: Good call. It only applies in modern since the default theme is already pretty compact.

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
johnhwhite: Good call. Updated.

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
johnhwhite: By using the observables rather than making the changes directly, the `setTimeout` was no longer needed.

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
johnhwhite: This is the exclusion of the previous theme.

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid.service.spec.ts
johnhwhite: Might've been a coverage thing, but I like that we have a test for the style so I'm not going to pull it out.

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
johnhwhite: Good call, added.

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
johnhwhite: The extra `?`'s are due to insufficient mocks in tests. Added a comment for the `autoHeight` bit.

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
johnhwhite: It's not necessarily provided in tests and requiring it becomes a breaking change.

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid.service.spec.ts
Blackbaud-ErikaMcVey: Oh I wasn't suggesting pulling it out of testing entirely--moreso renaming the test or, if the style is not relevant to anything else in the test, giving the styles their own separate test.

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-wrapper.component.ts
Blackbaud-ErikaMcVey: interesting about the warning. If we're redrawing all the rows every time do we need to call `resetRowHeights` at all?

johnhwhite: Right! We don't need to force redraw rows. Good call.

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid.service.spec.ts
johnhwhite: After thinking about it, the visual tests are good enough for testing the CSS. I'm going to leave this as-is, and the test can be removed when `getHeaderHeight` gets removed.

Blackbaud-ErikaMcVey (approved):

PR #2246, opened 2024-04-18
fix(components/datetime): date range picker announces required state to screen readers

Blackbaud-TrevorBurch:
[AB#2837769](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2837769)

### DIFF
--- a/apps/playground/src/app/components/datetime/date-range-picker/date-range-picker.component.ts
+++ b/apps/playground/src/app/components/datetime/date-range-picker/date-range-picker.component.ts
@@ -22,6 +22,7 @@ export class DateRangePickerComponent {
   public dateFormat: string;
   public disabled = false;
   public endDateRequired = false;
+  public hintText: string | undefined;
   public reactiveForm: UntypedFormGroup;
   public startDateRequired = false;
 
@@ -122,5 +123,14 @@ export class DateRangePickerComponent {
 
   public toggleStartDateRequired(): void {
     this.startDateRequired = !this.startDateRequired;
+  }
+
+  public toggleHintText(): void {
+    if (this.hintText) {
+      this.hintText = undefined;
+    } else {
+      this.hintText =
+        'Really long hint text that should wrap and be confined to below the date range picker element. We set the text to have a small margin below the selection and datepicker controls, and if there are any errors (shown below) there will be a slight spacing between this element and those errors.';
+    }
   }
 }
--- a/apps/playground/src/app/components/datetime/date-range-picker/date-range-picker.component.ts
+++ b/apps/playground/src/app/components/datetime/date-range-picker/date-range-picker.component.ts
@@ -1,4 +1,4 @@
-import { Component } from '@angular/core';
+import { ChangeDetectorRef, Component, inject } from '@angular/core';
 import {
   AbstractControl,
   FormControl,
@@ -25,6 +25,7 @@ export class DateRangePickerComponent {
   public reactiveForm: UntypedFormGroup;
   public startDateRequired = false;
 
+  readonly #changeDetector = inject(ChangeDetectorRef);
   #required = false;
 
   public get pickerFormControl(): AbstractControl {
@@ -42,10 +43,12 @@ export class DateRangePickerComponent {
         status,
         this.pickerFormControl.errors,
       );
+      this.#changeDetector.detectChanges();
     });
 
     this.pickerFormControl.valueChanges.subscribe((value) => {
       console.log('Date range value change:', value);
+      this.#changeDetector.detectChanges();
     });
   }
 
--- a/apps/playground/src/app/components/datetime/date-range-picker/date-range-picker.component.ts
+++ b/apps/playground/src/app/components/datetime/date-range-picker/date-range-picker.component.ts
@@ -44,12 +44,12 @@ export class DateRangePickerComponent {
         status,
         this.pickerFormControl.errors,
       );
-      this.#changeDetector.detectChanges();
+      this.#changeDetector.markForCheck();
     });
 
     this.pickerFormControl.valueChanges.subscribe((value) => {
       console.log('Date range value change:', value);
-      this.#changeDetector.detectChanges();
+      this.#changeDetector.markForCheck();
     });
   }
 
@@ -69,6 +69,7 @@ export class DateRangePickerComponent {
       this.pickerFormControl.addValidators(Validators.required);
     }
     this.pickerFormControl.updateValueAndValidity();
+    this.#changeDetector.markForCheck();
     this.#required = !this.#required;
   }
 
--- a/apps/playground/src/app/components/datetime/date-range-picker/date-range-picker.component.ts
+++ b/apps/playground/src/app/components/datetime/date-range-picker/date-range-picker.component.ts
@@ -1,9 +1,11 @@
 import { Component } from '@angular/core';
 import {
   AbstractControl,
+  FormControl,
   UntypedFormBuilder,
   UntypedFormControl,
   UntypedFormGroup,
+  Validators,
 } from '@angular/forms';
 import {
   SkyDateRangeCalculation,
@@ -23,12 +25,15 @@ export class DateRangePickerComponent {
   public reactiveForm: UntypedFormGroup;
   public startDateRequired = false;
 
+  #required = false;
+
   public get pickerFormControl(): AbstractControl {
     return this.reactiveForm.get('lastDonation');
   }
 
   constructor(formBuilder: UntypedFormBuilder) {
     this.reactiveForm = formBuilder.group({
+      firstName: new FormControl<string>(''),
       lastDonation: new UntypedFormControl(),
     });
     this.pickerFormControl.statusChanges.subscribe((status) => {
@@ -51,6 +56,16 @@ export class DateRangePickerComponent {
     } else {
       this.reactiveForm.disable();
     }
+  }
+
+  public toggleRequiredDate(): void {
+    if (this.#required) {
+      this.pickerFormControl.removeValidators(Validators.required);
+    } else {
+      this.pickerFormControl.addValidators(Validators.required);
+    }
+    this.pickerFormControl.updateValueAndValidity();
+    this.#required = !this.#required;
   }
 
   public resetForm(): void {
--- a/libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
+++ b/libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
@@ -357,12 +357,8 @@ export class SkyDateRangePickerComponent
   }
 
   public ngOnInit(): void {
-    this.#ngControl = this.#injector.get(NgControl, null, {
-      optional: true,
-    });
-    if (this.#ngControl) {
-      this.#ngControl.valueAccessor = this;
-    }
+    this.#ngControl = this.#injector.get(NgControl);
+    this.#ngControl.valueAccessor = this;
 
     this.#createForm();
 
--- a/libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
+++ b/libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
@@ -13,6 +13,7 @@ import {
   Optional,
   SimpleChanges,
   TemplateRef,
+  booleanAttribute,
   forwardRef,
   inject,
 } from '@angular/core';
@@ -212,16 +213,17 @@ export class SkyDateRangePickerComponent
   public display: string | undefined;
 
   /**
-   * The template driven required input. We need to watch for this - but do not want to document it outside of the typical Angular template driven form documentation.
-   * @internal
+   * Whether the date range picker requires a value.
    */
-  @Input()
-  public set required(value: boolean | undefined) {
-    this.#_required = value;
-    this.#updateRequiredStates();
+  @Input({ transform: booleanAttribute })
+  public set required(value: boolean) {
+    if (value !== this.#_required) {
+      this.#_required = value;
+      this.#updateRequiredStates();
+    }
   }
 
-  public get required(): boolean | undefined {
+  public get required(): boolean {
     return this.#_required;
   }
 
@@ -309,7 +311,7 @@ export class SkyDateRangePickerComponent
   ];
   #_dateFormat: string | undefined;
   #_disabled: boolean | undefined = false;
-  #_required: boolean | undefined = false;
+  #_required = false;
   #_valueOrDefault: SkyDateRangeCalculation | undefined;
 
   #changeDetector: ChangeDetectorRef;
@@ -710,8 +712,7 @@ export class SkyDateRangePickerComponent
   #updateRequiredStates(): void {
     const originalValue = this.outerControlRequired;
     this.outerControlRequired =
-      this.required ||
-      !!this.#ngControl?.control?.hasValidator(Validators.required);
+      this.required || !!this.#control?.hasValidator(Validators.required);
 
     if (this.outerControlRequired) {
       this.#calculatorIdControl?.addValidators(Validators.required);
@@ -725,6 +726,8 @@ export class SkyDateRangePickerComponent
 
     // We need to update the outer control - but only if the required state has changed. Otherwise - we would enter an infinite loop.
     if (originalValue !== this.outerControlRequired) {
+      // Ensure that child control validation is complete prior to updating the outer control. Without this - the inner controls may not have validated and may lead
+      // to an incorrect outer control state until something else re-runs validation.
       this.#ngZone.onStable.pipe(first()).subscribe(() => {
         this.#control?.updateValueAndValidity();
       });
--- a/libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
+++ b/libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
@@ -165,6 +165,7 @@ export class SkyDateRangePickerComponent
   /**
    * [Persistent inline help text](https://developer.blackbaud.com/skyux/design/guidelines/user-assistance#inline-help) that provides
    * additional context to the user.
+   * @preview
    */
   @Input()
   public hintText: string | undefined;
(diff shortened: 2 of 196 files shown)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/2246/storybook/)

Component Storybooks:

- [a11y-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/a11y-storybook/)
- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/colorpicker-storybook/)
- [core-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/core-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/errors-storybook/)
- [flyout-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/flyout-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/forms-storybook/)
- [help-inline-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/help-inline-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/lookup-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/modals-storybook/)
- [navbar-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/navbar-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/phone-field-storybook/)
- [popovers-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/popovers-storybook/)
- [progress-indicator-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/progress-indicator-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/theme-storybook/)
- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/tiles-storybook/)
- [toast-storybook](https://blackbaud.github.io/skyux-pr-preview/2246/storybooks/toast-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/2246/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/2246/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/2246/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/2246/playground/)

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
```diff
@@ -172,13 +172,15 @@ export class SkyDateRangePickerComponent
...
    * @default false
+   * @deprecated Use standard form field validation to make the entire control as required instead.
    */
   @Input()
   public startDateRequired: boolean | undefined = false;
 
   /**
    * Whether to require users to specify a end date.
    * @default false
+   * @deprecated Use standard form field validation to make the entire control as required instead.
```
Blackbaud-SteveBrush: For these deprecation messages, can we be a bit more specific about what "standard form field validation" means?
Maybe something like the following?
```
Use Angular's `Validators.required` on the form control to mark the date range picker as required.
```

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
```diff
@@ -411,6 +414,14 @@ export class SkyDateRangePickerComponent
   public validate(control: AbstractControl): ValidationErrors | null {
     if (!this.#control) {
       this.#control = control;
+
+      this.#updateRequiredStates();
+
+      this.#control.statusChanges
+        ?.pipe(takeUntil(this.#ngUnsubscribe))
+        .subscribe(() => {
```
Blackbaud-SteveBrush: Subscribing in the `validate` method is a bit dangerous since it gets called every time the field's validations trigger.

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
```diff
@@ -631,6 +642,29 @@ export class SkyDateRangePickerComponent
       });
   }
 
+  #updateRequiredStates(): void {
```
Blackbaud-SteveBrush: I agree, this solution feels a bit "hacky". 😅 Could we instead use Angular's `AbstractControl` API to respond to (and set) the required states for the child controls? I found a few examples online where folks were getting the `AbstractControl` of the parent `NgControl` and using that to set the required states. Here's a rough compilation of those findings that we might be able to use for this feature: https://stackblitz.com/edit/stackblitz-starters-javnqs?file=src%2Fcustom-form-control.ts

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
Blackbaud-TrevorBurch: This unfortunately is the only thing we can watch to then react if validators are dynamically added or removed. We do the same thing in text editor. I wish Angular had a `validatorChanges` event - but it doesn't :(

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
Blackbaud-TrevorBurch: This has given me some ideas. I'll work on tweaks.

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
Blackbaud-SteveBrush: It's safe to subscribe to the `statusChanges` event, just not in the `validate` method. 🖖

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
Blackbaud-SteveBrush: This is the article I robbed some stuff from: https://levelup.gitconnected.com/angular-get-control-in-controlvalueaccessor-b7f09a485fba

Blackbaud-TrevorBurch: I reworked this. NgControl is now pulled in during construction instead of being set in the validator. 

I also updated the shared form utility to be able to handle different form types and to not call the validator directly.

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
Blackbaud-TrevorBurch: Updated.

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
```diff
@@ -207,6 +211,20 @@ export class SkyDateRangePickerComponent
   @HostBinding('style.display')
   public display: string | undefined;
 
+  /**
+   * The template driven required input. We need to watch for this - but do not want to document it outside of the typical Angular template driven form documentation.
+   * @internal
+   */
+  @Input()
```
Blackbaud-SteveBrush: In the Stackblitz example, I used Angular's `booleanAttribute` transform to allow the consumer to pass in string "true" or "false". To detect changes, you can use the `OnChanges` lifecycle hook. Angular does the same thing with their components: https://github.com/angular/components/blob/main/src/material/checkbox/checkbox.ts#L173

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
```diff
@@ -42,12 +45,6 @@ import { SkyDateRangeCalculator } from './types/date-range-calculator';
 import { SkyDateRangeCalculatorId } from './types/date-range-calculator-id';
 import { SkyDateRangeCalculatorType } from './types/date-range-calculator-type';
 
-const SKY_DATE_RANGE_PICKER_VALUE_ACCESSOR = {
```
Blackbaud-SteveBrush, line 45: Curious why this provider was removed?

apps/playground/src/app/components/datetime/date-range-picker/date-range-picker.component.ts
```diff
@@ -37,10 +43,12 @@ export class DateRangePickerComponent {
         status,
         this.pickerFormControl.errors,
       );
+      this.#changeDetector.detectChanges();
```
Blackbaud-SteveBrush: A hard `detectChanges()` is needed?

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
```diff
@@ -332,6 +355,13 @@ export class SkyDateRangePickerComponent
   }
 
   public ngOnInit(): void {
+    this.#ngControl = this.#injector.get(NgControl, null, {
+      optional: true,
+    });
+    if (this.#ngControl) {
+      this.#ngControl.valueAccessor = this;
```
Blackbaud-SteveBrush: I'm assuming this is why you removed the provider, above. Do we need to do it this way? Can we remove this line and just keep the provider?

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
```diff
@@ -332,6 +355,13 @@ export class SkyDateRangePickerComponent
   }
 
   public ngOnInit(): void {
+    this.#ngControl = this.#injector.get(NgControl, null, {
+      optional: true,
```
Blackbaud-SteveBrush: You explained this to me earlier, but can you remind me in which situation the `NgControl` is optional?

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
```diff
@@ -664,6 +707,30 @@ export class SkyDateRangePickerComponent
       });
   }
 
+  #updateRequiredStates(): void {
+    const originalValue = this.outerControlRequired;
+    this.outerControlRequired =
+      this.required ||
+      !!this.#ngControl?.control?.hasValidator(Validators.required);
```
Blackbaud-SteveBrush: Can you replace `this.#ngControl?.control` with `this.#control`?

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
```diff
@@ -664,6 +707,30 @@ export class SkyDateRangePickerComponent
...
+      this.#calculatorIdControl?.removeValidators(Validators.required);
+    }
+
+    this.#calculatorIdControl?.updateValueAndValidity();
+    this.#startDateControl?.updateValueAndValidity();
+    this.#endDateControl?.updateValueAndValidity();
+
+    // We need to update the outer control - but only if the required state has changed. Otherwise - we would enter an infinite loop.
+    if (originalValue !== this.outerControlRequired) {
+      this.#ngZone.onStable.pipe(first()).subscribe(() => {
```
Blackbaud-SteveBrush: Curious about the use of `ngZone`. I removed it and everything seemed fine; what scenario causes a problem that requires this line?

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
Blackbaud-TrevorBurch: With the way this provider works - you can't use `inject` and this provider or Angular throws a circular reference error. Other places that we use `inject` on the `NgControl` such as checkbox follow this pattern of not using the provider and then assigning the `valueAccessor` property.

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
Blackbaud-TrevorBurch: Addressed in my other comment. TL;DR - it causes a circular reference error.

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
Blackbaud-TrevorBurch: I added a comment here. It is needed because without it the child control validation may not have finished. I verified this in the playground. Without the `ngZone` the outer control will not show as invalid if the required validator is added until something else retriggers validation. But with this wrapper it will automatically show as invalid when the validator is added.

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
Blackbaud-TrevorBurch: Done

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
Blackbaud-TrevorBurch: I added the transform. We chatted some in flex - the transform here causes the value to always be a `boolean` type. I think it is more efficient to do this in the setter that will only fire on changes to `required` than having logic in `ngOnChanges` that would run every time change detection runs on the component. 

I did add an if-statement in the setter that ensures that the logic only runs on changes to `required`. This gives the same check that Angular was doing in `ngOnChanges` but ensures that it only runs when the setter is called instead of on every change detection cycle.

apps/playground/src/app/components/datetime/date-range-picker/date-range-picker.component.ts
Blackbaud-TrevorBurch: `markForCheck` was not updating the playground to show the changes on the controls when the validators changed without this. I wasn't able to find anything I could do in the playground or datepicker that would pick up that change without the hard `detectChanges` here. Since this was the playground - I didn't worry to much about it. Happy to chat if you'd like around this one.

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
Blackbaud-TrevorBurch: I did this because we do have a couple unit tests that test the control without a form control or ngModel attached. I am a little worried this might be a breaking change if we don't make this optional (as much as I would like to just make it required here).

apps/playground/src/app/components/datetime/date-range-picker/date-range-picker.component.ts
Blackbaud-SteveBrush: Yes, it's the playground, but it's concerning that we have to call `detectChanges` to get it to work (will the consumer have to do this too?). It seems to only affect the deprecated `startDateRequired` and `endDateRequired` inputs, so we can ignore this for now.

libs/components/forms/src/lib/modules/required-state/required-state.directive.ts
```diff
@@ -1,27 +1,80 @@
...
+  #ngControl?: NgControl | null;
+  readonly #ngUnsubscribe = new Subject<void>();
+
+  public ngOnDestroy(): void {
+    this.#ngUnsubscribe.next();
+    this.#ngUnsubscribe.complete();
+  }
+
+  public ngAfterViewInit(): void {
+    this.#ngControl = this.#injector.get(NgControl, undefined, {
```
Blackbaud-TrevorBurch, line 67: Doing this in the `AfterViewInit` was needed to get around a circular reference when the ControlValueAccessor gave the `NG_VALIDATORS` provider in its providers array. We do this in a couple other components. Some do it on `ngOnInit` but `statusChanges` is not ready until `ngAfterViewInit`.

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.ts
```diff
@@ -337,9 +353,32 @@ export class SkyDateRangePickerComponent
       .subscribe(() => {
         this.#changeDetector.markForCheck();
       });
+
+    this.requiredState.requiredStateChange
```
Blackbaud-TrevorBurch, line 357: I had to add a `requiredStateChange` event to the directive. This is required because we need to update the outer form control once the required states change. With text-editor change detection was enough - but it is not here as we need validators to rerun and such in order to get all changes back to the outer control. We have to listen to when the required state changes, then update the inner controls to be required, then run the validators again for the outer control.

Blackbaud-SteveBrush:
Closing in favor of https://github.com/blackbaud/skyux/pull/2311

PR #2344, opened 2024-06-03
feat(components/forms): make `headingLevel` optional for checkbox and radio group components

Blackbaud-ErikaMcVey:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/2344/storybook/)

Component Storybooks:

- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/colorpicker-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/errors-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/forms-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/lookup-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/phone-field-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/2344/storybooks/theme-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/2344/code-examples/)
- [dep-graph](https://blackbaud.github.io/skyux-pr-preview/2344/dep-graph/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/2344/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/2344/playground/)

libs/components/forms/src/lib/modules/checkbox/checkbox-group.component.ts
```diff
@@ -69,22 +69,26 @@ export class SkyCheckboxGroupComponent {
...
   /**
-   * The semantic heading level in the document structure.
+   * The semantic heading level in the document structure. By default, the heading text is not wrapped in a heading element.
    * @preview
-   * @default 3
    */
   @Input({ transform: numberAttribute })
-  public headingLevel: SkyCheckboxGroupHeadingLevel = 3;
+  public set headingLevel(value: SkyCheckboxGroupHeadingLevel | undefined) {
+    this.#_headingLevel = value && !isNaN(value) ? value : undefined;
```
Blackbaud-TrevorBurch, line 77: Why is the `isNaN` check needed? The transform will mean that it is always either a number or undefined, correct?

libs/components/forms/src/lib/modules/radio/radio-group.component.ts
```diff
@@ -116,22 +116,26 @@ export class SkyRadioGroupComponent
...
   /**
-   * The semantic heading level in the document structure.
+   * The semantic heading level in the document structure. By default, the heading text is not wrapped in a heading element.
    * @preview
-   * @default 3
    */
   @Input({ transform: numberAttribute })
-  public headingLevel: SkyRadioGroupHeadingLevel = 3;
+  public set headingLevel(value: SkyRadioGroupHeadingLevel | undefined) {
+    this.#_headingLevel = value && !isNaN(value) ? value : undefined;
```
Blackbaud-TrevorBurch, line 124: Same `isNaN` question

libs/components/forms/testing/src/checkbox/checkbox-group-harness.ts
```diff
@@ -129,6 +138,18 @@ export class SkyCheckboxGroupHarness extends SkyComponentHarness {
...
   }
 
+  /**
+   * Whether the checkbox group is stacked.
+   */
+  public async getStacked(): Promise<boolean> {
+    const host = await this.host();
+
+    return (
+      (await host.hasClass('sky-margin-stacked-lg')) ||
```
Blackbaud-TrevorBurch: In theory - the consumer could specify these as well, correct?

Should we check for the class and the heading together to ensure that the correct class is there for the heading level and the consumer hasn't just added one of the classes to the host element?

libs/components/forms/src/lib/modules/checkbox/checkbox-group.component.ts
Blackbaud-ErikaMcVey: no, unfortunately the transform means it is either a number or `NaN` 😭 so even when the consumer doesn't set it and you would think it was undefined, it is actually `NaN`. `undefined` is easier to work with everywhere else so I transform it in the setter

libs/components/forms/testing/src/checkbox/checkbox-group-harness.ts
Blackbaud-ErikaMcVey: like only return `true` if xl is used with a heading or lg is used without a heading? The consumer still could've specified the class and just specified the right one for the scenario. Do you think if they didn't specify it by setting `stacked` we should return false? Or only return true if it is the right combo?

libs/components/forms/src/lib/modules/checkbox/checkbox-group.component.ts
Blackbaud-TrevorBurch: Ah. Ok. Makes sense.

libs/components/forms/testing/src/checkbox/checkbox-group-harness.ts
Blackbaud-TrevorBurch: > Do you think if they didn't specify it by setting stacked we should return false?

I don’t know that we need to account for accidental correctness. 

> Or only return true if it is the right combo?

Right, I would only return true though if the combination matches correctly.

Blackbaud-TrevorBurch (dismissed):

Blackbaud-TrevorBurch (approved):

PR #2564, opened 2024-07-26
feat(components/packages): schematic to remove compat stylesheets

johnhwhite:
[AB#2893837](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2893837)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/2564/storybook/)

Component Storybooks:

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/2564/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/2564/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/2564/playground/)

libs/components/packages/src/schematics/ng-generate/remove-compat-stylesheets/schema.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "http://json-schema.org/schema",
+  "type": "object",
+  "properties": {
+    "version": {
+      "type": "string",
+      "description": "Remove the compatibility stylesheets at or below the specified version.",
```
Blackbaud-SteveBrush: The "at OR below" threw me off (should it be "less than or equal to" or just "less than"?) Also, the fact that this argument takes a version number or "current" might cause confusion.

The name of the schematic communicates that ALL stylesheets will be removed when it's called without arguments. Could we simplify the API and add an argument that lets the user keep the current stylesheet intact? (I'm thinking about this from the perspective of one of our public-facing consumers running the command.)
```
ng g @skyux/packages:remove-compat-stylesheets --keep-current
```
If the story dictates that the default logic should be switched, you could go with:
```
ng g @skyux/packages:remove-compat-stylesheets --all
```
Let me know your thoughts.

johnhwhite: Makes sense. Updated the parameter, and the default is remove-all.

This
```
ng g @skyux/packages:remove-compat-stylesheets --below-version=9
```
removes e.g., `skyux8-compat.css`.

And this
```
ng g @skyux/packages:remove-compat-stylesheets
```
removes e.g., `skyux8-compat.css`, `skyux9-compat.css`, `skyux10-compat.css`, etc.

libs/components/packages/src/schematics/ng-generate/remove-compat-stylesheets/remove-compat-stylesheets.schematic.ts
```diff
@@ -0,0 +1,49 @@
...
+          const stylesheets = project.targets.get(targetName)?.options?.[
+            'styles'
+          ] as string[] | undefined;
+
+          if (stylesheets) {
+            for (let i = stylesheets.length - 1; i >= 0; i--) {
+              const stylesheet = stylesheets[i];
+
+              const check = stylesheet.match(COMPAT_STYLESHEET_REGEX);
+              if (check && parseInt(check[1]) <= belowVersion) {
```
Blackbaud-SteveBrush: The logic here is "less than or equal to", but the option is `belowVersion`. Should we change the logic to `<` or the name of the option to match the logic?

How likely do you think our consumers will want to delete stylesheets less than a specific version? The feature we're going for seems to be, "delete all _outdated_ compat stylesheets", but it feels like an overcomplication to give them the ability to select the version that they consider "outdated". I still think it would be a better API to have an option that communicates "keep the current version's stylesheet" instead of asking them to provide a "less than" version. The schematic would handle the logic of which version is considered "current", and the pipeline (or consumer) executing the command would simply pass the same arguments every time. Otherwise, we'll have to include "current version" logic in the pipeline when we decide to run this command.

I imagine the `skyuxconfig.json` option would be something akin to:
```
{
  "updateSettings": {
    "includeCompatStylesheet": false
  }
}
```
And the pipeline would apply the configuration to the command like this:
```
ng g @skyux/packages:remove-compat-stylesheets --keep-current={config.updateSettings.includeCompatStylesheet}
```
The argument doesn't necessarily need to be named `--keep-current`; the flag just needs to represent a feature, instead of an internal working detail.

johnhwhite: Updated the comparator. This approach lets the same schematic work for the update migration and, if the config opts-out of compat stylesheets, we can call `ng g @skyux/packages:remove-compat-stylesheets`.

Blackbaud-SteveBrush (approved):

PR #2651, opened 2024-08-27
fix(components/packages): update `move-class-to-library` to handle partial migration

johnhwhite:


libs/components/packages/src/schematics/utility/move-class-to-library.ts
```diff
@@ -27,47 +32,39 @@ export function moveClassToLibrary(
...
-      sourceFile,
-      ts.SyntaxKind.ImportSpecifier,
-    ).filter(
-      (node): node is ts.ImportSpecifier =>
-        ts.isImportSpecifier(node) && node.name.text === importName,
-    )[0];
+  moduleImports.forEach((moduleImport) => {
+    options.classNames.forEach((className) => {
+      const namedImportSpecifier = findNodes(
+        moduleImport,
```
johnhwhite, line 38: Only look for nodes within this import, rather than the whole source file.

Blackbaud-SteveBrush (approved):

PR #2743, opened 2024-09-13
fix(components/ag-grid): sort direction button is not visible when sorting is not active on a column

Blackbaud-TrevorBurch:
[AB#3058366](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3058366)


libs/components/ag-grid/src/lib/modules/ag-grid/header/header.component.html
```diff
@@ -35,12 +35,12 @@
         /></span>
       }
       @if (params?.enableSorting) {
-        <button
-          class="ag-sort-indicator-container sky-btn sky-btn-icon-borderless"
-          type="button"
-          (click)="onSortRequested($event)"
-        >
-          @if (sortOrder$ | async; as sortDirection) {
+        @if (sortOrder$ | async; as sortDirection) {
```
Blackbaud-TrevorBurch, line 38: This does technically change things to where it is not possible to have the "sortIndexDisplay` visible without `sortDirection`. However, based on my research of AG Grid - this isn't possible anyway as there is only ever an index when a column is sorting and part of a multi-sort scenario. So there is no functional change here.

We needed the screen reader text pulled out of the button so that it could be in the accessible display without the button also being rendered. We had an existing `:empty` css selector for this button; however, it wasn't working due to the addition of the screen reader text.

blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/2743/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/2743/storybooks/ag-grid-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/2743/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/2743/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/2743/playground/)

libs/components/ag-grid/src/lib/modules/ag-grid/header/header.component.html
Blackbaud-SteveBrush: Seems like you could still write a unit test for this behavior? On init, the sort button should not exist, and after clicking the heading it should?

Also, if the `:empty` style isn't being used anymore, let's remove it.

Blackbaud-TrevorBurch: I've added unit testing and removed the `:empty` style

Blackbaud-SteveBrush (approved):

PR #2768, opened 2024-09-23
feat(components/colorpicker): add colorpicker test harness 

Blackbaud-SandhyaRajasabeson:
[AB#2195390](https://dev.azure.com/blackbaud/Products/_boards/board/t/SKY%20UX%20Program/Stories/?workitem=2195390)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/2768/storybook/)

Component Storybooks:

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/2768/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/2768/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/2768/playground/)

libs/components/help-inline/testing/src/help-inline/help-inline-harness.ts
```diff
@@ -105,9 +104,7 @@ export class SkyHelpInlineHarness extends SkyComponentHarness {
   /**
    * Gets the help popover content.
    */
-  public async getPopoverContent(): Promise<
-    TemplateRef<unknown> | string | undefined
-  > {
+  public async getPopoverContent(): Promise<string | undefined> {
```
Blackbaud-SteveBrush: We should revert this change since it would be considered "breaking".

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  /**
+   * Whether the colorpicker component is stacked.
+   */
+  public async isStacked(): Promise<boolean> {
+    return (await this.host()).hasClass('sky-margin-stacked-lg');
+  }
+
+  private async getOverlayId(): Promise<string | null> {
```
Blackbaud-SteveBrush: `#getOverlayId`

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
```diff
@@ -0,0 +1,201 @@
...
+    const overlayId = await this.getOverlayId();
+
+    if (!overlayId) {
+      throw new Error(
+        'Unable to get colorpicker dropdown because dropdown is closed',
+      );
+    }
+
+    return await this.#documentRootLocator.locatorFor(
+      SkyColorpickerDropdownHarness.with({ id: overlayId }),
```
Blackbaud-SteveBrush: You probably don't need to create a new `id` filter for this. See how the dropdown harness uses the `selector` filter: https://github.com/blackbaud/skyux/blob/main/libs/components/popovers/testing/src/dropdown/harness/dropdown-harness.ts#L87-L89

libs/components/colorpicker/testing/src/colorpicker/colorpicker-dropdown-harness.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  private async getId(): Promise<string | null> {
+    return (await this.host()).getAttribute('id');
+  }
+
+  private async getRedInput(): Promise<TestElement> {
+    return (await this.#getInputs())[1];
+  }
+
+  private async getSwatchButtons(): Promise<TestElement[]> {
```
Blackbaud-SteveBrush: ☝️ All the "private" methods should start with `#`.

libs/components/colorpicker/testing/src/colorpicker/colorpicker-dropdown-harness.filters.ts
```diff
@@ -0,0 +1,10 @@
+import { BaseHarnessFilters } from '@angular/cdk/testing';
+
+/**
+ * A set of criteria that can be used to filter a list of `SkyColorpickerDropdownHarness` instances.
+ */
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface SkyColorpickerDropdownHarnessFilters
+  extends BaseHarnessFilters {
+  id?: string;
```
Blackbaud-SteveBrush: Probably not needed (see my comment below about using `selector`).

libs/components/colorpicker/testing/src/colorpicker/colorpicker-dropdown-harness.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  /**
+   * Enters a value into the alpha input box.
+   * @param value A decimal value from 0-1.
+   */
+  public async setAlphaValue(value: string): Promise<void> {
+    const input = await this.getAlphaInput();
+
+    input.setInputValue(value);
+    input.dispatchEvent('input');
```
Blackbaud-SteveBrush: This operation is copied in multiple methods. What do you think about creating a shared private method to handle all cases?
```
const input = await this.getAlphaInput();
await this.#setInputValue(input, value);
```

libs/components/colorpicker/testing/src/colorpicker/colorpicker-dropdown-harness.ts
```diff
@@ -0,0 +1,153 @@
...
+      }
+    }
+  }
+
+  /**
+   * Enters a value into the alpha input box.
+   * @param value A decimal value from 0-1.
+   */
+  public async setAlphaValue(value: string): Promise<void> {
+    const input = await this.getAlphaInput();
```
Blackbaud-SteveBrush: This should throw a helpful error if the alpha button cannot be found (since it's enabled by the `allowTransparency` flag).

libs/components/colorpicker/testing/src/colorpicker/colorpicker-dropdown-harness.ts
```diff
@@ -0,0 +1,153 @@
...
+   */
+  public async clickCancelButton(): Promise<void> {
+    return (await this.#getCancelButton()).click();
+  }
+
+  /**
+   * Clicks a specified swatch in the color preset section.
+   * @param swatchHex Hex code of the swatch to click.
+   */
+  public async clickSwatch(swatchHex: string): Promise<void> {
```
Blackbaud-SteveBrush: Let's name this `clickPresetColorSwatch` to match the name used in the design guidelines anatomy section. This will also protect us if we create a different type of swatch in the future.

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
```diff
@@ -0,0 +1,201 @@
...
+  );
+  #resetButtonModern = this.locatorForOptional(
+    'button.sky-colorpicker-reset-button-modern',
+  );
+
+  async #getFormErrors(): Promise<SkyFormErrorsHarness> {
+    return await this.locatorFor(SkyFormErrorsHarness)();
+  }
+
+  async #getHelpInline(): Promise<SkyHelpInlineHarness> {
```
Blackbaud-SteveBrush: Private methods should go beneath public ones.

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    return await this.#documentRootLocator.locatorFor(
+      SkyColorpickerDropdownHarness.with({ id: overlayId }),
+    )();
+  }
+
+  /**
+   * Gets the `SkyIconHarness` for the colorpicker icon.
+   * @internal
```
Blackbaud-SteveBrush, line 107: Do we even need this method since it's exposing an internal detail of the colorpicker?

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
```diff
@@ -0,0 +1,201 @@
...
+      return await defaultButton?.click();
+    } else if (modernButton) {
+      return await modernButton?.click();
+    }
+  }
+
+  /**
+   * Gets the colorpicker component's aria label.
+   */
+  public async getAriaLabel(): Promise<string | null> {
```
Blackbaud-SteveBrush, line 76: There are 11 `aria-label` attributes being set on the colorpicker. The JSDocs say "colorpicker component's aria label" but it's not clear what that means. Maybe make this method more specific to which `aria-label` you're retrieving and update the JSDocs accordingly?

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
```diff
@@ -0,0 +1,201 @@
...
+   * Gets the colorpicker component's label text.
+   */
+  public async getLabelText(): Promise<string> {
+    return (await (await this.#getLabel()).text()).trim();
+  }
+
+  /**
+   * Gets the colorpicker button's title.
+   */
+  public async getTitle(): Promise<string | null> {
```
Blackbaud-SteveBrush: This is an internal detail of the component and probably shouldn't be leaked to the consumer.

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
```diff
@@ -0,0 +1,201 @@
...
+   * Whether the required error has fired.
+   */
+  public async hasRequiredError(): Promise<boolean> {
+    return (await this.#getFormErrors())?.hasError('required');
+  }
+
+  /**
+   * Whether the colorpicker component's label is hidden.
+   */
+  public async isLabelHidden(): Promise<boolean> {
```
Blackbaud-SteveBrush: We name this `getLabelHidden` in the other harnesses.

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
```diff
@@ -0,0 +1,201 @@
...
+    }
+    if (defaultButton) {
+      return await defaultButton?.click();
+    } else if (modernButton) {
+      return await modernButton?.click();
+    }
+  }
+
+  /**
+   * Gets the colorpicker component's aria label.
```
Blackbaud-SteveBrush: For all your JSDocs comments: when talking about ARIA, it should be all-caps; if you're referencing a specific attribute, put it in ticks so it's clear to the consumer: \`aria-label\`.

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
```diff
@@ -0,0 +1,201 @@
...
+   */
+  public async getAriaLabel(): Promise<string | null> {
+    return (await this.#getButton()).getAttribute('aria-label');
+  }
+
```
Blackbaud-SteveBrush, line 79: Do we have a method to retrieve the `aria-labelledby` value? There's a `labelledBy` input on the colorpicker, so we should probably add one. Maybe take a look at the help inline harness for a similar implementation:
https://github.com/blackbaud/skyux/blob/main/libs/components/help-inline/testing/src/help-inline/help-inline-harness.ts#L80-L90

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
```diff
@@ -0,0 +1,201 @@
...
+    if (!defaultButton && !modernButton) {
+      throw new Error('No reset button found.');
+    }
+    if (defaultButton) {
+      return await defaultButton?.click();
+    } else if (modernButton) {
+      return await modernButton?.click();
+    }
+  }
+
```
Blackbaud-SteveBrush, line 72: There's an input for `showResetButton`, but there's not an easy way to check that it exists or not (without catching an error when clicking on it). Maybe add a `hasResetButton(): Promise<boolean>` method?

libs/components/colorpicker/testing/src/colorpicker/colorpicker-dropdown-harness.ts
```diff
@@ -0,0 +1,153 @@
...
+   * Enters a value into the alpha input box.
+   * @param value A decimal value from 0-1.
+   */
+  public async setAlphaValue(value: string): Promise<void> {
+    const input = await this.getAlphaInput();
+
+    input.setInputValue(value);
+    input.dispatchEvent('input');
+  }
+
```
Blackbaud-SteveBrush, line 95: I'm not seeing methods to capture `outputFormat`, `presetColors`, `alphaChannel`, or `allowTransparency` on the `colorpicker-input.directive`.

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
Blackbaud-SandhyaRajasabeson: The only place we use this is in text editor and could be useful for our internal testing purposes, that's why i added it. Also with the internal tag consumers wouldn't see this in our docs and so wouldnt be using it right?

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
Blackbaud-SandhyaRajasabeson: went off of what the button that triggers opening the colorpicker dropdown is called in the anatomy docs

libs/components/colorpicker/testing/src/colorpicker/colorpicker-dropdown-harness.ts
Blackbaud-SandhyaRajasabeson: i can add functions for allowTransparency and one that returns an array for presetColors but `outputFormat` and `alphaChannel` does not affect the dom for colorpicker

libs/components/colorpicker/testing/src/colorpicker/colorpicker-dropdown-harness.ts
```diff
@@ -0,0 +1,169 @@
...
+      if ((await swatch.getAttribute('aria-label')) === ariaLabel) {
+        return await swatch.click();
+      }
+    }
+  }
+
+  /**
+   * Gets an array of the hex codes of the preset swatches.
+   */
+  public async getPresetSwatches(): Promise<(string | undefined)[]> {
```
Blackbaud-SteveBrush: Small, but we have `clickPresetColorSwatch` above; could we call this method `getPresetColorSwatches`?

Blackbaud-SteveBrush: The return type is an array of `string` or `undefined`. So the result could possibly be `['#fff', undefined]`? Can we ensure only defined values get returned? `Promise<string[]>`

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
```diff
@@ -0,0 +1,211 @@
...
+   * Gets the colorpicker button's `aria-label`.
+   */
+  public async getAriaLabel(): Promise<string | null> {
+    return (await this.#getButton()).getAttribute('aria-label');
+  }
+
+  /**
+   * Gets the colorpicker button's `aria-labelledBy`
+   */
+  public async getAriaLabelledBy(): Promise<string | null> {
```
Blackbaud-SteveBrush: The attribute has a lowercase "b" for "by" (not great, but it is what it is). We can mimic what we're doing in the other harnesses by calling this `getAriaLabelledby` (the JSDoc also needs to use `aria-labelledby`).

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
Blackbaud-SteveBrush: It won't be documented, but it doesn't mean consumers won't use it. I'm inclined to remove this method until it's absolutely needed.

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
Blackbaud-SandhyaRajasabeson: I was trying to have it match what we call the input in our docs

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
Blackbaud-SandhyaRajasabeson: mmmmm id be down to comment it out just so its already done and ready for when we do text editor harness

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
Blackbaud-SteveBrush: The input is named `labelledBy` (does not include "aria"). This method serves to return the value of the native attribute `aria-labelledby`, so we need to rename it accordingly (similar to how we name it in our other harnesses).

**Checkbox, as an example:**
`labelledBy` input: https://github.com/blackbaud/skyux/blob/main/libs/components/forms/src/lib/modules/checkbox/checkbox.component.ts#L85
`getAriaLabelledby` method on the harness: https://github.com/blackbaud/skyux/blob/main/libs/components/forms/testing/src/checkbox/checkbox-harness.ts#L78-L80

libs/components/colorpicker/testing/src/colorpicker/colorpicker-harness.ts
Blackbaud-SteveBrush: I guess that's fair. Keep it.

libs/components/colorpicker/testing/src/colorpicker/colorpicker-dropdown-harness.ts
Blackbaud-SteveBrush: Thanks for fixing the return type of this method. The first question still applies, however.

Blackbaud-SteveBrush (approved):

PR #2785, opened 2024-09-30
feat(components/layout): add test harnesses to fluid grid

Blackbaud-CoreyArcher:
[AB#2195450](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2195450)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/2785/storybook/)

Component Storybooks:

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/2785/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/2785/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/2785/playground/)

libs/components/layout/testing/src/fluid-grid/fluid-grid-harness.ts
```diff
@@ -0,0 +1,62 @@
...
+   * Gets all of the rows in the grid.
+   */
+  public async getRows(): Promise<SkyRowHarness[]> {
+    return await this.#getRows();
+  }
+
+  /**
+   * Whether the fluid grid has additional margins on the sides.
+   */
+  public async hasMargin(): Promise<boolean> {
```
Blackbaud-SandhyaRajasabeson, line 57: is this linked to the `disableMargin` input? if so can we update the jsDocs to say something like 
```Whether the fluid grid has margin enabled.```

File: libs/components/layout/testing/src/fluid-grid/row-harness.spec.ts [hunk 2]
```diff
@@ -0,0 +1,59 @@
...
+
+    const rowHarness: SkyRowHarness = options.dataSkyId
+      ? await loader.getHarness(
+          SkyRowHarness.with({
+            dataSkyId: options.dataSkyId,
+          }),
+        )
+      : await loader.getHarness(SkyRowHarness);
+
+    return { rowHarness, fixture, loader };
```
Blackbaud-SandhyaRajasabeson: this loader is not being used, can we remove it from the return type?

libs/components/layout/testing/src/fluid-grid/column-harness.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    return result || (await this.#getColumnSize(index - 1));
+  }
+
+  // eslint-disable-next-line complexity
+  async #getColumnClass(size: string): Promise<number> {
+    const host = await this.host();
+
+    return (await host.hasClass(`sky-column-${size}-12`))
+      ? 12
```
Blackbaud-SandhyaRajasabeson: Can this not be simplified with a regex to mask out only the value after the last hyphen?

libs/components/layout/testing/src/fluid-grid/column-harness.ts
```diff
@@ -0,0 +1,98 @@
...
+    return this.#getColumnSize(3);
+  }
+
+  async #getColumnSize(index: number): Promise<number> {
+    const size = sizes[index];
+
+    const result = await this.#getColumnClass(size);
+
+    /* istanbul ignore if */
+    if (!result && !index) {
```
Blackbaud-SandhyaRajasabeson, line 61: is this not a testable scenario?

libs/components/layout/testing/src/fluid-grid/fluid-grid-harness.ts
Blackbaud-CoreyArcher: updated

libs/components/layout/testing/src/fluid-grid/row-harness.spec.ts
Blackbaud-CoreyArcher: removed

libs/components/layout/testing/src/fluid-grid/column-harness.ts
Blackbaud-CoreyArcher: I don't think so. The host `TestElement` only has this `hasClass()` method, and doesn't seem to support complex regex matching on them.

libs/components/layout/testing/src/fluid-grid/column-harness.ts
Blackbaud-CoreyArcher: No. There should always be a class to fall back to. If we get a result of 0 after running the function against `xsmall`, which has an index of 0 and should always have a positive value, then something went wrong and without this exception the function would never stop looping.

libs/components/layout/testing/src/fluid-grid/column-harness.ts
Blackbaud-SandhyaRajasabeson: ok i think simplifying this to a for loop that replaces the number makes more sense than this long list of ternaries

Blackbaud-CoreyArcher: done

Blackbaud-SandhyaRajasabeson (approved):

PR #2681, opened 2024-09-04
feat(components/ag-grid): support AG Grid 32.1.0

johnhwhite:
[AB#3067793](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3067793)

PR #2802, opened 2024-10-04
feat(components/core): add `SkyMediaQueryController` to interact with breakpoints in tests

Blackbaud-SteveBrush:
[AB#2195494](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/2195494)

### Testing
Consumers can add the following to their tests to interact with breakpoints:
```
import { provideSkyMediaQueryTesting } from '@skyux/core/testing';

  providers: [provideSkyMediaQueryTesting()]
});

const mediaQueryController = TestBed.inject(SkyMediaQueryTestingController);

mediaQueryController.setBreakpoint('xs');
await mediaQueryController.expectBreakpoint('xs');
```

### Creating responsive hosts

Consumers can apply a container breakpoint observer to an element in their template like this:
```
<div #responsiveHost="skyResponsiveHost" skyResponsiveHost>
  <p>Breakpoint within container: {{ responsiveHost.breakpointChange | async }}</p>
</div>
```
Or on their component's host element:
```
@Component({
  hostDirectives: [SkyResponsiveHostDirective],
  standalone: true,
  styles: `
    :host {
      display: block;
    }
  `,
  template: ` <p>Breakpoint within container: {{ breakpoint() }}</p> `,
})
export class DemoComponent {
  protected breakpoint = toSignal(
    inject(SkyMediaQueryService).breakpointChange,
  );
}

```
For embedded views, the responsive host can pass its injector:
```
<div #responsiveHost="skyResponsiveHost" skyResponsiveHost>
  <ng-container
    [ngTemplateOutlet]="myTemplate"
    [ngTemplateOutletInjector]="responsiveHost.injector"
</div>
```


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/2802/storybook/)

Component Storybooks:

- [a11y-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/a11y-storybook/)
- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/colorpicker-storybook/)
- [core-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/core-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/errors-storybook/)
- [flyout-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/flyout-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/forms-storybook/)
- [help-inline-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/help-inline-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/lookup-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/modals-storybook/)
- [navbar-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/navbar-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/phone-field-storybook/)
- [popovers-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/popovers-storybook/)
- [progress-indicator-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/progress-indicator-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/theme-storybook/)
- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/tiles-storybook/)
- [toast-storybook](https://blackbaud.github.io/skyux-pr-preview/2802/storybooks/toast-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/2802/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/2802/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/2802/playground/)

libs/components/flyout/src/lib/modules/flyout/flyout.component.scss
```diff
@@ -24,6 +24,10 @@ $header-height: 50px;
     .sky-flyout-resize-handle {
       cursor: initial;
     }
+
+    .sky-flyout-header {
+      width: 100%;
```
Blackbaud-SteveBrush, line 29: This fixes a bug with fullscreen flyouts I noticed while working on this feature.

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar-collapsible.pipe.ts
```diff
@@ -9,11 +9,11 @@ import { SkySummaryActionBarType } from './types/summary-action-bar-type';
 export class SkySummaryActionBarCollapsiblePipe implements PipeTransform {
   public transform(
     summaryActionBarType: SkySummaryActionBarType | undefined,
-    mediaBreakpoint: SkyMediaBreakpoints,
+    mediaBreakpoint: SkyMediaBreakpointType | null,
```
Blackbaud-SteveBrush: 👀 I created a new type `SkyMediaBreakpointType`, a string union instead of an enum, which is supremely easier to work with.

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.html
```diff
@@ -1,8 +1,12 @@
+@let isCollapsible =
+  type
+    | skySummaryActionBarCollapsible
+      : (mediaQueryService.breakpointChange | async);
```
Blackbaud-SteveBrush: 👀 I was experiencing some strange change-detection errors when I switched the summary action bar to the new media query override. The fix was to create a public `breakpointChange` observable on the `SkyMediaQueryService` (to replace the `subscribe` method).

libs/components/core/src/lib/modules/media-query/media-query.service.ts
```diff
@@ -35,18 +44,31 @@ export class SkyMediaQueryService implements OnDestroy {
    */
   public static lg = '(min-width: 1200px)';
 
+  /**
+   * Emits when the breakpoint changes.
+   */
+  public get breakpointChange(): Observable<SkyMediaBreakpointType> {
```
Blackbaud-SteveBrush: 👀 New public API: exposes an observable with the new string union `SkyMediaBreakpointType`.

libs/components/core/testing/src/media-query/media-query-testing.service.ts
```diff
@@ -0,0 +1,149 @@
+/* eslint-disable @nx/enforce-module-boundaries */
+import {
```
Blackbaud-SteveBrush: 👀 👇 This file needs special attention.

apps/code-examples/src/app/code-examples/core/media-query/basic/demo.component.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+      providers: [provideSkyMediaQueryTesting()],
+    });
+  });
+
+  it('should change the breakpoint', async () => {
+    const { fixture, mediaController } = setupTest();
+
+    fixture.detectChanges();
+
+    await mediaController.expectBreakpoint('md');
```
Blackbaud-PaulCrowder: Expecting a certain breakpoint before it is ever set by the test seems odd to me. Should consumers expect `md` to be the default breakpoint, or should we make consumers set a breakpoint before making any assertions about it?

libs/components/core/src/lib/modules/media-query/media-query-service-interface.ts
```diff
@@ -0,0 +1,33 @@
+import { Observable, Subscription } from 'rxjs';
+
+import { SkyMediaBreakpointType } from './media-breakpoint-type';
+import { SkyMediaBreakpoints } from './media-breakpoints';
+import { SkyMediaQueryListener } from './media-query-listener';
+
+/**
+ * @internal
+ */
+export interface SkyMediaQueryServiceInterface {
```
Blackbaud-SteveBrush: 👀 This interface is used by `SkyMediaQueryService` and `SkyMediaQueryTestingService` to ensure all public properties/methods are the same.

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.html
Blackbaud-PaulCrowder: This is a good opportunity to get rid of the `skySummaryActionBarCollapsible` pipe and merge its logic with the media query service into a computed signal.

```
#breakpoint = toSignal(inject(SkyMediaQueryService).breakpointChange);

// Push a value to this when the type is detected from the adapter service.
#type = signal<SkySummaryActionBarType | undefined>();

protected readonly isCollapsible = computed(
  () => this.#type() === SkySummaryActionBarType.StandardModal || this.#breakpoint === 'xs')
);
```

```
@if (isCollapsible() && !isSummaryCollapsed) {
  <div class="sky-summary-action-bar-details-collapse">
    <ng-container *ngTemplateOutlet="chevron" />
  </div>
}
```

libs/components/core/src/lib/modules/media-query/media-breakpoint-type.ts
```diff
@@ -0,0 +1,65 @@
...
+    value !== null &&
+    value !== undefined &&
+    SKY_MEDIA_BREAKPOINT_TYPES.includes(value as SkyMediaBreakpointType)
+  );
+}
+/**
+ * The default media breakpoint type.
+ * @internal
+ */
+export const SKY_MEDIA_BREAKPOINT_TYPE_DEFAULT: SkyMediaBreakpointType = 'md';
```
Blackbaud-PaulCrowder: The concept of a default breakpoint is odd to me. When would it be set to a default? When there is no screen? If it's for testing, I'd prefer the consumer have to explicitly set a breakpoint before making any assertions about the screen size.

libs/components/core/src/lib/modules/media-query/provide-media-query-service-override.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { SKY_MEDIA_QUERY_SERVICE_OVERRIDE } from './media-query-service-override.token';
+import { SkyMediaQueryService } from './media-query.service';
+
+/**
+ * Overrides the `SkyMediaQueryService` provider, usually within an element injector.
+ * @internal
+ */
+export function provideSkyMediaQueryServiceOverride(
+  svc: Type<SkyMediaQueryService>,
```
Blackbaud-PaulCrowder: Should this be `Type<SkyMediaQueryServiceInterface>`?

libs/components/core/testing/src/media-query/media-query-testing.service.ts
```diff
@@ -0,0 +1,142 @@
...
+  Subscription,
+  firstValueFrom,
+} from 'rxjs';
+
+import { SkyMediaQueryTestingController } from './media-query-testing.controller';
+
+/**
+ * @internal
+ */
+@Injectable({ providedIn: 'root' })
```
Blackbaud-PaulCrowder: This shouldn't be provided in root, correct? Shouldn't it only be provided by `provideSkyMediaQueryTesting()`?

libs/components/core/testing/src/media-query/media-query-testing.service.ts
```diff
@@ -0,0 +1,142 @@
...
+        if (options?.updateResponsiveClasses) {
+          const el = elementRef.nativeElement;
+
+          this.#removeResponsiveClasses(el);
+          this.#addResponsiveClass(el, breakpoint);
+        }
+      },
+    );
+
+    return this as unknown as SkyResizeObserverMediaQueryService;
```
Blackbaud-PaulCrowder: This makes me think we're overloading the purpose of this service. Why not have a separate `SkyResizeObserverMediaQueryTestingService`?

libs/components/split-view/src/lib/modules/split-view/split-view-drawer.component.ts
```diff
@@ -82,6 +78,9 @@ export class SkySplitViewDrawerComponent
     return this.#_width || this.widthDefault;
   }
 
+  @ViewChild('drawerRef', { static: true })
+  protected drawerRef!: ElementRef;
```
Blackbaud-PaulCrowder: You can use a [signal for this](https://angular.dev/guide/signals/queries):

```
protected drawerRef = viewChild('drawerRef');
```

libs/components/split-view/src/lib/modules/split-view/split-view-drawer.component.ts
```diff
@@ -138,7 +134,9 @@ export class SkySplitViewDrawerComponent
   }
 
   public ngAfterViewInit(): void {
-    this.#updateBreakpoint();
+    this.#mediaQuerySvc.observe(this.drawerRef, {
```
Blackbaud-PaulCrowder: One you make `drawerRef` a signal, you can use an effect to start observing:

```
constructor() {
  effect(() => {
    const drawerRef = this.#drawerRef();

    if (drawerRef) {
      this.#mediaQuerySvc.observe(drawerRef, { updateResponsiveClasses: true });          
    }
  });
}
```

Or if you want similar behavior as `ngAfterViewInit()`, you can use `afterNextRender()` instead:

```
constructor() {
  afterNextRender({
    read: () => this.#mediaQuerySvc.observe(this.drawerRef(), { updateResponsiveClasses: true })
  });
}
```

apps/code-examples/src/app/code-examples/core/media-query/basic/demo.component.spec.ts
Blackbaud-SteveBrush: The default breakpoint is `md`, but I guess that's meaningless since the width of the screen (or container) on init is what matters, here. I'll adjust the test.

libs/components/core/src/lib/modules/media-query/provide-media-query-service-override.ts
Blackbaud-SteveBrush: I could be misunderstanding, but the `Type` interface needs to reference a "class" type. If I change it to the interface, I get a TypeScript error, `Type 'SkyMediaQueryServiceInterface' is missing the following properties from type 'SkyMediaQueryService': #currentSubject, #breakpointChange, #breakpointChangeObs, #currentBreakpoint, and 7 more.ts(2740)` because it's expecting the token value to be an instance of `SkyMediaQueryService`. Is there something I'm missing in my assumptions?

libs/components/core/testing/src/media-query/media-query-testing.service.ts
Blackbaud-SteveBrush: Doh!

libs/components/core/testing/src/media-query/media-query-testing.service.ts
Blackbaud-SteveBrush: For components that use "container queries", we're overloading the `SkyMediaQueryService` injection token with an instance of `SkyResizeObserverMediaQueryService`. In other words, we're never _really_ injecting `SkyResizeObserverMediaQueryService` into our components, we're only overloading the `SkyMediaQueryService` token with the functionality we want (and asserting the type as `SkyResizeObserverMediaQueryService` so we can call its methods).

For unit tests, the `SkyMediaQueryService` token is overridden with the `SkyMediaQueryTestingService`, so it also needs to include the public methods from the resize observer service in the cases that `observe` and `unobserve` are assumed to exist on the overloaded `SkyMediaQueryService`. 

For example, this is how we're overloading the `SkyMediaQueryService` token in our components:

```
providers: [
  {
    provide: SkyMediaQueryService,
    useClass: SkyResizeObserverMediaQueryService
              ^
              This class has public methods that aren't defined in SkyMediaQueryService
  }
]
```

```
// Inject the `SkyMediaQueryService` token, but assert its type as `SkyResizeObserverMediaQueryService`
// so we can access its public methods.
#querySvc = inject(SkyMediaQueryService) as SkyResizeObserverMediaQueryService;
```

Blackbaud-SteveBrush: In hindsight, we should have built the `SkyResizeObserverMediaQueryService` using the same exact methods as the `SkyMediaQueryService` to avoid this confusion.

libs/components/core/src/lib/modules/media-query/media-breakpoint-type.ts
Blackbaud-SteveBrush: Good call. I converted `breakpointChange` to a ReplaySubject (instead of a BehaviorSubject) so it won't require a default value (I made the same change to the testing controller).

libs/components/split-view/src/lib/modules/split-view/split-view-drawer.component.ts
Blackbaud-SteveBrush: This is neat! I didn't know about it.

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.html
Blackbaud-SteveBrush: I like this a lot. 👍

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid-data-manager-adapter.directive.ts
```diff
@@ -411,6 +411,7 @@ export class SkyAgGridDataManagerAdapterDirective
   }
 
   #applyColumnWidths(): void {
+    /* istanbul ignore if: safety check */
```
Blackbaud-SteveBrush, line 414: See comment from John's PR a while ago (this condition was added for tests): https://github.com/blackbaud/skyux/pull/2681/files#r1745689224

libs/components/core/src/lib/modules/breakpoint-observer/container-breakpoint-observer.ts
```diff
@@ -0,0 +1,65 @@
+import { ElementRef, Injectable, inject } from '@angular/core';
```
Blackbaud-SteveBrush, line 1: 👀 👇

libs/components/core/src/lib/modules/breakpoint-observer/media-breakpoint-observer.ts
```diff
@@ -0,0 +1,59 @@
+import { Injectable } from '@angular/core';
```
Blackbaud-SteveBrush, line 1: 👀 👇

libs/components/core/src/lib/modules/breakpoint-observer/provide-breakpoint-observer.ts
```diff
@@ -0,0 +1,30 @@
+import { Provider, Type, inject } from '@angular/core';
```
Blackbaud-SteveBrush, line 1: 👀 👇

libs/components/core/src/lib/modules/breakpoint-observer/responsive-host.directive.ts
```diff
@@ -0,0 +1,41 @@
+import { Directive, ElementRef, inject } from '@angular/core';
```
Blackbaud-SteveBrush, line 1: 👀 👇

libs/components/core/src/lib/modules/media-query/media-query.service.ts
```diff
@@ -1,16 +1,33 @@
-import { Injectable, NgZone, OnDestroy } from '@angular/core';
+import { Injectable, NgZone, OnDestroy, inject } from '@angular/core';
```
Blackbaud-SteveBrush, line 1: 👀 👇

libs/components/core/testing/src/media-query/media-query-testing-controller.ts
```diff
@@ -0,0 +1,46 @@
+import { Injectable, inject } from '@angular/core';
```
Blackbaud-SteveBrush, line 1: 👀 👇

libs/components/core/testing/src/media-query/provide-media-query-testing.ts
```diff
@@ -0,0 +1,33 @@
+import { Provider } from '@angular/core';
```
Blackbaud-SteveBrush, line 2: 👀 👇

libs/components/split-view/src/lib/modules/split-view/split-view-workspace.component.ts
```diff
@@ -1,22 +1,11 @@
 import {
```
Blackbaud-SteveBrush, line 1: 👀 👇 Split view saw the greatest number of changes since it was doing some fairly "boutique" breakpoint handling.

libs/components/core/src/lib/modules/breakpoint-observer/responsive-host.directive.ts
```diff
@@ -0,0 +1,58 @@
+import { Directive, ElementRef, Injector, inject } from '@angular/core';
```
Blackbaud-SteveBrush, line 1: 👀 👇

libs/components/core/src/lib/modules/breakpoint-observer/container-breakpoint-observer.ts
```diff
@@ -0,0 +1,84 @@
+import {
```
Blackbaud-SteveBrush, line 1: 👀 👇

libs/components/core/src/lib/modules/breakpoint-observer/media-breakpoint-observer.ts
```diff
@@ -0,0 +1,65 @@
+import { Injectable, OnDestroy } from '@angular/core';
```
Blackbaud-SteveBrush, line 1: 👀 👇

apps/code-examples/src/app/code-examples/core/media-query/responsive-host/demo.component.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [provideSkyMediaQueryTesting()],
+    });
+  });
+
+  it('should change the breakpoint', async () => {
+    const { fixture, mediaController } = setupTest();
+
+    const containerWithHostDirective = fixture.debugElement.query(
```
Blackbaud-PaulCrowder: Use `nativeElement` instead of `debugElement` here per [Angular's guidance](https://angular.dev/guide/testing/components-basics#bycss).

>When you're filtering by CSS selector and only testing properties of a browser's native element, the By.css approach might be overkill.
>
>It's often more straightforward and clear to filter with a standard HTMLElement method such as querySelector() or querySelectorAll().

```
const containerWithHostDirective = (fixture.nativeElement as HTMLElement).querySelector('[data-sky-id="container-w-host-directive"]');
```

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.ts
```diff
@@ -145,6 +164,7 @@ export class SkySummaryActionBarComponent implements AfterViewInit, OnDestroy {
 
   public onDirectionChange(direction: string): void {
     this.slideDirection = direction;
+    this.#changeDetector.markForCheck();
```
Blackbaud-PaulCrowder: This function is called as a result of an event handler, so in theory `markForCheck()` should not be needed. Can you make `slideDirection` a signal?

libs/components/core/src/lib/modules/breakpoint-observer/breakpoint-utils.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/**
+ * Transforms a `SkyBreakpoint` value to `SkyMediaBreakpoints`.
+ * @internal
+ */
+export function toSkyMediaBreakpoints(
+  breakpoint: SkyBreakpoint,
+): SkyMediaBreakpoints {
+  const breakpointsMap = new Map<SkyBreakpoint, SkyMediaBreakpoints>([
```
Blackbaud-PaulCrowder: These don't change between calls to `toSkyMediaBreakpoints()`. Move to the top of this file so it can be reused.

libs/components/core/src/lib/modules/breakpoint-observer/breakpoint-utils.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/**
+ * Transforms a `SkyMediaBreakpoints` value to `SkyBreakpoint`.
+ * @internal
+ */
+export function toSkyBreakpoint(
+  breakpoint: SkyMediaBreakpoints,
+): SkyBreakpoint {
+  const breakpointsMap = new Map<SkyMediaBreakpoints, SkyBreakpoint>([
```
Blackbaud-PaulCrowder: These don't change between calls to `toSkyBreakpoint()`. Move to the top of this file so it can be reused.

libs/components/core/src/lib/modules/breakpoint-observer/container-breakpoint-observer.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import { Observable, ReplaySubject } from 'rxjs';
+
+import { SkyResizeObserverService } from '../resize-observer/resize-observer.service';
+
+import { SkyBreakpoint } from './breakpoint';
+import { SkyBreakpointObserver } from './breakpoint-observer';
+
+const QUERIES = new Map<SkyBreakpoint, (width: number) => boolean>([
+  ['xs', (w): boolean => w > 0 && w <= 767],
```
Blackbaud-PaulCrowder: Do you need the `: boolean` return type, or can it be inferred from the `Map` declaration (or array declaration after the suggested refactor above)? Also, I'd make the parameter names match by using `width` instead of `w` for readability.

libs/components/core/src/lib/modules/breakpoint-observer/container-breakpoint-observer.ts
```diff
@@ -0,0 +1,84 @@
...
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
+
+import { Observable, ReplaySubject } from 'rxjs';
+
+import { SkyResizeObserverService } from '../resize-observer/resize-observer.service';
+
+import { SkyBreakpoint } from './breakpoint';
+import { SkyBreakpointObserver } from './breakpoint-observer';
+
+const QUERIES = new Map<SkyBreakpoint, (width: number) => boolean>([
```
Blackbaud-PaulCrowder: I don't see anywhere that values are retrieved from `QUERIES` by its key. The one place it's referenced loops over the entries, so you can avoid the overhead of `Map` and just make this an array instead.

libs/components/core/src/lib/modules/breakpoint-observer/media-breakpoint-observer.ts
```diff
@@ -0,0 +1,65 @@
...
+  implements OnDestroy, SkyBreakpointObserver
+{
+  public get breakpointChange(): Observable<SkyBreakpoint> {
+    return this.#breakpointChangeObs;
+  }
+
+  readonly #breakpointChange = new ReplaySubject<SkyBreakpoint>(1);
+  readonly #breakpointChangeObs = this.#breakpointChange.asObservable();
+
+  #listeners = new Map<MediaQueryList, (evt: MediaQueryListEvent) => void>();
```
Blackbaud-PaulCrowder, line 32: Why does this exist? Items are only ever added to it; those values are never referenced later. Did you mean to loop over them and remove each listener on destroy?

libs/components/core/src/lib/modules/breakpoint-observer/responsive-host.directive.ts
```diff
@@ -0,0 +1,58 @@
...
+   */
+  public get injector(): Injector {
+    return this.#injector;
+  }
+
+  constructor() {
+    const adapter = inject(SkyCoreAdapterService);
+    const elementRef = inject(ElementRef);
+
+    this.breakpointChange.subscribe((breakpoint) => {
```
Blackbaud-PaulCrowder, line 54: Does this need a `takeUntilDestroyed()`? You should't have to use `DestroyRef` since it's called in the construction phase.

libs/components/core/testing/src/media-query/media-query-testing-controller.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Throws an error if the current breakpoint does not match the expected breakpoint.
+   */
+  public async expectBreakpoint(
+    expectedBreakpoint: SkyBreakpoint,
+  ): Promise<void> {
+    if (!this.#breakpoint) {
+      throw new Error(
+        `A media breakpoint has not been set. Call \`setBreakpoint()\` and try again.`,
```
Blackbaud-PaulCrowder: Is calling `setBreakpoint()` on this class the only way to change the current breakpoint? If so, why have the `expectBreakpoint()` method?

libs/components/lookup/src/lib/modules/search/search.component.ts
```diff
@@ -199,30 +197,39 @@ export class SkySearchComponent implements OnDestroy, OnInit, OnChanges {
...
 
     this.contentInfoObs = this.#contentInfoProvider?.getInfo();
+
+    effect(() => {
+      const breakpoint = this.#breakpoint();
+
+      untracked(() => {
+        if (this.#searchShouldCollapse()) {
+          if (breakpoint === 'xs') {
+            this.inputAnimate = INPUT_HIDDEN_STATE;
```
Blackbaud-PaulCrowder: Setting properties in an `effect()` seems like an anti-pattern to me. Can `inputAnimate` and `mobileSearchShown` be computed signals instead?

libs/components/split-view/src/lib/modules/split-view/split-view-drawer.component.ts
```diff
@@ -28,20 +23,14 @@ let skySplitViewNextId = 0;
...
+  standalone: true,
   styleUrls: ['split-view-drawer.component.scss'],
-  changeDetection: ChangeDetectionStrategy.OnPush,
-  providers: [
-    {
-      provide: SkyMediaQueryService,
-      useExisting: SkySplitViewMediaQueryService,
-    },
-  ],
+  templateUrl: 'split-view-drawer.component.html',
```
Blackbaud-PaulCrowder: `templateUrl` and `styleUrls` are missing the `./` prefix. Didn't we hit issues a while back with this? I'm kind of surprised it works. Other split view files also have this problem.

libs/components/tiles/src/lib/modules/tiles/tile-dashboard/tile-dashboard.component.ts
```diff
@@ -96,61 +99,53 @@ export class SkyTileDashboardComponent implements AfterViewInit, OnDestroy {
 
   public moveInstructionsId: string;
 
+  protected layoutClassName = computed(() => {
+    const breakpoint = this.#breakpoint();
+
+    return untracked(() => {
```
Blackbaud-PaulCrowder: Why does this use `untracked()`? it doesn't reference any signals.

libs/components/tiles/src/lib/modules/tiles/tile-dashboard/tile-dashboard.service.ts
```diff
@@ -49,17 +52,35 @@ export class SkyTileDashboardService {
...
+  readonly #dynamicComponentService = inject(SkyDynamicComponentService);
+  readonly #uiConfigService = inject(SkyUIConfigService);
+
+  #mode: SkyTileDashboardColumnMode | undefined;
 
   constructor() {
+    effect(() => {
+      const breakpoint = this.#breakpoint();
+
+      untracked(() => {
```
Blackbaud-PaulCrowder: Same issue with `untracked()` as above.

libs/components/tiles/src/lib/modules/tiles/tile-dashboard/tile-dashboard.service.ts
```diff
@@ -49,17 +52,35 @@ export class SkyTileDashboardService {
...
+        const mode =
+          breakpoint === 'xs' || breakpoint === 'sm' ? 'single' : 'multi';
+
+        if (mode !== this.#mode) {
+          this.#mode = mode;
```
Blackbaud-PaulCrowder: Another instance of setting a property in an effect. See if this can be a computed signal instead.

apps/code-examples/src/app/code-examples/core/media-query/responsive-host/demo.component.spec.ts
Blackbaud-SteveBrush: Interesting; that's new to me.

libs/components/tiles/src/lib/modules/tiles/tile-dashboard/tile-dashboard.component.ts
Blackbaud-SteveBrush: I misunderstood the use-case; I thought it marked code that _doesn't_ include any signals. Reverted.

libs/components/split-view/src/lib/modules/split-view/split-view-drawer.component.ts
Blackbaud-SteveBrush: Yeah, strange. I updated all the paths.

libs/components/core/testing/src/media-query/media-query-testing-controller.ts
Blackbaud-SteveBrush: Lol, good point. I'll remove it.

libs/components/core/src/lib/modules/breakpoint-observer/responsive-host.directive.ts
Blackbaud-SteveBrush: I didn't add one because it's pointing to the `breakpointChange` observable on the `SkyMediaQueryService`, which has a `ngOnDestroy` call to complete the underlying subject. Since the `SkyMediaQueryService` is being provided by the directive, the service will be destroyed when the directive is destroyed. Am I missing another scenario?

libs/components/core/src/lib/modules/breakpoint-observer/media-breakpoint-observer.ts
Blackbaud-SteveBrush: YES, good catch. /smh

libs/components/core/src/lib/modules/breakpoint-observer/container-breakpoint-observer.ts
Blackbaud-SteveBrush: Good call.

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.ts
Blackbaud-SteveBrush: Yes, definitely.

libs/components/lookup/src/lib/modules/search/search.component.ts
Blackbaud-SteveBrush: I spent a day trying to refactor search to make `inputAnimate` and `mobileSearchShown` signals, but the lift to get this done might be too much for now. I ended up reverting the `effect` implementation and making only the absolute necessary changes to adopt the new `breakpointChange` observable.

libs/components/tiles/src/lib/modules/tiles/tile-dashboard/tile-dashboard.service.ts
Blackbaud-SteveBrush: Done.

apps/code-examples/src/app/code-examples/core/media-query/basic/demo.component.spec.ts
```diff
@@ -24,25 +23,19 @@ describe('Media query demo', () => {
...
+
     mediaController.setBreakpoint('xs');
     fixture.detectChanges();
 
-    await mediaController.expectBreakpoint('xs');
-
-    expect(
-      fixture.debugElement.query(By.css('.my-nav-mobile')).nativeElement,
-    ).toBeDefined();
+    expect(el.querySelector('.my-nav-mobile')).toBeDefined();
```
Blackbaud-PaulCrowder: Doesn't `querySelector()` return `null` when no match is found, therefore causing `toBeDefined()` to always pass since it [only checks that the value is not undefined](https://jasmine.github.io/api/2.6/matchers.html)? This should probably be `toBeTruthy()` instead.

>toBeDefined()
[expect](https://jasmine.github.io/api/2.6/global.html#expect) the actual value to be defined. (Not undefined)
>
>Example
>expect(result).toBeDefined();

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.ts
```diff
@@ -210,7 +209,7 @@ export class SkySummaryActionBarComponent implements AfterViewInit, OnDestroy {
       (args: SkyMediaBreakpoints) => {
         if (args !== SkyMediaBreakpoints.xs) {
           this.isSummaryCollapsed = false;
```
Blackbaud-PaulCrowder: Is this one of those cases where it would cause too much refactoring if changed to a signal? If it could be a signal then you could get rid of the `detectChanges()` call (though shouldn't that have been `markForCheck()`? This code is weird).

libs/components/lookup/src/lib/modules/search/search.component.ts
```diff
@@ -211,25 +211,17 @@ export class SkySearchComponent implements OnDestroy, OnInit, OnChanges {
...
-      });
-    });
   }
 
   public ngOnInit(): void {
+    if (this.#searchShouldCollapse()) {
+      this.breakpointSubscription =
+        this.#mediaQuerySvc.breakpointChange.subscribe((breakpoint) => {
+          this.#mediaQueryCallback(breakpoint);
+          this.#changeRef.detectChanges();
```
Blackbaud-PaulCrowder: Should this be `markForCheck()`?

libs/components/lookup/src/lib/modules/search/search.component.ts
```diff
@@ -211,25 +211,17 @@ export class SkySearchComponent implements OnDestroy, OnInit, OnChanges {
...
-          }
-        }
-      });
-    });
   }
 
   public ngOnInit(): void {
+    if (this.#searchShouldCollapse()) {
+      this.breakpointSubscription =
+        this.#mediaQuerySvc.breakpointChange.subscribe((breakpoint) => {
```
Blackbaud-PaulCrowder: Use `takeUntilDestroyed()` here.

libs/components/action-bars/src/lib/modules/summary-action-bar/summary-action-bar.component.ts
Blackbaud-SteveBrush: I converted this to a signal, and was able to remove one of the `markForCheck()` calls. However, when I removed (or replaced) the `detectChanges()` call here, the unit tests failed. 🤷‍♂️

libs/components/lookup/src/lib/modules/search/search.component.ts
Blackbaud-SteveBrush: Removed!

libs/components/lookup/src/lib/modules/search/search.component.ts
Blackbaud-SteveBrush: Done.

Blackbaud-PaulCrowder (approved):

PR #2885, opened 2024-11-01
feat(components/datetime): add date range picker harness

Blackbaud-SandhyaRajasabeson:
[AB#2195414](https://dev.azure.com/blackbaud/Products/_boards/board/t/SKY%20UX%20Program/Stories/?workitem=2195414)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/2885/storybook/)

Component Storybooks:

- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/colorpicker-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/errors-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/forms-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/lookup-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/phone-field-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/2885/storybooks/theme-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/2885/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/2885/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/2885/playground/)

libs/components/datetime/testing/src/datepicker/datepicker-harness-input.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export class SkyDatepickerInputHarness extends SkyInputHarness {
+  /**
+   * @internal
+   */
+  public static hostSelector = '[skyDatepickerInput]';
+
+  /**
+   * Sets the value of the input. The value will be set by simulating key
+   * presses that correspond to the given value and dispatching event changes.
```
Blackbaud-SteveBrush: We shouldn't leak how we're implementing the value (since those details could change at any time). We're good to be terse with, "Sets the value of the input".

libs/components/datetime/testing/src/date-range-picker/date-range-picker-harness.ts
```diff
@@ -0,0 +1,194 @@
...
+  ): Promise<void> {
+    const select = await this.#getCalculatorSelect();
+    return select.selectOptions(calculatorId);
+  }
+
+  /**
+   * Sets the end date.
+   * @param newDate date input as a formatted string.
+   */
+  public async setEndDate(newDate: string): Promise<void> {
```
Blackbaud-SteveBrush: `setEndDateValue`, to mimic the `setValue` input method.

apps/code-examples/src/app/code-examples/datetime/date-range-picker/basic/demo.component.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  it('should throw an error if a weekend is selected', async () => {
+    const { harness } = await setupTest({
+      dataSkyId: 'last-donation',
+    });
+
+    await harness.selectCalculator(SkyDateRangeCalculatorId.SpecificRange);
+    await harness.setStartDate(
+      new Date('05/07/2022').toLocaleDateString('en-us', {
```
Blackbaud-SteveBrush: Why won't a simple `'05/07/2022'` work for this?

apps/code-examples/src/app/code-examples/datetime/date-range-picker/basic/demo.component.ts
```diff
@@ -6,9 +6,12 @@ import {
   FormGroup,
   FormsModule,
   ReactiveFormsModule,
+  ValidationErrors,
 } from '@angular/forms';
 import { SkyDateRangePickerModule } from '@skyux/datetime';
 
+import { isDate } from 'moment';
```
Blackbaud-SteveBrush: We shouldn't expose `moment` to the code examples because it's an internal detail. It's not perfect, but for the sake of the code example could we use `date instanceof Date` instead?

apps/code-examples/src/app/code-examples/datetime/date-range-picker/basic/demo.component.ts
```diff
@@ -26,11 +29,37 @@ export class DemoComponent {
...
+          if (isDate(endDate)) {
+            return endDate.getDay() === 6 || endDate.getDay() === 0
+              ? { endDateWeekend: true }
+              : null;
+          }
+          return null;
+        },
+      ],
+    });
+    if (this.disabled) {
```
Blackbaud-SteveBrush: The `disabled` property isn't used by the template, or anywhere else. Is it needed for the code example? Can we just remove it?

apps/code-examples/src/app/code-examples/datetime/date-range-picker/basic/demo.component.ts
```diff
@@ -26,11 +29,37 @@ export class DemoComponent {
   protected hintText =
     'Donations received today are updated at the top of each hour.';
   protected labelText = 'Last donation';
+  protected lastDonation: FormControl<string | null>;
   protected required = true;
 
   constructor() {
+    this.lastDonation = new FormControl('', {
+      validators: [
```
Blackbaud-SteveBrush: For code example custom validators, I think it's helpful to create a function outside of the scope of the class to communicate to the consumer that 1) the validator does not require `this` context, and 2) validators are designed to be reused between form controls. (Take a look at the [datepicker's basic demo](https://github.com/blackbaud/skyux/blob/main/apps/code-examples/src/app/code-examples/datetime/datepicker/basic/demo.component.ts#L20-L31) for an example of this.) For the date range picker code example, it looks like the logic is shared between start and end date, so maybe you could evaluate both in the same validate function. 
```
function dateRangeExcludesWeekend(
  control: AbstractControl,
): ValidationErrors | null {
  const startDate = control.value.startDate;
  const endDate = control.value.endDate;

  const isWeekend = (value: unknown): boolean => {
    return (
      value instanceof Date && (value.getDay() === 6 || value.getDay() === 0)
    );
  };

  if (isWeekend(startDate)) {
    return { startDateWeekend: true };
  }

  if (isWeekend(endDate)) {
    return { endDateWeekend: true };
  }

  return null;
}
```
This can be added to the form control like this:
```
this.lastDonation = new FormControl('', {
  validators: [dateRangeExcludesWeekend],
});
```

apps/code-examples/src/app/code-examples/datetime/date-range-picker/basic/demo.component.ts
```diff
@@ -26,11 +29,37 @@ export class DemoComponent {
...
+          return null;
+        },
+        (control): ValidationErrors | null => {
+          const endDate = control.value.endDate;
+          if (isDate(endDate)) {
+            return endDate.getDay() === 6 || endDate.getDay() === 0
+              ? { endDateWeekend: true }
+              : null;
+          }
+          return null;
```
Blackbaud-SteveBrush: Something I noticed when playing with the code example: when I selected a weekend value, the custom error showed up, but the form controls didn't have the red borders around them.

libs/components/datetime/testing/src/date-range-picker/date-range-picker-harness.spec.ts
```diff
@@ -0,0 +1,324 @@
...
+    await expectAsync(dateRangePickerHarness.isStacked()).toBeResolvedTo(true);
+  });
+
+  it('should get whether start date required error is thrown', async () => {
+    const { dateRangePickerHarness, fixture } = await setupTest();
+
+    await dateRangePickerHarness.selectCalculator(
+      SkyDateRangeCalculatorId.SpecificRange,
+    );
+    fixture.componentInstance.myForm.markAllAsTouched();
```
Blackbaud-SteveBrush, line 185: We should call `blur()` inside the `selectCalculator()` method so that consumers don't need to call `markAllAsTouched()`.

libs/components/datetime/testing/src/date-range-picker/date-range-picker-harness.ts
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  /**
+   * Selects the specified calculator.
+   */
+  public async selectCalculator(
+    calculatorId: SkyDateRangeCalculatorId,
+  ): Promise<void> {
+    const select = await this.#getCalculatorSelect();
+    return select.selectOptions(calculatorId);
```
Blackbaud-SteveBrush: Check all methods and make sure they're returning the result of the awaited promise, instead of the promise itself. This will give us a proper error stack when debugging issues.
```
// Instead of...
return select.selectOptions(calculatorId);
// Do...
return await select.selectOptions(calculatorId);
```
Take a look at [this StackBlitz](https://stackblitz.com/edit/vitejs-vite-z4xrwj?file=main.js&terminal=dev) which demonstrates the issue.

libs/components/datetime/testing/src/date-range-picker/date-range-picker-harness.ts
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * Harness for interacting with date range picker components in tests.
+ */
+export class SkyDateRangePickerHarness extends SkyComponentHarness {
+  /**
+   * @internal
+   */
+  public static hostSelector = 'sky-date-range-picker';
+
+  #getStartDateInput = this.locatorFor(
```
Blackbaud-SteveBrush: Initially, I thought this was pointing to the `input` element, but it's actually the input box harness. Can we be explicit in the name for these and suffix with `InputBoxHarness`? e.g. `#getStartDateInputBoxHarness`

libs/components/datetime/testing/src/date-range-picker/date-range-picker-harness.ts
```diff
@@ -0,0 +1,194 @@
...
+    SkyInputBoxHarness.with({
+      ancestor: '.sky-date-range-picker-start-date',
+    }),
+  );
+  #getEndDateInput = this.locatorFor(
+    SkyInputBoxHarness.with({
+      ancestor: '.sky-date-range-picker-end-date',
+    }),
+  );
+  #getCalculatorSelectInput = this.locatorFor(
```
Blackbaud-SteveBrush: We use `calculatorId` everywhere else; let's be consistent and name this `#getCalculatorIdInputBoxHarness`.

libs/components/datetime/testing/src/date-range-picker/date-range-picker-harness.ts
```diff
@@ -0,0 +1,194 @@
...
+    SkyInputBoxHarness.with({
+      ancestor: '.sky-date-range-picker-select-calculator',
+    }),
+  );
+  #getCalculatorSelect = this.locatorFor(
```
Blackbaud-SteveBrush: This is only used once in the harness, so we can probably put it inside the `selectCalculator` method, and use its input box harness to query the `select` element:
```
const select = await (
  await this.#getCalculatorIdInputBoxHarness()
).querySelector('select');
```

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.html
```diff
@@ -91,7 +93,7 @@
     </sky-input-box>
   </div>
   <div
-    class="sky-date-range-picker-form-group sky-date-range-datepicker-wrapper"
+    class="sky-date-range-picker-form-group sky-date-range-datepicker-wrapper sky-date-range-picker-end-date"
```
Blackbaud-SteveBrush, line 96: Instead of adding classes just for the harness, could you check for "hidden" by checking the required state instead? The datepicker inputs are marked as "required" when they're visible.
```
public async isStartDateVisible(): Promise<boolean> {
  const input = await (
    await this.#getStartDateInputBoxHarness()
  ).querySelector('input');

  const required = await input?.getProperty('required');

  return required === true;
}
```

apps/code-examples/src/app/code-examples/datetime/date-range-picker/basic/demo.component.spec.ts
Blackbaud-SandhyaRajasabeson: I wanted to show consumers how to do this if they wanted to, there is a simpler version right below it. bc they cant just send a `new Date('05/07/2022')`

Blackbaud-SandhyaRajasabeson: further down there is a simpler version of this

apps/code-examples/src/app/code-examples/datetime/date-range-picker/basic/demo.component.ts
Blackbaud-SandhyaRajasabeson: showing consumers how to set disabled for a reactive date range picker was already there in the code example, i didnt want to remove it

apps/code-examples/src/app/code-examples/datetime/date-range-picker/basic/demo.component.ts
Blackbaud-SandhyaRajasabeson: i just noticed that too 😞 has to be a bug, ill file it

libs/components/datetime/testing/src/date-range-picker/date-range-picker-harness.spec.ts
Blackbaud-SandhyaRajasabeson: we did so for set value bc consumers would not know that we do not run check for validation unless we blurred. In this scenario im calling markAllAsTouched so that the required errors would throw on the start and end date pickers. So i dont think we should call blur?

Though when a consumer does select a calculator from the dropdown the select does get blurred and causes the required errors to show up anyways....

libs/components/datetime/src/lib/modules/date-range-picker/date-range-picker.component.html
Blackbaud-SandhyaRajasabeson: this class i added is also to get the individual datepickers to set the value, not just to check for hidden

libs/components/datetime/testing/src/date-range-picker/date-range-picker-harness.spec.ts
Blackbaud-PaulCrowder: >Though when a consumer does select a calculator from the dropdown the select does get blurred and causes the required errors to show up anyways....

Should we wait until that bug is fixed to before merging this then?

Blackbaud-SandhyaRajasabeson: the datepicker bug about them being marked as touched when moving to the calendar wouldnt affect this. Honestly I feel more strongly that we shouldn't blur inside the `selectCalculator` method now bc if all a consumer does is switch between two calculators, date range picker does not become blurred. 

in the demo: switch to a calculator with datepickers, the datepickers do not show the required error until the consumer blurs off.

(I know this ^ is different than my earlier comment, no idea what i was on 3 days ago. different sandhya)

apps/code-examples/src/app/code-examples/datetime/date-range-picker/basic/demo.component.ts
Blackbaud-SandhyaRajasabeson: This is a known bug and is being addressed here: https://dev.azure.com/blackbaud/Products/_boards/board/t/SKY%20UX%20Program/Stories/?workitem=3140330

Blackbaud-PaulCrowder (approved):

PR #3029, opened 2025-01-14
feat(components/forms): file drop harness

Blackbaud-SandhyaRajasabeson:
[AB#2195438](https://dev.azure.com/blackbaud/Products/_boards/board/t/SKY%20UX%20Program/Stories/?workitem=2195438)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/3029/storybook/)

Component Storybooks:

- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/colorpicker-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/errors-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/forms-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/lookup-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/phone-field-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/3029/storybooks/theme-storybook/)

Apps:

- [code-examples](https://blackbaud.github.io/skyux-pr-preview/3029/code-examples/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/3029/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/3029/playground/)

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-item-harness-filters.ts
```diff
@@ -0,0 +1,13 @@
+import { SkyHarnessFilters } from '@skyux/core/testing';
+
+/**
+ * A set of criteria that can be used to filter a list of `SkyFileItemHarness` instances.
+ * @internal
```
Blackbaud-SteveBrush: This class is used in the `SkyFileItemHarness` `public with()` method, so it should be documented.

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-item-harness.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Gets a `HarnessPredicate` that can be used to search for a
+   * `SkyFileItemHarness` that meets certain criteria.
+   */
+  public static with(
+    filters: SkyFileItemHarnessFilters,
+  ): HarnessPredicate<SkyFileItemHarness> {
+    return new HarnessPredicate(SkyFileItemHarness, filters).addOption(
+      'name',
```
Blackbaud-SteveBrush: Can we use `'fileName'` so that it matches the name of the property in the filter?

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-drop-link-upload-harness.ts
```diff
@@ -0,0 +1,64 @@
...
+   * Gets the hint text.
+   */
+  public async getHintText(): Promise<string | undefined> {
+    return await (await this.#inputBoxHarness()).getHintText();
+  }
+
+  /**
+   * Gets the input harness for upload link.
+   */
+  public async getInput(): Promise<SkyFileDropLinkUploadInputHarness> {
```
Blackbaud-SteveBrush: I'd make this method private `#getInput` to avoid confusion on the consumer's end.

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-drop-link-upload-harness.ts
```diff
@@ -0,0 +1,64 @@
+import { SkyComponentHarness } from '@skyux/core/testing';
+
+import { SkyInputBoxHarness } from '../../input-box/input-box-harness';
+
+import { SkyFileDropLinkUploadInputHarness } from './file-drop-link-upload-input-harness';
+
+/**
+ * Harness for interacting with a file drop link upload component in tests.
```
Blackbaud-SteveBrush: The "link upload" feature isn't really a component. Maybe we reword this to, "Harness for interacting with the link upload feature in tests."?

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-drop-harness.ts
```diff
@@ -25,26 +32,201 @@ export class SkyFileDropHarness extends SkyComponentHarness {
...
+   * Whether file drop has stacked enabled.
+   */
+  public async isStacked(): Promise<boolean> {
+    return await (await this.host()).hasClass('sky-form-field-stacked');
+  }
+
+  /**
+   * Loads multiple files through the file drop.
+   */
+  public async uploadFiles(files: File[] | null): Promise<void> {
```
Blackbaud-SteveBrush: We had a similar conversation with the file attachment harness around the "upload" naming. We're not really uploading anything; also, the public methods should reflect user actions. What if you used `dropFiles` for the public method (looks like you already have a similar `dropFile` method)?

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-drop-harness.ts
```diff
@@ -25,26 +32,201 @@ export class SkyFileDropHarness extends SkyComponentHarness {
...
+    return await this.#dropFiles(files);
+  }
+
+  /**
+   * Uploads a link.
+   */
+  public async uploadLink(link: string): Promise<void> {
```
Blackbaud-SteveBrush: I feel like it would be confusing to the developer to have this method, and a method to return the link upload harness that does the same thing. I would create a single way to do things, so we don't end up with offshoot implementations.

I imagine that it's perfectly fine for the tester to write:
```
const linkUpload = await fileDropHarness.getLinkUpload();

await linkUpload.enterText('https://foo.com/file.txt');
await linkUpload.clickDoneButton();
```

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-drop-link-upload-harness.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (await this.isDisabled()) {
+      throw new Error('Done button is disabled and cannot be clicked.');
+    }
+    return await (await this.#button()).click();
+  }
+
+  /**
+   * Sets the input value.
+   */
+  public async enterLink(link: string): Promise<void> {
```
Blackbaud-SteveBrush: We name this `enterText` on our other harnesses.

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-drop-harness.ts
```diff
@@ -25,26 +32,201 @@ export class SkyFileDropHarness extends SkyComponentHarness {
...
+   * Gets the label text.
+   */
+  public async getLabelText(): Promise<string> {
+    return (await (await this.#label()).text()).trim();
+  }
+
+  /**
+   * Gets the link upload harness.
+   */
+  public async getLinkUpload(): Promise<SkyFileDropLinkUploadHarness> {
```
Blackbaud-SteveBrush: Can we be a bit more verbose so it's clearer what's being returned? In the colorpicker harness we have a method named `getColorpickerDropdownHarness`; maybe we could go with `getLinkUploadHarness`.

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-drop-harness.ts
```diff
@@ -25,26 +32,201 @@ export class SkyFileDropHarness extends SkyComponentHarness {
     return SkyFileDropHarness.getDataSkyIdPredicate(filters);
   }
 
+  /**
+   * Clicks the file upload button.
+   */
+  public async clickFileUploadButton(): Promise<void> {
```
Blackbaud-SteveBrush: From our consumers' perspective, I don't think it's obvious what the "file upload button" is, since the only thing in the file drop component that looks like a traditional button is the "Done" button. I feel a better name would be "drop target" since that's the common nomenclature used for this feature outside of SKY UX. Could we rename this method to `clickDropTarget` to fit that reasoning?

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-drop-harness.ts
```diff
@@ -25,26 +32,201 @@ export class SkyFileDropHarness extends SkyComponentHarness {
...
+   * Gets the accepted file types.
+   */
+  public async getAcceptedTypes(): Promise<string | null> {
+    return await (await this.#input()).getAttribute('accept');
+  }
+
+  /**
+   * Gets the aria-label for the file upload button.
+   */
+  public async getFileUploadAriaLabel(): Promise<string | null> {
```
Blackbaud-SteveBrush, line 73: I think we're safe to name this `getAriaLabel` since the context is the file drop component as a whole. The `getHintText` shares a similar context.

libs/components/forms/src/lib/modules/file-attachment/file-drop/file-drop.component.html
```diff
@@ -202,6 +202,7 @@
   [dirty]="ngControl?.dirty"
   [errors]="ngControl?.errors"
 >
+  <ng-content select="sky-form-error" />
```
Blackbaud-SteveBrush, line 205: Nice!

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-drop-link-upload-harness.ts
Blackbaud-SandhyaRajasabeson: I kept this public so that they would have access to the `SkyInputHarness` functions from `SkyFileDropLinkUploadInputHarness`

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-drop-harness.ts
Blackbaud-SandhyaRajasabeson: So this was something i went back and forth on. Ideally I would like to have a `loadFile` `loadFiles` `dropFile` and `dropFiles` function. However currently we can only dispatch the `drop` event and not the `load` event. I want to provide consumers functions and documentation in a way that we can allow for `load` events to be allowed through the harness in the future. 

How about we just have a `loadFile` and `loadFiles` functions and we dont tell them whether we use `drop` or `load` so that essentially we can change how we implement those functions internally in the future? I can't really think of why consumers would need to know how we accomplish loading their files into the file drop.

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-drop-harness.ts
Blackbaud-SandhyaRajasabeson: For file drop inputs there is a `fileUploadAriaLabel` input and a `linkUploadAriaLabel` input. I think it would be better to differentiate here so consumers know exactly what this gets. 

However to get the `linkUploadAriaLabel` they would have to get the `LinkUploadHarness` and then call `getAriaLabel` on it 🤔 Would it make sense to add a helper function called `getLinkUploadAriaLabel` to make it clear that there are two?

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-drop-harness.ts
Blackbaud-SteveBrush: I'm okay with that. 👍

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-drop-link-upload-harness.ts
Blackbaud-SteveBrush: Since consumers don't have access to the underlying input element, they won't be able to add additional functionality to it (such as a `blur` or `click` event handler), so the SkyInputHarness is a bit overkill for what the harness is trying to accomplish. We really only need the essentials: set a value, get the hint text, get the ARIA label, and click the Done button.

libs/components/forms/testing/src/modules/file-attachment/file-drop/file-drop-harness.ts
Blackbaud-SteveBrush: I misread the contents of this method. Since you have a `getLinkUploadHarness` method that returns the link upload harness, could we just rely on that for retrieving the ARIA label? I think it's a better user experience if we wrap everything to do with "link upload" in its harness, and let the parent `SkyFileDropHarness` deal with the file drop target and the form control's label, help text, etc.
```
const label = await (await fileDropHarness.getLinkUploadHarness()).getAriaLabel();
```
Alternatively (I'm torn between which is better), you could move every public method from the `SkyFileDropLinkUploadHarness` into this harness, and prefix the methods with `linkUpload`. I.e., `enterLinkUploadText`, `isLinkUploadDisabled`, `clickLinkUploadDoneButton`, etc. 

I could be convinced either way which setup is better (having one less harness is nice, but it also adds more methods to the parent harness). At any rate, I think you're correct: we need to either 1) bring every method from the `SkyFileDropLinkUploadHarness` into the`SkyFileDropHarness` and prefix them with `linkUpload`, or 2) move any public methods that deal with "link upload" to the `SkyFileDropLinkUploadHarness`.

Blackbaud-SandhyaRajasabeson: i like the idea of less harnesses bc from the consumer's perspective file drop is one component. It does make the parent harness bigger, but it does make it more intuitive from the consumers perspective. 

for things like dropdown, the items being its own harness makes sense since they would be interacting with the items together while testing. but for file drop i can see scenarios where consumers swap between interacting with link upload and file drop back and forth

I'm gong to go with the one big harness way.

Blackbaud-SteveBrush: Cool! I was leaning toward that decision as well.

Blackbaud-SteveBrush (approved):

blackbaud-sky-build-user:
Cherry-pick [failed](https://github.com/blackbaud/skyux/actions/runs/12813726493)! :x: Please resolve conflicts and create a new PR.

PR #3495, opened 2025-06-04
feat(components/action-bars): add summary action bar code example with error

Blackbaud-SandhyaRajasabeson:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/3495/storybook/)

Component Storybooks:

- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/3495/storybooks/action-bars-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/3495/storybooks/pages-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/3495/storybooks/split-view-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/3495/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/3495/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/3495/playground/)

libs/components/code-examples/src/lib/modules/action-bars/summary-action-bar/error/example.component.ts
```diff
@@ -0,0 +1,81 @@
...
+      {
+        value: 1000,
+        label: 'Given this year',
+      },
+      {
+        value: 1300,
+        label: 'Given all time',
+      },
+    ]);
+    this.exampleForm = inject(FormBuilder).group({
```
Blackbaud-SteveBrush: Both `donationSummary` and `exampleForm` can be defined outside of the constructor.

libs/components/code-examples/src/lib/modules/action-bars/summary-action-bar/error/example.component.ts
```diff
@@ -0,0 +1,81 @@
...
+    alert('The primary action button was clicked.');
+  }
+
+  public singleError(): void {
+    this.donationSummary.setErrors({
+      singleError: 'There is an error.',
+    });
+  }
+
+  public multipleErrors(): void {
```
Blackbaud-SteveBrush: Can these be `protected`? ☝️

libs/components/code-examples/src/lib/modules/action-bars/summary-action-bar/error/example.component.ts
```diff
@@ -0,0 +1,81 @@
...
+      {
+        value: 1300,
+        label: 'Given all time',
+      },
+    ]);
+    this.exampleForm = inject(FormBuilder).group({
+      donationSummary: this.donationSummary,
+    });
+
+    this.donationSummary.statusChanges.subscribe(() => {
```
Blackbaud-SteveBrush: Needs a `takeUntilDestroyed()`? Is there potential for this subscription to be a signal so that you can call `formErrors()` in the template?

Blackbaud-SandhyaRajasabeson: I sadly do not this is possible bc converting `statusChanges` or `events` to a signal only updates when the value changes or when the status goes from `VALID` to `INVALID` so if there is one error, and then two more errors are added the signal doesnt update 😞 I have no idea why there isnt an observable to hook into for errors changing

libs/components/code-examples/src/lib/modules/action-bars/summary-action-bar/error/example.component.ts
Blackbaud-SandhyaRajasabeson: i need `onPrimaryActionClick` to stay public for the unit test, but the other two are now protected 🫡 good call!

libs/components/code-examples/src/lib/modules/action-bars/summary-action-bar/error/example.component.ts
```diff
@@ -0,0 +1,108 @@
...
+ * @title Summary action bar with errors
+ * @docsDemoHidden
+ */
+@Component({
+  selector: 'app-action-bars-summary-action-bar-error-example',
+  templateUrl: './example.component.html',
+  imports: [SkyKeyInfoModule, SkySummaryActionBarModule, ReactiveFormsModule],
+})
+export class ActionBarsSummaryActionBarErrorExampleComponent {
+  public donationSummary = new FormControl<donationSummary[] | null>([
```
Blackbaud-SteveBrush: Can these `public` properties be `protected`?

libs/components/code-examples/src/lib/modules/action-bars/summary-action-bar/error/example.component.ts
```diff
@@ -0,0 +1,108 @@
...
+    {
+      value: 1300,
+      label: 'Given all time',
+    },
+  ]);
+  public exampleForm = inject(FormBuilder).group({
+    donationSummary: this.donationSummary,
+  });
+
+  public valueChanged = toSignal(
```
Blackbaud-SteveBrush: Is this being used?

libs/components/code-examples/src/lib/modules/action-bars/summary-action-bar/error/example.component.ts
```diff
@@ -0,0 +1,108 @@
...
+    this.donationSummary.valueChanges.pipe(takeUntilDestroyed()),
+  );
+
+  public formErrors = computed(() => {
```
Blackbaud-SteveBrush: This is close, but could you use the above `valueChanged` signal with `switchMap` to combine the two statements?
```
protected formErrors = toSignal<SkySummaryActionBarError[] | undefined>(
  this.donationSummary.statusChanges.pipe(
    switchMap(() => {
      const errors = this.donationSummary.errors;

      return of(
        errors
          ? Object.values(errors).map((message) => ({ message }))
          : undefined,
      );
    }),
  ),
);
```

libs/components/code-examples/src/lib/modules/action-bars/summary-action-bar/error/example.component.ts
```diff
@@ -0,0 +1,108 @@
...
+      {
+        value: 1300,
+        label: 'Given all time',
+      },
+    ]);
+
+    this.donationSummary.setErrors([]);
+  }
+
+  protected singleError(): void {
```
Blackbaud-SteveBrush, line 65: It's preferred to have an actual form in the template (maybe an input box with a required field?) instead of a dropdown to show errors. I didn't know how to trigger the error state at first. Inserting a form will also be closer to what a consumer would implement in their SPA and be more helpful. (Maybe ask Claude to create the form for you? 😎)

Blackbaud-SandhyaRajasabeson: So the secondary buttons are only in dropdown form when the screen is smaller. While I would usually support having a more real life example, I think adding a proper form would bury the `summary-action-bar` part of the template more than it already is, and would make this code example differ more than the other code examples too much. I'd rather keep it a little unusable to avoid full copy pasting + keep the code similar/not too cluttered.

libs/components/code-examples/src/lib/modules/action-bars/summary-action-bar/error/example.component.ts
Blackbaud-SandhyaRajasabeson: yeah that's what's having the computed value to update 😭

libs/components/code-examples/src/lib/modules/action-bars/summary-action-bar/error/example.component.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        return of(
+          errors
+            ? Object.values(errors).map((error) => ({
+                message: error as string,
+              }))
+            : undefined,
+        );
+      }),
+      takeUntilDestroyed(),
```
Blackbaud-SteveBrush: `toSignal` automatically unsubscribes during the "destroy" step of the component, so this can be removed.

libs/components/code-examples/src/lib/modules/action-bars/summary-action-bar/error/example.component.ts
```diff
@@ -0,0 +1,78 @@
...
+    ),
+  );
+
+  public onPrimaryActionClick(): void {
```
Blackbaud-SteveBrush: You mentioned requiring this method to be `public` because of the unit tests. Do we even need to test this for the "errors" code example? We already demonstrate primary/secondary actions in the "basic" code example; might be redundant. I'd probably just remove this public method and the test to keep the focus on the error states.

Blackbaud-SandhyaRajasabeson: im going to remove testing this button, but leave the experience for users so that a primary button is still there and interactable

Blackbaud-SteveBrush (approved):

PR #3547, opened 2025-06-26
chore(components/filter-bar): add filter item component and begin filter modal workflow

Blackbaud-CoreyArcher:
[AB#3345400](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3345400)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/3547/storybook/)

Component Storybooks:

- [filter-bar-storybook](https://blackbaud.github.io/skyux-pr-preview/3547/storybooks/filter-bar-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/3547/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/3547/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/3547/playground/)

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar-item.component.scss
```diff
@@ -0,0 +1,65 @@
...
+}
+
+@include compatMixins.sky-modern-overrides('.sky-filter-bar-btn-selected') {
+  --sky-override-filter-bar-item-border-selected: inset 0 0 0
+    var(--sky-border-width-action-base) var(--sky-color-border-selected);
+}
+
+.sky-filter-bar-btn {
+  display: flex;
+  max-width: 300px;
```
Blackbaud-CoreyArcher, line 15: Placeholder?

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-value.ts
```diff
@@ -0,0 +1,4 @@
+export interface SkyFilterBarFilterValue {
+  value: unknown;
+  displayValue?: string;
```
johnhwhite, line 3: Should `displayValue` be required?

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-item.ts
```diff
@@ -0,0 +1,9 @@
+import { SkyFilterBarFilterModalConfig } from './filter-bar-filter-modal-config';
+import { SkyFilterBarFilterValue } from './filter-bar-filter-value';
+
+export interface SkyFilterBarFilterItem {
+  id: string | number;
```
johnhwhite: It seems like it'd be simpler if `id` were only a `string`. Is there a reason to allow `number`?

Blackbaud-CoreyArcher: removed number type

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-value.ts
Blackbaud-CoreyArcher: `displayValue` being optional is for the case where the filter value itself is a string value that can be displayed in the app. It is intended to be used in most cases, so we can change it to required if we think that's a better workflow for the consumer

johnhwhite (approved):

PR #3628, opened 2025-07-10
chore: integrate filter bar with data manager service

Blackbaud-CoreyArcher:
[AB#3364676](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3364676)

### DIFF
--- a/libs/components/data-manager/src/lib/modules/data-manager/data-manager-toolbar/data-manager-toolbar.component.html
+++ b/libs/components/data-manager/src/lib/modules/data-manager/data-manager-toolbar/data-manager-toolbar.component.html
@@ -97,6 +97,9 @@
     </sky-toolbar-section>
     <ng-content select="sky-data-manager-toolbar-section" />
   </sky-toolbar>
+  @if (activeView?.filterBarEnabled) {
+    <ng-content select="sky-filter-bar" />
+  }
 
   @if (activeView?.multiselectToolbarEnabled) {
     <sky-toolbar class="sky-data-manager-multiselect-toolbar">
--- a/libs/components/data-manager/src/lib/modules/data-manager/data-manager-toolbar/data-manager-toolbar.component.ts
+++ b/libs/components/data-manager/src/lib/modules/data-manager/data-manager-toolbar/data-manager-toolbar.component.ts
@@ -4,9 +4,11 @@ import {
   Component,
   OnDestroy,
   OnInit,
+  effect,
   inject,
   isStandalone,
 } from '@angular/core';
+import { toSignal } from '@angular/core/rxjs-interop';
 import { FormsModule } from '@angular/forms';
 import { SkyLogService } from '@skyux/core';
 import {
@@ -16,7 +18,11 @@ import {
 } from '@skyux/forms';
 import { SkyIconModule } from '@skyux/icon';
 import { SkyToolbarModule } from '@skyux/layout';
-import { SkyFilterModule, SkySortModule } from '@skyux/lists';
+import {
+  SkyFilterModule,
+  SkyFilterStateService,
+  SkySortModule,
+} from '@skyux/lists';
 import { SkySearchModule } from '@skyux/lookup';
 import {
   SkyModalCloseArgs,
@@ -40,6 +46,8 @@ import { SkyDataManagerSortOption } from '../models/data-manager-sort-option';
 import { SkyDataManagerState } from '../models/data-manager-state';
 import { SkyDataViewConfig } from '../models/data-view-config';
 
+import { SkyDataManagerFilterStateService } from './data-manager-filter-state.service';
+
 /**
  * Renders a `sky-toolbar` with the contents specified by the active view's `SkyDataViewConfig`
  * and the `SkyDataManagerToolbarLeftItemComponent`, `SkyDataManagerToolbarRightItemComponent`,
@@ -60,6 +68,12 @@ import { SkyDataViewConfig } from '../models/data-view-config';
     SkySearchModule,
     SkySortModule,
     SkyToolbarModule,
+  ],
+  providers: [
+    {
+      provide: SkyFilterStateService,
+      useClass: SkyDataManagerFilterStateService,
+    },
   ],
 })
 export class SkyDataManagerToolbarComponent implements OnDestroy, OnInit {
@@ -117,7 +131,29 @@ export class SkyDataManagerToolbarComponent implements OnDestroy, OnInit {
   readonly #changeDetector = inject(ChangeDetectorRef);
   readonly #columnPickerService = inject(SkyDataManagerColumnPickerService);
   readonly #dataManagerService = inject(SkyDataManagerService);
+  readonly #filterStateService = inject(SkyFilterStateService);
   readonly #modalService = inject(SkyModalService);
+
+  readonly #filterStateUpdated = toSignal(
+    this.#filterStateService.filterBarChanges,
+  );
+
+  constructor() {
+    effect(() => {
+      const filterState = this.#filterStateUpdated();
+
+      if (
+        filterState &&
+        this.dataState &&
+        JSON.stringify(this.dataState.extensions?.filterState) !==
+          JSON.stringify(filterState)
+      ) {
+        this.dataState.extensions ??= {};
+        this.dataState.extensions.filterState = filterState;
+        this.#dataManagerService.updateDataState(this.dataState, this.#_source);
+      }
+    });
+  }
 
   public ngOnInit(): void {
     this.#dataManagerService
@@ -148,6 +184,15 @@ export class SkyDataManagerToolbarComponent implements OnDestroy, OnInit {
       .getDataStateUpdates(this.#_source)
       .pipe(takeUntil(this.#ngUnsubscribe))
       .subscribe((dataState) => {
+        if (
+          dataState?.extensions?.filterState &&
+          JSON.stringify(dataState.extensions.filterState) !==
+            JSON.stringify(this.dataState?.extensions?.filterState)
+        ) {
+          this.#filterStateService.updateFilterBar(
+            dataState.extensions.filterState,
+          );
+        }
         this.#_dataState = dataState;
         this.onlyShowSelected = dataState.onlyShowSelected;
         this.#changeDetector.markForCheck();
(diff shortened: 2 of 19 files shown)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/3628/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/3628/storybooks/ag-grid-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/3628/storybooks/data-manager-storybook/)
- [filter-bar-storybook](https://blackbaud.github.io/skyux-pr-preview/3628/storybooks/filter-bar-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/3628/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/3628/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/3628/storybooks/lookup-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/3628/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/3628/storybooks/phone-field-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/3628/storybooks/split-view-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/3628/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/3628/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/3628/playground/)

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar-summary-item.component.spec.ts
Blackbaud-CoreyArcher, line 1: These changes came from the last cherry pick to v13.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.spec.ts
```diff
@@ -72,18 +72,20 @@ describe('Filter bar component', () => {
...
-            filterValue: { value: 'dm-value1' },
-          },
-          {
-            id: 'dm-filter2',
-            name: 'Data Manager Filter 2',
-            filterValue: undefined,
-          },
-        ],
+        filters: {
+          filterItems: [
```
Blackbaud-ErikaMcVey: So I was expecting `filterItems` would be a sibling to `filters`, not nested with it. If we add a new property `filterBarItems` to `SkyDataManagerFilterData`, then we can have reliably typed data, and `filters` can continue to be used by consumers for additional filters and we won't step on it.

Blackbaud-CoreyArcher: `filterBarItems` property added

libs/components/lists/src/lib/modules/filter/filter-state.service.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Used to send and receive filter states between components.
+ * @internal
+ */
+export abstract class SkyFilterStateService {
+  public abstract readonly dataStateChanges: Observable<SkyFilterState>;
+  public abstract readonly filterBarChanges: Observable<SkyFilterState>;
+
+  public abstract updateDataState(value: SkyFilterState): void;
+  public abstract updateFilterBar(value: SkyFilterState): void;
```
Blackbaud-SteveBrush, line 14: This is an internal service, but since it's purpose is abstract we should probably avoid using terms found in data-manager and filter-bar so that anything can use it.

I'm assuming you tried to implement `distinctUntilChanged` with only a single "update" method? Copilot recommended trying the "source identification" pattern for this service. It gave me the following class as an example. I asked Copilot to implement this service pattern with data-manager and filter-bar and it didn't look like too many changes. (This is actually similar to what data-manager does with its own state service.) Mind giving it a shot?
```
/**
 * Used to send and receive filter states between components.
 * Implements source identification to prevent feedback loops when components
 * both listen for and emit state changes.
 * @internal
 */
export abstract class SkyFilterStateService {
  /**
   * Set a new filter state with source identification.
   * @param value The filter state to emit
   * @param sourceId Unique identifier for the component emitting this change
   */
  public abstract setState(value: SkyFilterState, sourceId: string): void;

  /**
   * Get state changes excluding those from a specific source to prevent feedback loops.
   * @param excludeSourceId The source ID to exclude from the observable
   * @returns Observable of filter states excluding changes from the specified source
   */
  public abstract getStateChanges(
    excludeSourceId: string,
  ): Observable<SkyFilterState>;
}
```
With the above abstract class, Copilot created the following implementation:
```
@Injectable()
export class SkyDataManagerFilterStateService implements SkyFilterStateService {
  readonly #stateSubject = new BehaviorSubject<SkyFilterStateChange | null>(
    null,
  );

  readonly #stateChanges = this.#stateSubject.pipe(
    filter((change): change is SkyFilterStateChange => change !== null),
  );

  public setState(value: SkyFilterState, sourceId: string): void {
    this.#stateSubject.next({ data: value, sourceId });
  }

  public getStateChanges(excludeSourceId: string): Observable<SkyFilterState> {
    return this.#stateChanges.pipe(
      filter((change) => change.sourceId !== excludeSourceId),
      map((change) => change.data),
    );
  }
}
```

libs/components/lists/src/lib/modules/filter/filter-state.ts
```diff
@@ -0,0 +1,15 @@
+import { SkyFilterStateItem } from './filter-state-item';
+
+/**
+ * Represents the current filter state.
+ */
+export interface SkyFilterState {
+  /**
+   * The filters that have been set.
+   */
+  filters?: SkyFilterStateItem[];
```
Blackbaud-SteveBrush, line 10: Can we name this property (and it's companion input) something more specific? A consumer might think this configures filters in general, but it's really a property that represents active filter values. `appliedFilters`, or something similar.

libs/components/data-manager/src/lib/modules/data-manager/models/data-manager-state-options.ts
```diff
@@ -13,6 +14,10 @@ export interface SkyDataManagerStateOptions {
    */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   additionalData?: any;
+  /**
+   * Stores settings for data manager extensions.
```
Blackbaud-SteveBrush, line 18: Let's throw `@internal` on this property to communicate it's not for consumer use. Maybe prefix with `_` to further communicate the intention.

libs/components/data-manager/src/lib/modules/data-manager/models/data-manager-state-extensions.ts
```diff
@@ -0,0 +1,10 @@
+import { SkyFilterState } from '@skyux/lists';
+
+export type SkyDataManagerExtensionsFilterState = SkyFilterState;
+
+export interface SkyDataManagerStateExtensions {
+  /**
+   * Stores data for filter bar filters.
```
Blackbaud-SteveBrush, line 7: This property is used for any filter "thing", so I wouldn't mention filter bar in the description.

libs/components/data-manager/src/lib/modules/data-manager/models/data-manager-state-extensions.ts
```diff
@@ -0,0 +1,10 @@
+import { SkyFilterState } from '@skyux/lists';
+
+export type SkyDataManagerExtensionsFilterState = SkyFilterState;
+
+export interface SkyDataManagerStateExtensions {
```
Blackbaud-SteveBrush, line 5: Let's slap an `@internal` on this type just to be safe.

libs/components/lists/src/lib/modules/filter/filter-state.service.ts
Blackbaud-SteveBrush: The above suggestion might allow you to get rid of the expensive `JSON.stringify(` calls in the DataManagerToolbar, too.

libs/components/data-manager/src/lib/modules/data-manager/data-manager-toolbar/data-manager-toolbar.component.html
```diff
@@ -97,6 +97,9 @@
     </sky-toolbar-section>
     <ng-content select="sky-data-manager-toolbar-section" />
   </sky-toolbar>
+  @if (activeView?.filterBarEnabled) {
```
Blackbaud-SteveBrush, line 100: ~~What do you think about making the default value for `filterBarEnabled` equal true? That way, simply putting the filter-bar in the toolbar will make it appear. This would avoid the confusion of a consumer adding the filter-bar and not understanding why it doesn't show up in their grid right away.~~
```
@if (activeView?.filterBarEnabled !== false) {
```

libs/components/data-manager/src/lib/modules/data-manager/data-manager-toolbar/data-manager-toolbar.component.ts
```diff
@@ -117,8 +131,30 @@ export class SkyDataManagerToolbarComponent implements OnDestroy, OnInit {
...
+  );
+
+  constructor() {
+    effect(() => {
+      const filterState = this.#filterStateUpdated();
+
+      if (
+        filterState &&
+        this.dataState &&
+        JSON.stringify(this.dataState.extensions?.filterState) !==
```
Blackbaud-SteveBrush, line 148: JSON.stringify is expensive, and has some drawbacks when comparing two objects. For example, two objects might have the same properties, but if the properties are in a different order, JSON.stringify will say they're different. Also, JSON.stringify doesn't handle undefined/null properties very well. If you absolutely need this check, I'd ask Copilot to write a more efficient object comparison utility that can be used with `distinctUntilChanged`.

libs/components/data-manager/src/lib/modules/data-manager/data-manager-toolbar/data-manager-toolbar.component.html
Blackbaud-SteveBrush: Actually, on second thought, the filter bar can be hidden/shown by the consumer, manually. If it needs to be shown/hidden based on the view, the host component can listen for data state changes, see what the active view is, and act accordingly.
```
@if (activeView() === 'grid') {
  <sky-filter-bar />
}
```

PR #3641, opened 2025-07-14
fix: disabled elements use `not-allowed` cursor

Blackbaud-TrevorBurch:
[AB#3420454](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3420454)

### DIFF
--- a/libs/components/indicators/src/lib/modules/tokens/token.component.html
+++ b/libs/components/indicators/src/lib/modules/tokens/token.component.html
@@ -1,8 +1,8 @@
 <div
   class="sky-token sky-btn sky-btn-default"
+  [attr.disabled]="disabled ? 'true' : null"
   [attr.role]="role"
   [ngClass]="{
-    'sky-btn-disabled': disabled,
     'sky-token-disabled': disabled,
     'sky-token-dismissible': dismissible,
     'sky-token-active': tokenActive && !closeActive,
@@ -24,9 +24,6 @@
       skyId
       [attr.tabindex]="tabIndex"
       [disabled]="disabled"
-      [ngClass]="{
-        'sky-btn-disabled': disabled
-      }"
     >
       <ng-content />
     </button>
@@ -51,7 +48,6 @@
         "
         [disabled]="disabled"
         [ngClass]="{
-          'sky-btn-disabled': disabled,
           'sky-token-btn-close-active': closeActive
         }"
         (click)="dismissToken($event)"
--- a/libs/components/select-field/src/lib/modules/select-field/select-field.component.html
+++ b/libs/components/select-field/src/lib/modules/select-field/select-field.component.html
@@ -45,7 +45,7 @@
         singleSelectOpenButtonTitle ||
         ('skyux_select_field_single_select_open_button_title' | skyLibResources)
       "
-      [ngClass]="{ 'sky-btn-disabled': disabled }"
+      [disabled]="disabled"
       (click)="openPicker()"
       (keydown.enter)="
         openPicker(); $event.preventDefault(); $event.stopPropagation()
--- a/libs/components/select-field/src/lib/modules/select-field/select-field.component.html
+++ b/libs/components/select-field/src/lib/modules/select-field/select-field.component.html
@@ -41,11 +41,11 @@
       [attr.tabindex]="disabled ? false : 0"
       [attr.aria-label]="ariaLabel"
       [attr.aria-labelledby]="ariaLabelledBy"
+      [attr.disabled]="disabled"
       [attr.title]="
         singleSelectOpenButtonTitle ||
         ('skyux_select_field_single_select_open_button_title' | skyLibResources)
       "
-      [disabled]="disabled"
       (click)="openPicker()"
       (keydown.enter)="
         openPicker(); $event.preventDefault(); $event.stopPropagation()
--- a/libs/components/theme/src/lib/styles/_buttons.scss
+++ b/libs/components/theme/src/lib/styles/_buttons.scss
@@ -32,9 +32,12 @@ button {
   &.sky-btn-disabled,
   &[disabled] {
     cursor: not-allowed;
-    pointer-events: none;
     opacity: 0.65;
     box-shadow: none;
+  }
+
+  &.sky-btn-disabled {
+    pointer-events: none;
   }
 }
 
--- a/apps/e2e/theme-storybook/src/app/buttons/buttons.component.html
+++ b/apps/e2e/theme-storybook/src/app/buttons/buttons.component.html
@@ -187,14 +187,16 @@
   </a>
 
   <button
-    class="sky-btn sky-btn-default sky-btn-icon sky-btn-disabled sky-margin-inline-compact"
+    class="sky-btn sky-btn-default sky-btn-icon sky-margin-inline-compact"
+    disabled
     type="button"
   >
     <i class="fa fa-angle-double-down"></i>
   </button>
 
   <button
-    class="sky-btn sky-btn-icon-borderless sky-btn-disabled sky-margin-inline-compact"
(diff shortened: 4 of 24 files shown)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/3641/storybook/)

Component Storybooks:

- [a11y-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/a11y-storybook/)
- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/colorpicker-storybook/)
- [core-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/core-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/errors-storybook/)
- [filter-bar-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/filter-bar-storybook/)
- [flyout-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/flyout-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/forms-storybook/)
- [help-inline-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/help-inline-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/lookup-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/modals-storybook/)
- [navbar-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/navbar-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/phone-field-storybook/)
- [popovers-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/popovers-storybook/)
- [progress-indicator-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/progress-indicator-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/split-view-storybook/)
- [storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/theme-storybook/)
- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/tiles-storybook/)
- [toast-storybook](https://blackbaud.github.io/skyux-pr-preview/3641/storybooks/toast-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/3641/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/3641/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/3641/playground/)

libs/components/theme/src/lib/styles/_buttons.scss
```diff
@@ -32,10 +32,13 @@ button {
   &.sky-btn-disabled,
   &[disabled] {
     cursor: not-allowed;
-    pointer-events: none;
     opacity: 0.65;
     box-shadow: none;
   }
+
+  &.sky-btn-disabled {
+    pointer-events: none;
```
Blackbaud-TrevorBurch, line 40: The underlying issue here was the `pointer-events` was blocking the cursor rule. The main disabled attribute that people normally should be using blocks click and mouse events; however, if someone is only relying on the class then we need to maintain this to avoid a breaking change. So I moved the rule into its own selector to where we maintain the pointer events being off for the class but fix the cursor for things that are disabled the normal way.

Blackbaud-SteveBrush: Can you help me understand why this would be a breaking change for consumers? Wouldn't we want the class and the `disabled` attribute to produce the same cursor?

libs/components/select-field/src/lib/modules/select-field/select-field.component.html
```diff
@@ -41,11 +41,11 @@
       [attr.tabindex]="disabled ? false : 0"
       [attr.aria-label]="ariaLabel"
       [attr.aria-labelledby]="ariaLabelledBy"
+      [attr.disabled]="disabled"
       [attr.title]="
         singleSelectOpenButtonTitle ||
         ('skyux_select_field_single_select_open_button_title' | skyLibResources)
       "
-      [ngClass]="{ 'sky-btn-disabled': disabled }"
```
Blackbaud-SteveBrush, line 48: I like this change, but do you forsee this being a problem for consumers that have unit tests expecting these class names to exist?

Blackbaud-TrevorBurch: I would argue that if it is - they are utilizing things that we have explicitly said are not a part of the public API.

I modified any test harnesses that looked for the classes to still work after this change. 

And with the leaving of the pointer events off on the class - I can not fix this bug without a ton more css and leave the class.

libs/components/theme/src/lib/styles/_buttons.scss
Blackbaud-TrevorBurch: Because the `pointer-events: none;` turns off any mouse and touch events from firing on the element.

If a consumer has used `sky-btn-disabled` to style a button as disabled but did not put the `[disabled]` attribute as well - then the class and `pointer-events: none` is the only thing that is causing the button to not be clickable. However, `pointer-events: none` also has a side-effect of causing `cursor` CSS rules to not apply. 

This solution allows us to fix the issue for the vast majority of disabled things - while not breaking consumers that might have omitted the disabled attribute on their element.

libs/components/indicators/src/lib/modules/tokens/token.component.html
```diff
@@ -1,8 +1,8 @@
 <div
   class="sky-token sky-btn sky-btn-default"
+  [attr.disabled]="disabled ? 'true' : null"
```
Blackbaud-SteveBrush, line 3: Is this valid HTML? From the spec it doesn't look like a DIV has a `disabled` attribute. https://www.w3.org/TR/html4/index/attributes.html

Could you use `aria-disabled=true` instead?

PR #3659, opened 2025-07-15
fix(components/packages): updated the `ng add` schematic to configure a workspace with a single project; workspaces with multiple projects should use the `add-skyux-to-project` schematic

Blackbaud-SteveBrush:
## Before this change

Consumers ran `ng add @skyux/packages --project my-app`, but requiring a project name for the `add` schematic isn't an ideal pattern since an `add` schematic is primarily used to install dependencies and add configurations for the workspace. Most consumers would be accustomed to running `ng add` without arguments, so the changes described below are necessary. 

## Summary of changes

Updated the `ng add` schematic to only configure a workspace if it includes a single project. Workspaces with multiple projects should use the new `add-skyux-to-project` schematic. This paradigm is used by `angular-eslint` schematics as well.

Run for workspaces with a single project (likely an application):
```
ng add @skyux/packages
```

Run for each project in a multi-project workspace (likely a library and a "showcase" application):
```
ng g @skyux/packages:add-skyux-to-project --project my-lib
ng g @skyux/packages:add-skyux-to-project --project my-lib-showcase
```


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/3659/storybook/)

Component Storybooks:

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/3659/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/3659/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/3659/playground/)

libs/components/packages/src/schematics/migrations/update-13/i18n-resources-module/i18n-resources-module.schematic.ts
```diff
@@ -1,14 +1,14 @@
 import { Rule, chain, externalSchematic } from '@angular-devkit/schematics';
+import { getWorkspace } from '@schematics/angular/utility/workspace';
```
Blackbaud-SteveBrush, line 2: We can use Angular's utilities, instead of creating our own.

libs/components/packages/src/schematics/ng-add/ng-add.schematic.ts
```diff
@@ -1,38 +1,44 @@
...
+    const rules = [
       workspaceCheck(),
-      installEssentialSkyUxPackages(),
-      installAngularCdk(),
-      addCommonJsConfig(projectName),
-      addPolyfillsConfig(projectName, ['build', 'test']),
-      applySkyuxStylesheetsToWorkspace(projectName),
-    ]);
+      addDependencies(),
+      configureWorkspaceIfSingleProject(),
```
Blackbaud-SteveBrush, line 20: ☝️ Primary change.

libs/components/packages/src/schematics/rules/add-commonjs-config.spec.ts
Blackbaud-SteveBrush, line 1: All of these deleted schematics were simplified into a single function in the `add-skyux-to-project` schematic. (We got a little too "rule" happy with these.)

Blackbaud-TrevorBurch:
@Blackbaud-SteveBrush - I'm not seeing any message for consumers if they try to run `ng add @skyux/packages` with two packages. Should we output something so that consumers know that we aborted and what they could do to get where they are trying to go?

Blackbaud-TrevorBurch (approved):

PR #3679, opened 2025-07-17
feat(components/filter-bar): add filter bar library and component

Blackbaud-CoreyArcher:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/3679/storybook/)

Component Storybooks:

- [a11y-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/a11y-storybook/)
- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/colorpicker-storybook/)
- [core-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/core-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/errors-storybook/)
- [filter-bar-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/filter-bar-storybook/)
- [flyout-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/flyout-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/forms-storybook/)
- [help-inline-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/help-inline-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/lookup-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/modals-storybook/)
- [navbar-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/navbar-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/phone-field-storybook/)
- [popovers-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/popovers-storybook/)
- [progress-indicator-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/progress-indicator-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/split-view-storybook/)
- [storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/theme-storybook/)
- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/tiles-storybook/)
- [toast-storybook](https://blackbaud.github.io/skyux-pr-preview/3679/storybooks/toast-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/3679/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/3679/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/3679/playground/)

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-config.ts
```diff
@@ -1,9 +1,22 @@
...
+   * The component should expect a `SkyFilterBarFilterModalContext` object to be provided by the modal service on instantiation.
+   * The return value of the modal save action should be a `SkyFilterBarFilterValue`.
+   * @required
+   */
   modalComponent: Type<unknown>;
-
-  modalSize?: 'small' | 'medium' | 'large' | 'full';
-
+  /**
+   * The size of the modal to display.
```
blackbaud-johnly:
```suggestion
   * The size of the modal to display. The valid options are `small`, `medium`, `large`, and `full`.
```
Seems like we should call out the size options like we do in the corresponding `size` description in the modal docs. That description also has a third sentence that says "This property defaults to `medium`." but I assume the "@default" here makes that unnecessary.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-config.ts
```diff
@@ -1,9 +1,22 @@
...
+   * @default 'medium'
+   */
+  modalSize?: SkyFilterBarFilterModalSizeType;
+  /**
+   * Any additional properties to be passed in to the filter modal.
```
blackbaud-johnly:
```suggestion
   * Additional properties to pass to the filter modal. For example, on a single-filter modal with all filter types, use this property to pass in available options for each filter field.
```
I think we can tighten up and avoid passive voice, but please just ignore if we need passive voice in this case for clarity.

Also, can we provide more description of what additional properties would be passed? I think even just an example here would be helpful to explain the purpose of the property. (i.e. "Additional properties, such as x and y, to pass to the filter modal."

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-config.ts
```diff
@@ -1,9 +1,22 @@
 import { Type } from '@angular/core';
 
+import { SkyFilterBarFilterModalSizeType } from './filter-bar-filter-modal-size';
+
 export interface SkyFilterBarFilterModalConfig {
+  /**
+   * The modal component to display when the user selects the filter.
+   * The component should expect a `SkyFilterBarFilterModalContext` object to be provided by the modal service on instantiation.
```
blackbaud-johnly:
```suggestion
   * The component expects the modal service to provide a `SkyFilterBarFilterModalContext` object on instantiation.
```
The "should" phrasing makes it sound like this is optional. I don't think it is, so I reworded to avoid that implication. If I'm wrong and it is optional, we should ignore my suggestions but should also expand the description to explain what the other options are or change the wording to make this a recommendation/best practice.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-config.ts
```diff
@@ -1,9 +1,22 @@
...
+   * The return value of the modal save action should be a `SkyFilterBarFilterValue`.
```
blackbaud-johnly:
```suggestion
   * The return value of the modal save action is a `SkyFilterBarFilterValue`.
```
Same issue with "should" making it unclear whether this is a requirement or an option. My suggested edit just replaces the "should be" with "is," but if this is truly just optional or a recommendation, we should reword differently to make that clearer.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
```diff
@@ -1,10 +1,21 @@
 import { SkyFilterBarFilterValue } from './filter-bar-filter-value';
 
+/**
+ * The context object that is provided to a filter modal.
+ */
 export class SkyFilterBarFilterModalContext {
+  /**
+   * The name of the filter for displaying in the modal title.
```
blackbaud-johnly:
```suggestion
   * The name of the filter to display in the modal title.
```

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
```diff
@@ -1,10 +1,21 @@
...
+   */
   public filterName: string;
-
+  /**
+   * The value of the filter.
+   */
   public filterValue?: SkyFilterBarFilterValue;
-
+  /**
+   * An untyped property that can track any config information relevant to the filter modal
```
blackbaud-johnly:
```suggestion
   * An untyped property that can track additional config information for the filter modal
```

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
```diff
@@ -1,10 +1,21 @@
...
+   * that existing options do not include.
```
blackbaud-johnly:
```suggestion
   * that isn't provided through other properties.
```
I'm not completely sure if this is better, but I was initially confused about what we meant by "that existing options do not include" so I played around with the text a bit. Hopefully this edit and the edit on the previous line make things clearer, but please just disregard the suggestions if not.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-config.ts
blackbaud-johnly: After adding this suggestion, I noticed below that SkyFilterBarFilterModalSizeType also lists `full` as an option. But it's not listed as an option in the modal docs for the `size` property. Is that a bug in the modal docs? Or a discrepancy between the modal docs and filter bar docs?

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/example.component.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    // Get all filter items
+    const filterItems = await filterBarHarness.getItems();
+    expect(filterItems).toHaveSize(3);
+
+    // Check specific filter items exist
+    const staffAssignedFilter = await filterBarHarness.getItem({
+      filterId: 'staff-assigned',
+    });
+    await expectAsync(staffAssignedFilter.getName()).toBeResolvedTo(
+      'Staff assigned',
```
blackbaud-johnly, line 47: I don't have the context of seeing this in the demo, but it seems like "Staff-assigned" should probably be hyphenated.

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-selection-values.ts
```diff
@@ -0,0 +1,81 @@
...
+
+// eslint-disable @cspell/spellchecker
+export const FILTER_SELECTION_VALUES: Record<
+  string,
+  SkyFilterBarFilterValue[]
+> = {
+  'community-connection': [
+    { value: 'child-of-faculty', displayValue: 'Child of faculty' },
+    { value: 'child-of-alum', displayValue: 'Child of alum' },
+    { value: 'grandparent-is-alum', displayValue: 'Grandparent is an alum' },
```
blackbaud-johnly:
```suggestion
    { value: 'grandparent-is-alum', displayValue: 'Grandchild of alum' },
```
I'd tweak for parallel construction with "Child of alum"

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-selection-values.ts
```diff
@@ -0,0 +1,81 @@
...
+    { value: 'related-to-trustee', displayValue: 'Related to trustee' },
+    { value: 'sibling-candidate', displayValue: 'Sibling is a candidate' },
```
blackbaud-johnly:
```suggestion
    { value: 'sibling-candidate', displayValue: 'Sibling is candidate' },
```
We don't use articles in the other items so far, so I'd delete it here too and also in the items that follow. ... Alternatively, we could add the missing articles to the items before this.

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-selection-values.ts
```diff
@@ -0,0 +1,81 @@
...
+    {
+      value: 'sibling-past-candidate',
+      displayValue: 'Sibling is a past candidate',
```
blackbaud-johnly:
```suggestion
      displayValue: 'Sibling is past candidate',
```

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-selection-values.ts
```diff
@@ -0,0 +1,81 @@
...
+    },
+    { value: 'sibling-student', displayValue: 'Sibling is a student' },
```
blackbaud-johnly:
```suggestion
    { value: 'sibling-student', displayValue: 'Sibling is student' },
```

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-selection-values.ts
```diff
@@ -0,0 +1,81 @@
...
+    { value: 'sibling alum', displayValue: 'Sibling is an alum' },
```
blackbaud-johnly:
```suggestion
    { value: 'sibling alum', displayValue: 'Sibling is alum' },
```

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-selection-values.ts
```diff
@@ -0,0 +1,81 @@
...
+    {
+      value: 'sibling-incoming',
+      displayValue: 'Sibling is an incoming student',
```
blackbaud-johnly:
```suggestion
      displayValue: 'Sibling is incoming student',
```

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-config.ts
Blackbaud-CoreyArcher: Rather than add an additional `fullScreen` property, the options for size and fullscreen are combined into one expanded option here

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-config.ts
Blackbaud-CoreyArcher: The code example demonstrates this by using a single filter modal component across all the different filter types, and using this property to pass in the available options for each filter field

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
Blackbaud-CoreyArcher: I borrowed this phrasing from the Data Manager documentation, where the additionalOptions pattern originated. We can update it if you feel it's better, but I just used what was already there for consistency.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
Blackbaud-CoreyArcher: Same as above

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/example.component.spec.ts
Blackbaud-CoreyArcher: `Staff assigned` is what was specced in the design mock-up for the demo

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
Blackbaud-CoreyArcher: I used this wording because this is a convenience property for the consumer to use if they wish, because they own the modal and its heading text. It is recommended that the heading of the modal be the name of the filter field, but not required.

libs/components/filter-bar/testing/src/modules/filter-bar/filter-item-harness.ts
```diff
@@ -43,6 +43,13 @@ export class SkyFilterItemHarness extends SkyComponentHarness {
     return await (await this.#getButton()).click();
   }
 
+  /**
+   * Gets the filter item value.
+   */
+  public async getValue(): Promise<string | undefined> {
```
Blackbaud-SteveBrush: 1. Does this get the raw value or the display value?
2. I see a method for `getName` (should that be `getLabelText`)?
3. For the method `hasActiveValue`, what is an "active" value? Is this referring to the "applied" value? I'm wondering if only having the `getValue` method is enough since if the filter has a value (and is not `undefined`), then the filter is effectively "applied"?

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
Blackbaud-SteveBrush: I think John's wording sounds good here, and doesn't communicate (necessarily) that the field is required.

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/example.component.spec.ts
Blackbaud-SteveBrush: I would defer to John for wording recommendations (unless the designers _also_ got the wording from a docs writer).

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/example.component.ts
```diff
@@ -0,0 +1,67 @@
...
+ * @title Filter bar basic example
+ */
+@Component({
+  selector: 'app-filter-bar-basic-example',
+  imports: [SkyFilterBarModule],
+  templateUrl: './example.component.html',
+})
+export class FilterBarBasicExampleComponent {
+  public appliedFilters: SkyFilterBarFilterItem[] | undefined;
+  public selectedFilterIds: string[] | undefined;
```
Blackbaud-SteveBrush: ☝️ These should be `protected`, and made into signals, if possible.

libs/components/filter-bar/documentation.json
```diff
@@ -0,0 +1,30 @@
...
+    "filter-bar": {
+      "development": {
+        "docsIds": [
+          "SkyFilterBarModule",
+          "SkyFilterBarComponent",
+          "SkyFilterBarFilterItem",
+          "SkyFilterBarFilterValue",
+          "SkyFilterBarFilterModalConfig",
+          "SkyFilterBarFilterModalContext",
+          "SkyFilterBarFilterModalSizeType"
```
Blackbaud-SteveBrush: When I ran `npx nx serve code-examples-playground --skip-nx-cache` I didn't see `SkyFilterItemModalComponent` listed under the Development tab. It probably needs to be added to this list.

libs/components/filter-bar/testing/src/modules/filter-bar/filter-bar-harness.spec.ts
```diff
@@ -103,12 +103,14 @@ describe('Filter bar test harness', () => {
...
     const filterItem1 = await filterBarHarness.getItem({ filterId: 'filter1' });
     await expectAsync(filterItem1.hasActiveValue()).toBeResolvedTo(true);
+    await expectAsync(filterItem1.getValue()).toBeResolvedTo('value1');
     await expectAsync(filterItem1.click()).toBeResolved();
 
     const filterItem2 = await filterBarHarness.getItem({
       labelText: 'Test filter 2',
     });
     await expectAsync(filterItem2.hasActiveValue()).toBeResolvedTo(false);
+    await expectAsync(filterItem2.getValue()).toBeResolvedTo(undefined);
```
Blackbaud-SteveBrush: Unrelated question: can you double-check the `getItem` and `getItems` methods and make sure they match the same signature that was agreed upon in Sandhya's PR? https://github.com/blackbaud/skyux/pull/3854

Blackbaud-SteveBrush: Also, I see the `hasActiveFilters` method. Should that be renamed to `hasAppliedFilters` to match the input name?

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
```diff
@@ -1,10 +1,21 @@
 import { SkyFilterBarFilterValue } from './filter-bar-filter-value';
 
+/**
+ * The context object that is provided to a filter modal.
+ */
 export class SkyFilterBarFilterModalContext {
+  /**
+   * The name of the filter for displaying in the modal title.
+   */
   public filterName: string;
```
Blackbaud-SteveBrush: Can we name this `filterLabelText` to match the input's name?

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/example.component.spec.ts
blackbaud-johnly: Do you have a link to the design mockup so that I can see it in context? (This definitely isn't a huge deal, so we can also always punt on this for now. I can make a note to follow up later and add the hyphen if it seems necessary after the docs are live.)

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-config.ts
blackbaud-johnly: Got it. I updated the suggestion to include `full` in the list of sizes.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-config.ts
blackbaud-johnly: I rewrote the suggestion to include a second sentence describing the example. I can rework the update as necessary; just let me know.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
blackbaud-johnly: OK. I don't think that was clear from the existing wording, and I'm still not sure I completely follow. I _think_ you are saying that this property is *NOT* setting a modal title but instead is naming the filter and that we recommend the two values. If so, I think this property description should be something more like "The name of the filter. We recommend using the same value as the title from the modal."

However, I can't tell if you instead mean that this property *IS* setting a modal title that overwrites the title from the modal itself. If that's the case, it seems like this should be something more like "The name of the filter. This value overwrites the title from the modal, and we recommend using the same value in both places." Or something like that.

Sorry for the lack of certainty, but I'm not 100 percent sure what we're trying to communicate here. I don't think that my initial change really changes the meaning that we're conveying to users though, so I think we need a longer, more detailed description here to make it clearer what we mean. Let me know if you want to talk through it some more. Thanks!

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
blackbaud-johnly: If you're confident that the description in the data manager docs is completely clear to consumers, then I agree that we should stick with that wording for consistency. However, if my concerns about this wording not being clear are legit, then I think we should make changes here and then create a docs follow-up story for me to go back and make the same changes to data manager. I'm not sure if my suggestion in the next line fixes things, but I do think "that existing options do not include" is hard to process. And in this first line, I think "any" seems confusing because a consumer who is just skimming and misses (or doesn't understand) the bit at the end could misunderstood "any" config properties to mean "all" properties, which I think is inaccurate. ... Again, no worries if you think this I'm overthinking things and this description works just fine as is. If so, we should just resolve the two comments without making any changes. But if the concerns are legit, I can circle back to make corresponding changes to data manager for consistency.

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/example.component.spec.ts
blackbaud-johnly: I checked the mockup and confirmed that we don't need the hyphen here. Thanks!

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modal.component.ts
```diff
@@ -0,0 +1,54 @@
...
+  protected selectedValue = this.#context.filterValue;
+
+  protected formGroup: FormGroup = this.#formBuilder.group({
+    selectedOption: this.#formBuilder.control(this.selectedValue?.value),
+  });
+
+  protected save(): void {
+    const selectedValue = this.formGroup.get('selectedOption')?.value as string;
+
+    if (selectedValue) {
```
Blackbaud-SteveBrush: **I encountered an issue on the code example:**
1. Select a value for one of the filters and save the modal.
2. Click on the filter button again and try to unselect the value by choosing "-- Select an option --". Click "Save".
3. Note that the filter's value is not unset.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
```diff
@@ -1,10 +1,21 @@
...
+   */
   public filterValue?: SkyFilterBarFilterValue;
-
+  /**
+   * An untyped property that can track any config information relevant to the filter modal
+   * that existing options do not include.
+   */
   public additionalContext?: Record<string, unknown>;
 
   constructor(
```
Blackbaud-SteveBrush: It's preferred to have a single `args` object, instead of having multiple parameters for a class constructor. This will make the API easier to interact with, and allow for adding new properties in the future.

```
constructor(args: SkyFilterBarFilterModalContextArgs) {
  const { filterLabelText, filterValue, etc. } = args;
}
```
Now, when newing up the context, the properties are named for easy identification:
```
new SkyFilterBarFilterModalContext({
  filterLabelText: this.labelText(),
  filterValue: this.filterValue(),
  ...
})
```

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-size.ts
```diff
@@ -0,0 +1,5 @@
+export type SkyFilterBarFilterModalSizeType =
+  | 'small'
+  | 'medium'
+  | 'large'
+  | 'full';
```
Blackbaud-SteveBrush: Let's call this `'fullScreen'` to make it clear what's being "filled".

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/example.component.ts
```diff
@@ -0,0 +1,67 @@
...
+      'staff-assigned',
+      'entering-grade',
+      'current-grade',
+    ];
+  }
+
+  #getFilterContext(filterId: string): {
+    items: SkyFilterBarFilterValue[];
+  } {
+    return { items: FILTER_SELECTION_VALUES[filterId] };
```
Blackbaud-SteveBrush: ### Support asynchronous modal context/data

The `additionalContext` property allows sending static/synchronous data to a filter modal, but we need to support the ability to load _asynchronous_ data when the filter button is pressed, and _before_ the modal is opened. Otherwise, the consumer would need to load all the filter data on page load, and this could get fairly cumbersome (and affect load times) if there are several filters. If you look at our [modal code examples](https://developer.blackbaud.com/skyux/components/modal?docs-active-tab=examples), we show a sky-wait component before the modal is shown to the user. Since the filter bar is opening the modal for the consumer, it's not possible for the consumer to create a sky-wait before launching the modal, so we need to support this behavior as a first-class feature of the filter modal. To get this working, we can use Angular's built-in output event, similar to how we handle `searchAsync` [on the lookup component](https://github.com/blackbaud/skyux/blob/main/libs/components/code-examples/src/lib/modules/lookup/lookup/multi-select/example.component.html#L20). Since we're creating a new `sky-filter-item-modal` output, it would be best to make the other modal-related properties in `filterModalConfig` (required) component inputs, mainly for consistency and to remove the need for a complex config object.

**consumer-foo.html**

```html
<sky-filter-bar>
  <sky-filter-item-modal
    filterId="foo"
    labelText="Foo"
    [modalComponent]="FooModalComponent" <!-- required -->
    modalSize="small" <!-- required -->
    (modalOpen)="onModalOpen($event)" <!-- optional -->
  />
</sky-filter-bar>
```

**consumer-foo.ts**

```typescript
protected onModalOpen(args: SkyFilterBarFilterModalOpenArgs): void {
  // You don't have to do this exactly. The idea is that the consumer would
  // give the filter bar a callback that returns an observable. The filter bar
  // would show a sky-wait, resolve the observable, and then open the modal
  // with the resolved data.
  args.data = this.#svc.getStuffFromDatabase().pipe(
    map((result) => ({
      items: result.stuff
    })),
  );
}
```

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/example.component.ts
```diff
@@ -0,0 +1,67 @@
...
+    modalSize: 'medium',
+    additionalContext: this.#getFilterContext('entering-grade'),
+  };
+  protected roleConfig: SkyFilterBarFilterModalConfig = {
+    modalComponent: FilterModalComponent,
+    modalSize: 'medium',
+    additionalContext: this.#getFilterContext('role'),
+  };
+  protected staffAssignedConfig: SkyFilterBarFilterModalConfig = {
+    modalComponent: FilterModalComponent,
```
Blackbaud-SteveBrush: I realize having one component for all the filter modals is efficient, but we don't want to communicate to the consumer (via code example) that all filters should use only one modal and pass "additional context" to differentiate. For the sake of communicating that a filter button _can_ have a dedicated modal component, I would create a separate component for each filter (the consumer can then decide if using one component for multiple filters is a good idea for their use case); it's still appropriate to load the data through the async mechanism described in "Support asynchronous modal context/data", however.

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modal.component.ts
```diff
@@ -0,0 +1,54 @@
...
+    FormsModule,
+    ReactiveFormsModule,
+    SkyInputBoxModule,
+    SkyModalModule,
+    SkyRepeaterModule,
+  ],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class FilterModalComponent {
+  protected readonly modalInstance = inject(SkyModalInstance);
```
Blackbaud-SteveBrush: ### Create a modal instance specific for filter modals

Currently, the consumer is expected to inject the `SkyModalInstance` and `SkyFilterBarFilterModalContext` into their filter modal components. They also need to know to call `instance.save()` with an argument of a specific shape/type, and to call `instance.cancel()` (instead of `instance.close()` when closing their modal). To make this easier for the consumer to do the right thing, we should create a `SkyFilterBarFilterModalInstance` with the following methods and properties:
```
public readonly context: SkyFilterBarFilterModalContext;

public save(args: SkyFilterBarFilterModalSaveArgs): void;

public cancel(): void;
```
This way, the consumer only needs to import one injection token into their modal component to have everything they need for interacting with the filter bar.

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modal.component.ts
```diff
@@ -0,0 +1,54 @@
...
+  readonly #context = inject(SkyFilterBarFilterModalContext);
+  readonly #formBuilder: FormBuilder = inject(FormBuilder);
+  protected headingText = this.#context.filterName;
+  protected items: SkyFilterBarFilterValue[] = this.#context
+    .additionalContext?.['items'] as SkyFilterBarFilterValue[];
```
Blackbaud-SteveBrush: As a side note, the `data` property on the modal context (see feedback, above) can use generics to make the type easier to interact with in the component (and not require `as`):

```
interface SkyFilterBarFilterModalContext<TData = Record<string, unknown>> {
	data?: TData;
}

class SkyFilterBarFilterModalInstance<TData = Record<string, unknown>> {
  public readonly context: SkyFilterBarFilterModalContext<TData>;
}
```

**my-filter-modal.component.ts**
```
readonly #instance = inject<SkyFilterBarFilterModalInstance<MyDataType>>(
  SkyFilterBarFilterModalInstance
);

this.#instance.context.data?.myItems // no need to use `as MyDataType`
```

Blackbaud-SteveBrush (commented):
I took another look at the filter modal implementation since it wasn't part of the original HTML API PR. Paul and I talked it over, and I think we should tweak our approach in a couple areas: how we handle async data and giving consumers a proper modal instance so they "fall into the pit of success" when working with the filter bar. I also spotted some naming and pattern issues in the harnesses from an earlier PR.

libs/components/filter-bar/testing/src/modules/filter-bar/filter-item-harness.ts
Blackbaud-CoreyArcher: 1. Display value
2. changed
3. removed

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/example.component.ts
Blackbaud-CoreyArcher: Updated

libs/components/filter-bar/documentation.json
Blackbaud-CoreyArcher: Fixed

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/example.component.ts
Blackbaud-CoreyArcher: updated

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-size.ts
Blackbaud-CoreyArcher: changed

libs/components/filter-bar/testing/src/modules/filter-bar/filter-bar-harness.spec.ts
Blackbaud-CoreyArcher: Those changes were in 13 only, but I pulled them over.

Updated method name

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
Blackbaud-CoreyArcher: changed

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modal.component.ts
Blackbaud-CoreyArcher: I was not able to duplicate this problem locally after updating the filter modals to meet the new pattern

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
Blackbaud-CoreyArcher: updated

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/example.component.ts
Blackbaud-CoreyArcher: added

libs/components/filter-bar/src/lib/modules/filter-bar/filter-item-modal.component.ts
```diff
@@ -34,13 +39,39 @@ import { SkyFilterItem } from './models/filter-item';
...
 })
 export class SkyFilterItemModalComponent implements SkyFilterItem {
+  /**
+   * A unique identifier for the filter item.
+   * @required
+   */
   public readonly filterId = input.required<string>();
+
+  /**
+   * The label text for the filter item.
```
blackbaud-johnly:
```suggestion
   * The label to display for the filter item.
```
Very minor edit, so feel free to disregard if it would slow things down. But "label" and "text" back-to-back felt redundant and also repeats the property name. We could alternately just delete "text" and not replace it with "to display."

libs/components/filter-bar/src/lib/modules/filter-bar/filter-item-modal.component.ts
```diff
@@ -34,13 +39,39 @@ import { SkyFilterItem } from './models/filter-item';
...
+   * @required
+   */
   public readonly labelText = input.required<string>();
+
   /**
-   * The configuration options for showing a custom filter modal.
+   * The modal component to display when the user selects the filter.
+   * The component needs to inject a `SkyFilterBarFilterModalContext` object on instantiation to receive context from the modal service.
```
blackbaud-johnly: Do you think there's any need/value to add "modal" before "component" here to be explicitly clear about which component we're talking about?

libs/components/filter-bar/src/lib/modules/filter-bar/filter-item-modal.component.ts
```diff
@@ -34,13 +39,39 @@ import { SkyFilterItem } from './models/filter-item';
...
+   * The return value of the modal save action needs to be a `SkyFilterBarFilterValue`.
+   * @required
    */
-  public readonly filterModalConfig =
-    input.required<SkyFilterBarFilterModalConfig>();
+  public readonly modalComponent = input.required<Type<unknown>>();
+
+  /**
+   * The size of the modal to display. The valid options are `'small'`, `'medium'`, `'large'`, and `'fullScreen'`.
```
blackbaud-johnly: Are the single quote marks around the size options correct? We don't have them in the corresponding JSDocs for modal at https://github.com/blackbaud/skyux/blob/main/libs/components/modals/src/lib/modules/modal/modal.interface.ts#L17. If we need them here, do I need to add them there?

Also, is `fullScreen` correct? In the modal docs at https://developer.blackbaud.com/skyux/components/modal?docs-active-tab=development, I see a `fullPage` configuration property, but not a `fullScreen` one.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-item-modal.component.ts
```diff
@@ -34,13 +39,39 @@ import { SkyFilterItem } from './models/filter-item';
...
+   * @default 'medium'
+   */
+  public readonly modalSize = input<SkyFilterBarFilterModalSizeType>();
+
+  /**
+   * Fires when the user clicks on the filter item. Consumers passing additional context data to a filter modal
```
blackbaud-johnly:
```suggestion
   * Fires when the user clicks the filter item. To pass additional context data to a filter modal, consumers
```
In the first sentence, I'm just removing an extra word. In the second sentence, I rearranged to use an opening clause because I think it makes this easier to follow. Both are very minor tweaks, so please disregard either if you disagree.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-item-modal.component.ts
```diff
@@ -34,13 +39,39 @@ import { SkyFilterItem } from './models/filter-item';
...
+   * must subscribe to this and return the context via the observable property on the event args.
```
blackbaud-johnly:
```suggestion
   * must subscribe to this event and return the context using the observable property on the event args.
```

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
blackbaud-johnly: Do you think the description here and in data manager docs is completely clear to consumers? Or should we make any of my suggested edits here and then follow up with edits to the data manager docs separately?

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
blackbaud-johnly: Same question about whether we should leave this as is because it's completely clear here and in data manager docs or if we should make edits here and then follow up with edits to the data manager docs separately?

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-opened-args.ts
```diff
@@ -0,0 +1,11 @@
+import { Observable } from 'rxjs';
+
+/**
+ * Arguments passed when a filter modal is opened.
```
blackbaud-johnly:
```suggestion
 * Arguments passed to a filter modal when it opens.
```
Please disregard if I'm incorrectly changing the meaning here. But I assume the arguments are passed to the filter modal, and if that's the case, I think "passed to" is more clear and direct than "passed when," which leaves more room for misinterpretation. But let me know if I'm not following this correctly and got it wrong.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-opened-args.ts
```diff
@@ -0,0 +1,11 @@
...
+ */
+export interface SkyFilterBarFilterModalOpenedArgs {
+  /**
+   * An Observable representing data passed into the filter modal as additional context.
```
blackbaud-johnly:
```suggestion
   * An observable representing data that is passed to the filter modal as additional context.
```
The Angular docs don't treat "observable" as a proper noun, so I don't think we need to here.

libs/components/filter-bar/testing/src/modules/filter-bar/filter-item-harness.ts
```diff
@@ -44,16 +44,16 @@ export class SkyFilterItemHarness extends SkyComponentHarness {
...
    */
-  public async getName(): Promise<string> {
-    return await (await this.#getFilterName()).text();
+  public async getFilterValue(): Promise<string | undefined> {
+    return await (await this.#getFilterValue())?.text();
   }
 
   /**
-   * Checks if the filter item has an active value.
+   * Gets the filter item label text.
```
blackbaud-johnly:
```suggestion
   * Gets the filter item label.
```
I don't _think_ we need "text" here, but please ignore this suggestion if I'm overlooking some distinction.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-config.ts
blackbaud-johnly: I don't see this property in place anymore, so I think we can close this comment thread. Is that correct?

libs/components/filter-bar/src/lib/modules/filter-bar/filter-item-modal.component.ts
Blackbaud-CoreyArcher: I noticed that the use of the `@default` tag resulted in a `'medium'` being displayed, so I added the same quotes around the other values. But to be consistent with other places we've done that ( https://github.com/blackbaud/skyux/blob/104ac06f3d187c5d4cbb4c12af615a539c6aa63c/libs/components/avatar/src/lib/modules/avatar/avatar.component.ts#L84 ) they should be double quotes, so I will update that

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-config.ts
Blackbaud-CoreyArcher: Correct, this property has been functionally replaced by the `modalOpened` output

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
Blackbaud-CoreyArcher: My vote is leave the wording consistent for similar features across multiple components. Then, if we decide that the descriptions need work, update them all at once to the new wording

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modal.component.ts
Blackbaud-CoreyArcher: added filter item modal instance

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modal.component.ts
Blackbaud-CoreyArcher: added generic typing to the filter item component and supporting classes

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-opened-args.ts
Blackbaud-CoreyArcher: addressed in new file

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-opened-args.ts
Blackbaud-CoreyArcher: addressed in new file

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-item-modal-instance.ts
```diff
@@ -0,0 +1,49 @@
...
+   */
+  public readonly context = inject<SkyFilterItemModalContext<TData>>(
+    SkyFilterItemModalContext,
+  );
+
+  readonly #canceled = new Subject<SkyFilterBarFilterValue | undefined>();
+  readonly #saved = new Subject<SkyFilterBarFilterValue | undefined>();
+
+  /**
+   * Emits when the user cancels the modal.
```
blackbaud-johnly:
```suggestion
   * Fires when the user cancels the modal.
```
Probably doesn't matter too much since this is marked internal, but "emits" seems like the wrong word here. I think we usually use "fires" to indicate that an action (like cancelling the modal) triggers an event and that we reserve "emit" for describing when the event sends out data or something.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-item-modal-instance.ts
```diff
@@ -0,0 +1,49 @@
...
+   * @internal
+   */
+  public readonly canceled = this.#canceled.asObservable();
+  /**
+   * Emits when the user saves the modal.
```
blackbaud-johnly:
```suggestion
   * Fires when the user saves the modal.
```

libs/components/filter-bar/testing/src/modules/filter-bar/filter-bar-harness.ts
```diff
@@ -48,7 +48,8 @@ export class SkyFilterBarHarness extends SkyComponentHarness {
   }
 
   /**
-   * Gets a harness for a specific filter item that meets certain criteria.
+   * Gets a specific filter item based on the filter criteria.
+   * @param filter The filter criteria.
```
blackbaud-johnly, line 52: Can't remember if we already discussed this, so just checking: Is it correct that we need "optional" before "filter criteria" in the subsequent description for `getItems` but that we don't need it here for `getItem`?

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-modal-context.ts
blackbaud-johnly: Yeah, I'm all for consistency, but I don't think this description is clear and thought we should update it here and then follow up by updating the other property for consistency rather than reusing an unclear property description. But I think we're out of time to work on it here, so I'll just resolve the 2 comments related to this and make a note to hopefully circle back at some point. Thanks.

libs/components/filter-bar/testing/src/modules/filter-bar/filter-bar-harness.ts
Blackbaud-CoreyArcher: Correct. This comment came directly from the work Sandhya did in v13. It matches all of the other harnesses' methods.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-item-modal.component.ts
```diff
@@ -33,16 +43,47 @@ import { SkyFilterItem } from './models/filter-item';
...
+   * @required
+   */
+  public readonly modalComponent =
+    input.required<Type<SkyFilterItemModal<TData>>>();
+
+  /**
+   * The size of the modal to display. The valid options are `"small"`, `"medium"`, `"large"`, and `"fullScreen"`.
+   * @default "medium"
+   */
+  public readonly modalSize = input<SkyFilterItemModalSizeType>();
```
Blackbaud-SteveBrush: **Public API:** There was some prior feedback to make this input required. Reasoning: 
1) we don't want to lean on the modal service's default modal size (we say "medium" in the JSDoc comment, but `sky-filter-item-modal` wouldn't "know" that necessarily)
2) with new components, it's preferred to make most (if not all) inputs required so that we don't have to be concerned with default values on initial release. Optional values add branching logic, so if we can guarantee that a value is set, it reduces the overall complexity of the component (we use this same logic with new interfaces).

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/example.component.html
```diff
@@ -0,0 +1,35 @@
...
+    filterId="role"
+    labelText="Role"
+    [modalComponent]="roleModal"
+    (modalOpened)="onModalOpened($event, 'role')"
+  />
+  <sky-filter-item-modal
+    filterId="staff-assigned"
+    labelText="Staff assigned"
+    [modalComponent]="staffAssignedModal"
+    (modalOpened)="onModalOpened($event, 'staff-assigned')"
```
Blackbaud-SteveBrush: **Public API:** I know you're passing in the `filterId` to this callback for the sake of the code example, but I can imagine the consumer needing access to the `filterId` in a real-world scenario. We should add a `readonly filterId` property to the `SkyFilterItemModalOpenedArgs` object and pass it to the `onModalOpened` callback automatically; that way, the consumer can't accidentally set two different `filterId` values (one for the input, and one for the `modalOpened` callback).
```
(modalOpened)="onModalOpened($event)"
```

An example "real-world" scenario:
```
protected onModalOpened(
  args: SkyFilterItemModalOpenedArgs<FilterItems>
): void {
  const { filterId } = args;
  args.data = this.#http.get(`https://api/filters/${filterId}/options`).subscribe();
}
```

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modals/community-connection-filter-modal.component.ts
```diff
@@ -0,0 +1,59 @@
...
+    SkyInputBoxModule,
+    SkyModalModule,
+    SkyRepeaterModule,
+  ],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class CommunityConnectionFilterModalComponent
+  implements SkyFilterItemModal<FilterItems>
+{
+  public readonly modalInstance = inject<
```
Blackbaud-SteveBrush, line 33: Can be `protected`?

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modals/current-grade-filter-modal.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    SkyInputBoxModule,
+    SkyModalModule,
+    SkyRepeaterModule,
+  ],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class CurrentGradeFilterModalComponent
+  implements SkyFilterItemModal<FilterItems>
+{
+  public readonly modalInstance = inject<
```
Blackbaud-SteveBrush, line 33: Can be `protected`?

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modals/entering-grade-filter-modal.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    SkyInputBoxModule,
+    SkyModalModule,
+    SkyRepeaterModule,
+  ],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class EnteringGradeFilterModalComponent
+  implements SkyFilterItemModal<FilterItems>
+{
+  public readonly modalInstance = inject<
```
Blackbaud-SteveBrush, line 33: Can be `protected`?

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modals/role-filter-modal.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    SkyInputBoxModule,
+    SkyModalModule,
+    SkyRepeaterModule,
+  ],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class RoleFilterModalComponent
+  implements SkyFilterItemModal<FilterItems>
+{
+  public readonly modalInstance = inject<
```
Blackbaud-SteveBrush, line 33: Can be `protected`?

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modals/staff-assigned-filter-modal.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    SkyInputBoxModule,
+    SkyModalModule,
+    SkyRepeaterModule,
+  ],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class StaffAssignedFilterModalComponent
+  implements SkyFilterItemModal<FilterItems>
+{
+  public readonly modalInstance = inject<
```
Blackbaud-SteveBrush, line 33: Can be `protected`?

libs/components/filter-bar/src/index.ts
```diff
@@ -1,9 +1,12 @@
...
 
 export { SkyFilterBarFilterItem } from './lib/modules/filter-bar/models/filter-bar-filter-item';
-export { SkyFilterBarFilterModalConfig } from './lib/modules/filter-bar/models/filter-bar-filter-modal-config';
-export { SkyFilterBarFilterModalContext } from './lib/modules/filter-bar/models/filter-bar-filter-modal-context';
 export { SkyFilterBarFilterValue } from './lib/modules/filter-bar/models/filter-bar-filter-value';
+export { SkyFilterItemModal } from './lib/modules/filter-bar/models/filter-item-modal';
+export { SkyFilterItemModalContext } from './lib/modules/filter-bar/models/filter-item-modal-context';
+export { SkyFilterItemModalInstance } from './lib/modules/filter-bar/models/filter-item-modal-instance';
+export { SkyFilterItemModalOpenedArgs } from './lib/modules/filter-bar/models/filter-item-modal-opened-args';
+export { SkyFilterItemModalSizeType } from './lib/modules/filter-bar/models/filter-item-modal-size';
```
Blackbaud-SteveBrush, line 10: Thanks for cleaning up these names. 👍

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-item-modal-context.ts
```diff
@@ -0,0 +1,30 @@
...
+  public filterLabelText: string;
+  /**
+   * The value of the filter.
+   */
+  public filterValue?: SkyFilterBarFilterValue;
+  /**
+   * An untyped property that can track any config information relevant to the filter modal
+   * that existing options do not include.
+   */
+  public additionalContext?: TData;
```
Blackbaud-SteveBrush: ☝️ Is it possible to make all public properties `readonly`?

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-item-modal-instance.ts
```diff
@@ -0,0 +1,49 @@
...
+   * Closes the modal instance with `reason="cancel"`.
+   */
+  public cancel(result?: SkyFilterBarFilterValue): void {
+    this.#canceled.next(result);
+    this.#canceled.complete();
+  }
+
+  /**
+   * Closes the modal instance with `reason="save"`.
+   * @param args
```
Blackbaud-SteveBrush, line 22: The `@param args` doesn't match the name of the `result` parameter. Maybe just remove it?

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-item-modal-instance.ts
```diff
@@ -0,0 +1,49 @@
...
+  public readonly context = inject<SkyFilterItemModalContext<TData>>(
+    SkyFilterItemModalContext,
+  );
+
+  readonly #canceled = new Subject<SkyFilterBarFilterValue | undefined>();
+  readonly #saved = new Subject<SkyFilterBarFilterValue | undefined>();
+
+  /**
+   * Fires when the user cancels the modal.
+   * @internal
```
Blackbaud-SteveBrush: **Public API:** I know my StackBlitz prototype included the `canceled` and `saved` methods inside the instance, but it would be ideal if the consumer didn't have access to them at all. We want an API like this:
Screenshot 2025-09-12 at 14 55 50

Since we're using `runInInjectionContext`, we could create the context and the observables in the `SkyFilterItemModalComponent` before the modal instance is constructed. Then the consumer wouldn't have access to the public methods at all.

**filter-item-modal-instance.ts** (largely the same, just use an abstract class instead)
```
/**
 * A specialized `SkyModalInstance` wrapper.
 */
export abstract class SkyFilterItemModalInstance<
  TData = Record<string, unknown>,
> {
  /**
   * The context provided to the filter modal component.
   */
  public abstract readonly context: SkyFilterItemModalContext<TData>;

  /**
   * Closes the modal instance with `reason="cancel"`.
   */
  public abstract cancel(): void;

  /**
   * Closes the modal instance with `reason="save"`.
   */
  public abstract save(args: SkyFilterItemModalSaveArgs): void;
}
```

**filter-item-modal.component.ts**
```
runInInjectionContext(injector, () => {
  const destroyRef = injector.get(DestroyRef);

  // Create the observables.
  const saved = new Subject<unknown>();
  const canceled = new Subject<void>();

  // Create the context.
  const context = inject<SkyFilterItemModalContext<TData>>(
    SkyFilterItemModalContext,
  );

  // Create the instance.
  const filterModalInstance: SkyFilterItemModalInstance<TData> = {
    context,
    cancel() {
      canceled.next();
      canceled.complete();
    },
    save(result) {
      saved.next(result);
      saved.complete();
    },
  };

  canceled
    .pipe(takeUntilDestroyed(destroyRef))
    .subscribe(() => instance.cancel());

  saved
    .pipe(takeUntilDestroyed(destroyRef))
    .subscribe((result) => instance.save(result));

  const modalConfig: SkyModalConfigurationInterface = {
    providers: [
      {
        provide: SkyFilterItemModalInstance,
        useValue: filterModalInstance,
      },
    ],
  };

  // ...
```

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-item-modal-instance.ts
```diff
@@ -0,0 +1,49 @@
...
+   * Fires when the user saves the modal.
+   * @internal
+   */
+  public readonly saved = this.#saved.asObservable();
+
+  /**
+   * Closes the modal instance with `reason="cancel"`.
+   */
+  public cancel(result?: SkyFilterBarFilterValue): void {
+    this.#canceled.next(result);
```
Blackbaud-SteveBrush: **Public API:** What information would the consumer need to pass in a "cancel" event? I think we'd be safe to have a simple `cancel()` method without any arguments.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-item-modal-instance.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.#canceled.complete();
+  }
+
+  /**
+   * Closes the modal instance with `reason="save"`.
+   * @param args
+   */
+  public save(result?: SkyFilterBarFilterValue): void {
```
Blackbaud-SteveBrush: **Public API:** This `result` parameter limits us to only one property type, `SkyFilterBarFilterValue`. Let's create a `SkyFilterBarModalSaveArgs` interface that includes a required `filterValue: SkyFilterBarFilterValue | undefined` property. This will inform the consumer to save either a value, or deliberately set `undefined` to unset the value. 

Now, in the future, if we need to add another property, it can be done easily without changing the API.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-item-modal.ts
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * A type marker for passing context object data types into the filter modal component.
+ */
+export interface SkyFilterItemModal<
+  TData = Record<string, unknown> | undefined,
+> {
+  /**
+   * The filter modal instance to be injected into the component.
+   */
+  readonly modalInstance: SkyFilterItemModalInstance<TData>;
```
Blackbaud-SteveBrush, line 12: Good idea, and helpful to our consumers. 👍

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modals/community-connection-filter-modal.component.ts
Blackbaud-CoreyArcher: No, properties on the interface must be implemented as public.

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modals/current-grade-filter-modal.component.ts
Blackbaud-CoreyArcher: No, properties on the interface must be implemented as public.

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modals/entering-grade-filter-modal.component.ts
Blackbaud-CoreyArcher: No, properties on the interface must be implemented as public.

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modals/role-filter-modal.component.ts
Blackbaud-CoreyArcher: No, properties on the interface must be implemented as public.

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/filter-modals/staff-assigned-filter-modal.component.ts
Blackbaud-CoreyArcher: No, properties on the interface must be implemented as public.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-item-modal-instance.ts
Blackbaud-CoreyArcher: reverted argument to `args`

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-item-modal-instance.ts
Blackbaud-CoreyArcher: converted to abstract class/implementation

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-item-modal-instance.ts
Blackbaud-CoreyArcher: removed

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-item-modal-instance.ts
Blackbaud-CoreyArcher: added

libs/components/code-examples/src/lib/modules/filter-bar/filter-bar/basic/example.component.html
Blackbaud-CoreyArcher: added

libs/components/filter-bar/src/lib/modules/filter-bar/filter-item-modal.component.ts
Blackbaud-CoreyArcher: made required

Blackbaud-SteveBrush (approved):

blackbaud-sky-build-user:
Cherry-pick successful! :cherries: :tada: See https://github.com/blackbaud/skyux/pull/3934

PR #3774, opened 2025-07-30
chore(components/filter-bar): convert to template-driven API

Blackbaud-CoreyArcher:


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/3774/storybook/)

Component Storybooks:

- [filter-bar-storybook](https://blackbaud.github.io/skyux-pr-preview/3774/storybooks/filter-bar-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/3774/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/3774/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/3774/playground/)

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
```diff
@@ -44,13 +49,44 @@ type SelectionModalSearchAsyncFn = (
...
+   */
   public filters = model<SkyFilterBarFilterItem[] | undefined>();
-  public filterAsyncSearchFn = input<SelectionModalSearchAsyncFn>();
 
-  protected hasFilters = computed(() => {
-    const filters = this.filters();
+  /**
+   * Whether to include a button to launch a selection modal that adds and removes filters.
+   */
+  public enableSelectionModal = input<boolean>(false);
```
Blackbaud-SteveBrush: Can we name the input in a way that allows for shorthand usage of a `booleanAttribute` transform? I would also remove the word "modal" from the name in case we choose a different mechanism in the future to select filters.

If the filter selection button is hidden by default:
```
<sky-filter-bar showFilterSelector />
```

If the filter selection button is shown by default:
```
<sky-filter-bar hideFilterSelector />
```

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
```diff
@@ -44,13 +49,44 @@ type SelectionModalSearchAsyncFn = (
   styleUrl: './filter-bar.component.scss',
 })
 export class SkyFilterBarComponent {
+  /**
+   * The filters to be displayed on the filter bar. Filters defined in a provided `SkyDataManagerService` will supersede this input.
+   */
   public filters = model<SkyFilterBarFilterItem[] | undefined>();
```
Blackbaud-SteveBrush: From what I can gather, this input is combining two separate concepts into a single `model`: 
1) configures which filters to display, and 
2) emits an event when filter _values_ are applied.  

The issue with combining these into one `model` is that the user may select different filters to show using the filter selector, but we wouldn't want to emit those filters with the `filtersChange` event because they'll have `undefined` values and wouldn't be relevent to a service that only cares about applied filters (unless the filter selection removes a filter that is already applied).

To communicate these distinctions, what do you think about separating this into a `selectedFilters` input, and a `filtersApplied` output? The developer would set `selectedFilters` to determine which filters are shown by default (with any initial values), and listen to `filtersApplied` when filter values are changed by the end user.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar-item.component.ts
```diff
@@ -1,63 +1,35 @@
...
-      };
-      if (config.modalSize === 'full') {
-        modalConfig.fullPage = true;
-      } else {
-        modalConfig.size = config.modalSize;
-      }
+  /**
+   * The name of the filter displayed in the text label.
+   */
+  public filterName = input.required<string>();
```
Blackbaud-SteveBrush: We use the word "text" in our other components (e.g. `labelText`, `headingText`). The keyword "name" is a bit overloaded in JavaScript, so I'd opt for `labelText` to keep it consistent with our other components, and make it self-documenting.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar-item.component.ts
```diff
@@ -1,63 +1,35 @@
...
 import { SkyFilterBarFilterModalConfig } from './models/filter-bar-filter-modal-config';
-import { SkyFilterBarFilterModalContext } from './models/filter-bar-filter-modal-context';
-import { SkyFilterBarFilterValue } from './models/filter-bar-filter-value';
 
 /**
- * @internal
+ * A component used to define a filter bar item configuration.
  */
 @Component({
   selector: 'sky-filter-bar-item',
```
Blackbaud-SteveBrush: Similar to the toolbar and toolbar-item components, I could see a filter bar "item" being used for other things like separators, sections, or various buttons and controls in the future. Since this component is purely for configuring a filter, does the name `sky-filter` work, instead?

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar-item.component.ts
```diff
@@ -1,63 +1,35 @@
...
-          this.filterValue.set(args.data);
-          this.filterUpdated.emit(args.data);
-        }
-      });
-    }
-  }
+  /**
+   * The configuration options for showing a `SkySelectionModal`.
+   */
+  public filterSelectionModalConfig = input<SkySelectionModalOpenArgs>();
```
Blackbaud-SteveBrush: Do we need to give the consumer access to the selection modal config? What aspect of the filter selection needs to be configurable, from the perspective of the filter bar? For MVP, it's probably better to go out with the smallest API surface area as possible, so I would opt to remove this for now.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar-item.component.ts
```diff
@@ -1,63 +1,35 @@
...
+  /**
+   * The name of the filter displayed in the text label.
+   */
+  public filterName = input.required<string>();
 
-      const instance = this.#modalSvc.open(config.modalComponent, modalConfig);
+  /**
+   * The configuration options for showing a custom filter modal.
+   */
+  public filterModalConfig = input<SkyFilterBarFilterModalConfig>();
```
Blackbaud-SteveBrush: I would avoid complex objects for inputs, since it makes it difficult for change detection and managing state. Can we have a simple `modalRef` input here, or something similar? The `SkyFilterBarFilterModalConfig` interface includes options for the modal itself, and for additional injection context. It would probably be best to exclude complexity in the public API until we know a specific feature is needed. For example, if modal config is needed in the future, we could introduce a new input for that.

Blackbaud-SteveBrush: Something else to think through: We're going to have three types of filter buttons for the MVP release:

1. Custom filters (modal) - value is `unknown`
2. Boolean filters (checkbox) - value is `boolean | undefined`
3. Select filters (dropdown) - value is one of `choices` or `undefined`

What's your proposed API for how we handle types 2 and 3? I know they're in separate stories, but it would be good to have an idea in mind for those before moving forward with this PR since it will inform how we construct the needed components.

What we don't want is to have one component that includes all possible features for a filter because the branching logic in the component class would become unmanageable over time. For example, we wouldn't want one component that has inputs for both a modal and a dropdown to avoid consumers mixing and matching incompatible features.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar-button.component.ts
```diff
@@ -0,0 +1,110 @@
...
+  imports: [CommonModule, SkyIconModule],
+  templateUrl: './filter-bar-button.component.html',
+  styleUrl: './filter-bar-button.component.scss',
+})
+export class SkyFilterBarButtonComponent {
+  public filterId = input.required<string>();
+  public filterName = input.required<string>();
+  public filterValue = model<SkyFilterBarFilterValue>();
+  public filterModalConfig = input<SkyFilterBarFilterModalConfig>();
+  public filterSelectionModalConfig = input<SkySelectionModalOpenArgs>();
```
Blackbaud-SteveBrush: I imagine we'd have a dedicated component for the filter selection button, to keep the logic for the selection modal out of the standard filter button component. Possible?

libs/components/filter-bar/src/index.ts
```diff
@@ -4,8 +4,10 @@ export { SkyFilterBarFilterItem } from './lib/modules/filter-bar/models/filter-b
 export { SkyFilterBarFilterModalConfig } from './lib/modules/filter-bar/models/filter-bar-filter-modal-config';
 export { SkyFilterBarFilterModalContext } from './lib/modules/filter-bar/models/filter-bar-filter-modal-context';
 export { SkyFilterBarFilterValue } from './lib/modules/filter-bar/models/filter-bar-filter-value';
+export { SkyFilterBarItem } from './lib/modules/filter-bar/models/filter-bar-item';
 
 // Components and directives must be exported to support Angular's "partial" Ivy compiler.
 // Obscure names are used to indicate types are not part of public API.
+export { SkyFilterBarButtonComponent as λ3 } from './lib/modules/filter-bar/filter-bar-button.component';
```
Blackbaud-SteveBrush: Does this component need to be exported since it's only used internally?

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
```diff
@@ -44,13 +49,44 @@ type SelectionModalSearchAsyncFn = (
   styleUrl: './filter-bar.component.scss',
 })
 export class SkyFilterBarComponent {
```
Blackbaud-SteveBrush, line 53: For all of these new components, it would be nice to see some usage of modern Angular practices, such as using the host element as the primary wrapper, instead of creating a `<div class="sky-filter-bar">` wrapper in the template. You could make the `:host` element `display: block` in the stylesheet and remove the DIV altogether:
```
:host { display: block; }
```
This will simplify the HTML template by reducing the number of elements a given component is responsible for.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
```diff
@@ -143,14 +206,32 @@ export class SkyFilterBarComponent {
...
+  #filterAsyncSearchFn = (
+    args: SkySelectionModalSearchArgs,
+  ): Observable<SkySelectionModalSearchResult> | undefined => {
+    const items = this.filterItems().map((item) => ({
+      id: item.filterId(),
+      name: item.filterName(),
+    }));
+    const results = args.searchText
+      ? items.filter((item) =>
+          item.name.toUpperCase().includes(args.searchText.toUpperCase()),
```
Blackbaud-SteveBrush: Can we use `toLocaleUpperCase` to account for locale-specific characters?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleUpperCase

Blackbaud-SteveBrush (commented):
**Some more general questions:**
1. I noticed in the design proposal that there should be considerations for long lists of filter buttons, whether they should wrap or collapse. What did you and Matt decide for that?
2. I might've missed it, but is UI Config implemented into the filter bar? While it's not explicitly mentioned in the story, it's a necessary feature we should implement for MVP.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
Blackbaud-CoreyArcher: obsolete due to other changes

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
Blackbaud-CoreyArcher: converted to separate models

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar-item.component.ts
Blackbaud-CoreyArcher: updated

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar-item.component.ts
Blackbaud-CoreyArcher: obsolete

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar-item.component.ts
Blackbaud-CoreyArcher: removed for future story

libs/components/filter-bar/src/index.ts
Blackbaud-CoreyArcher: Removed. I have a question about the `SkyFilterItem` export above. that doesn't need to be here either since it's a convenience interface used internally by the filter bar component, right?

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
Blackbaud-CoreyArcher: fixed

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
Blackbaud-CoreyArcher: fixed

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-item.ts
```diff
@@ -0,0 +1,20 @@
+import { SkyFilterBarFilterValue } from './filter-bar-filter-value';
+
+/**
+ * A configuration object for how a filter is displayed on the filter bar.
+ * @internal
+ */
+export interface SkyFilterBarItem {
```
Blackbaud-SteveBrush, line 7: I noticed three similarly named types: `SkyFilterItem`, `SkyFilterBarItem`, and `SkyFilterBarFilterItem`. Without digging into the code, it wasn't immediately obvious their purpose and how they differ. It would help the developer/consumer experience if we came up with more descriptive names for each that better communicated their purpose. Again, Copilot might be able to provide some ideas.

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-item.ts
```diff
@@ -0,0 +1,20 @@
...
+  /**
+   * A unique identifier for the filter.
+   */
+  id: string;
+  /**
+   * The human-readable display string for the filter.
+   */
+  name: string;
```
Blackbaud-SteveBrush: Is there a reason we can't use the input names for these properties for consistency? `filterId`, `labelText`?

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-item.ts
```diff
@@ -1,9 +1,6 @@
-import { SkyFilterBarFilterModalConfig } from './filter-bar-filter-modal-config';
 import { SkyFilterBarFilterValue } from './filter-bar-filter-value';
 
 export interface SkyFilterBarFilterItem {
   id: string;
```
Blackbaud-SteveBrush: Can we name this `filterId` to be consistent with the input name?

libs/components/filter-bar/testing/src/modules/filter-bar/filter-item-harness-filters.ts
```diff
@@ -1,16 +1,16 @@
...
   /**
-   * Finds a filter bar item whose id matches the given value.
+   * Finds a filter item whose id matches the given value.
    */
   id?: string;
   /**
-   * Finds a filter bar item whose name matches the given value.
+   * Finds a filter item whose name matches the given value.
    */
   name?: string;
```
Blackbaud-SteveBrush: Are these matching against the "id" and "name" attributes of a native element? If they're meant to match against the input values, can we use the input names instead? `filterId`, `labelText`?

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-item.ts
```diff
@@ -0,0 +1,11 @@
+import { Signal, TemplateRef } from '@angular/core';
+
+import { SkyFilterBarFilterValue } from './filter-bar-filter-value';
+
+export interface SkyFilterItem {
+  filterId: Signal<string>;
+  labelText: Signal<string>;
+  filterValue: Signal<SkyFilterBarFilterValue | undefined>;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  templateRef: Signal<TemplateRef<any> | undefined>;
```
Blackbaud-SteveBrush: This can be `unknown` instead of `any`.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
```diff
@@ -16,44 +24,69 @@ import {
...
-
-type SelectionModalSearchAsyncFn = (
-  args: SkySelectionModalSearchArgs,
-) => Observable<SkySelectionModalSearchResult> | undefined;
+import { SkyFilterBarItem } from './models/filter-bar-item';
+import { SkyFilterItem } from './models/filter-item';
 
 /**
- * @internal
+ * The top-level filter bar component.
```
Blackbaud-SteveBrush, line 38: This description will be shown on the docs. Is there a better sentence we can use to describe the purpose of the filter bar and what it does for the consumer? Maybe Copilot can help summarize.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
```diff
@@ -16,44 +24,69 @@ import {
...
+   */
   public filters = model<SkyFilterBarFilterItem[] | undefined>();
-  public filterAsyncSearchFn = input<SelectionModalSearchAsyncFn>();
 
-  protected hasFilters = computed(() => {
-    const filters = this.filters();
+  /**
+   * An array of filter IDs that the user has selected using the built-in selection modal. Setting this input to undefined results in all filters being displayed.
+   */
+  public selectedFilters = model<string[] | undefined>();
```
Blackbaud-SteveBrush: The type for `selectedFilters` is `string[]`. Without reading the JSDoc comment, it might be confusing what `string` represents (`filterId`? `labelText`? `value`?). Can we name the input something that's self-documenting? `selectedFilterIds`?

libs/components/filter-bar/src/lib/modules/filter-bar/filter-item-modal.component.ts
```diff
@@ -0,0 +1,101 @@
...
+  public labelText = input.required<string>();
+  /**
+   * The configuration options for showing a custom filter modal.
+   */
+  public filterModalConfig = input.required<SkyFilterBarFilterModalConfig>();
+  /**
+   * The template of the filter that is rendered by the filter bar.
+   * @internal
+   */
+  public templateRef = viewChild(TemplateRef<unknown>);
```
Blackbaud-SteveBrush: Needs `readonly` so that consuming components don't accidentally override it.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-item-modal.component.ts
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * A unique identifier for the filter.
+   */
+  public filterId = input.required<string>();
+  /**
+   * The label for the filter.
+   */
+  public labelText = input.required<string>();
+  /**
+   * The configuration options for showing a custom filter modal.
```
Blackbaud-SteveBrush, line 40: You can move these JSDoc comments to the `SkyFilterItem` interface and they'll automatically carry through to the implementing component. Putting the descriptions on the interface properties will also make it easier for future filter items as they're created.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
```diff
@@ -66,55 +99,53 @@ export class SkyFilterBarComponent {
...
+    });
+
+    // Push filter value
+    effect(() => {
+      const filters = this.filters();
+      this.#updateFilters(filters);
+    });
+  }
+
+  public openFilters(): void {
```
Blackbaud-SteveBrush: I'm seeing a handful of public properties/methods on the public API surface area that should probably be marked as `protected`. Do you mind auditing the `sky-filter-bar` and `sky-filter-item-modal` components for public properties and set them to `protected` where appropriate? We have plans to expose our component classes to the public API in the future, and making these decisions now will make that process easier.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
```diff
@@ -16,44 +24,69 @@ import {
...
   ],
+  providers: [SkyFilterBarService],
   templateUrl: './filter-bar.component.html',
   styleUrl: './filter-bar.component.scss',
 })
 export class SkyFilterBarComponent {
+  /**
+   * An array of filter items containing the IDs and values of the filters that have been set.
+   */
   public filters = model<SkyFilterBarFilterItem[] | undefined>();
```
Blackbaud-SteveBrush: Any `public` inputs, outputs, models, etc. should be marked as `readonly` so that they can't be overridden by a consumer or host component. I saw a few in the `sky-filter-item-modal` component as well.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
```diff
@@ -66,55 +99,53 @@ export class SkyFilterBarComponent {
     }),
   );
 
-  protected openFilters(): void {
-    const searchFn = this.filterAsyncSearchFn();
+  constructor() {
+    // Subscribe to child filter updates
```
Blackbaud-SteveBrush: I would either remove these two comments (this one and the "Push filter value"), or provide a more detailed explanation of what the two `effect` calls are doing. The comments as they stand didn't help me much when I was trying to figure out the purpose of the code they're describing.

libs/components/filter-bar/src/lib/modules/filter-bar/filter-item-modal.component.ts
```diff
@@ -0,0 +1,101 @@
...
+import { switchMap } from 'rxjs/operators';
+
+import { SkyFilterBarService } from './filter-bar.service';
+import { SKY_FILTER_ITEM } from './filter-item.token';
+import { SkyFilterBarFilterModalConfig } from './models/filter-bar-filter-modal-config';
+import { SkyFilterBarFilterModalContext } from './models/filter-bar-filter-modal-context';
+import { SkyFilterItem } from './models/filter-item';
+
+/**
+ * A filter bar item component that can display a filter in a modal.
```
Blackbaud-SteveBrush: Since this comment will appear in the docs, I'd spend some time aligning the "voice" and structure of the description with what we're doing for other components. I'd also offer a bit more to the consumer as far as what the component does and when to use it. Maybe John Lyons can help (or Copilot). Here's what Copilot generated for me:
```
/**
 * A filter bar item that opens a modal for complex filter configuration.
 * Use this component when your filter requires a rich UI with multiple inputs,
 * date pickers, or other complex controls that don't fit in an inline filter.
 */ 
```

libs/components/filter-bar/src/index.ts
Blackbaud-SteveBrush: I'm honestly unsure. I would try exporting it and not exporting it, and see if the JSDoc comments look different when serving the `code-examples-playground` SPA.

Blackbaud-CoreyArcher: removed for now

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-item.ts
Blackbaud-CoreyArcher: Changed

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-item.ts
Blackbaud-CoreyArcher: Updated

libs/components/filter-bar/testing/src/modules/filter-bar/filter-item-harness-filters.ts
Blackbaud-CoreyArcher: Changed

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-item.ts
Blackbaud-CoreyArcher: Fixed

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
Blackbaud-CoreyArcher: This will be explored with SCD before the component releases

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
Blackbaud-CoreyArcher: Changed

libs/components/filter-bar/src/lib/modules/filter-bar/filter-item-modal.component.ts
Blackbaud-CoreyArcher: Fixed

libs/components/filter-bar/src/lib/modules/filter-bar/filter-item-modal.component.ts
Blackbaud-CoreyArcher: Moved

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
Blackbaud-CoreyArcher: Removed all unneeded public methods and made all inputs `readonly`

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
Blackbaud-CoreyArcher: Fixed

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
Blackbaud-CoreyArcher: Updated

libs/components/filter-bar/src/lib/modules/filter-bar/filter-item-modal.component.ts
Blackbaud-CoreyArcher: Updated

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-item.ts
Blackbaud-CoreyArcher: Only one of these, `SkyFilterBarFilterItem`, is exposed to consumers. The others are documented for their purposes, and I've added JSDocs comments for the external interface. `SkyFilterItem` is the interface of how `sky-filter-item-<type>` components are to be constructed. I could see a case for renaming `SkyFilterBarItem` but once its purpose as a representation of filter items on the filter bar is understood I can't think of a more concise name for it.

Blackbaud-SteveBrush (approved):

blackbaud-sky-build-user:
Cherry-pick [failed](https://github.com/blackbaud/skyux/actions/runs/17042525653)! :x: Please resolve conflicts and create a new PR.

PR #3895, opened 2025-09-04
feat(components/packages): create schematic to remove fixedWidth input on sky-icon

johnhwhite:
[AB#3537577](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3537577)

### DIFF
--- a/libs/components/packages/src/schematics/migrations/update-13/icon-name/icon-name.schematic.spec.ts
+++ b/libs/components/packages/src/schematics/migrations/update-13/icon-name/icon-name.schematic.spec.ts
@@ -124,6 +124,18 @@ describe('icons-name.schematic', () => {
       },
       {
         invalidTemplate: `
+      <sky-icon iconName="add" fixedWidth="true"/>
+      <sky-icon iconName="add" [fixedWidth]="true"/>
+      <sky-icon iconName="add" fixedWidth/>
+      `,
+        convertedTemplate: `
+      <sky-icon iconName="add" />
+      <sky-icon iconName="add" />
+      <sky-icon iconName="add" />
+      `,
+      },
+      {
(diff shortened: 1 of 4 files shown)


libs/components/packages/src/schematics/migrations/update-13/icon-name/icon-name.schematic.spec.ts
```diff
@@ -122,6 +122,18 @@ describe('icons-name.schematic', () => {
       <sky-icon iconName="bb-diamond" variant="solid"></sky-icon>
       `,
       },
+      {
+        invalidTemplate: `
+      <sky-icon iconName="add" fixedWidth="true"/>
+      <sky-icon iconName="add" [fixedWidth]="true"/>
+      <sky-icon iconName="add" fixedWidth/>
```
Blackbaud-TrevorBurch: This is a good test - but should we also have a case that checks converting icon and removing fixed width at the same time? 

And maybe one where `icon` can't be converted but `fixedWidth` is still removed?

Blackbaud-TrevorBurch:
@johnhwhite - `iconType` is another input that we removed that can just be removed. I wonder if we should add that. 

`size` was also removed - but that one is one that is a little harder because it isn't 1:1 - and I think we want to make the consumer make a decision - but we can flex on it if needed.

libs/components/packages/src/schematics/migrations/update-13/icon-name/icon-name.schematic.spec.ts
johnhwhite: Makes sense. Added more test cases.

Blackbaud-TrevorBurch (approved):

PR #3986, opened 2025-09-22
ci: fix main build

johnhwhite:


libs/sdk/eslint-config-skyux/dev-transpiler.cjs
```diff
@@ -5,7 +5,6 @@ const path = require('node:path');
  * eslint.config.js file.
  */
 require('ts-node').register({
-  transpileOnly: true,
```
Blackbaud-SteveBrush, line 8: The `transpileOnly` flag was removed from TypeDoc, but it's still a valid option for `ts-node`. Can we add it back for these?

Blackbaud-SteveBrush: Went ahead and reverted these.

Blackbaud-SteveBrush (approved):

PR #4029, opened 2025-10-10
feat: integrate filter bar with data manager

Blackbaud-CoreyArcher:
[AB#3364676](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3364676)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4029/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4029/storybooks/ag-grid-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/4029/storybooks/data-manager-storybook/)
- [filter-bar-storybook](https://blackbaud.github.io/skyux-pr-preview/4029/storybooks/filter-bar-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/4029/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/4029/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/4029/storybooks/lookup-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/4029/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/4029/storybooks/phone-field-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/4029/storybooks/split-view-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4029/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4029/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4029/playground/)

libs/components/data-manager/src/lib/modules/data-manager/data-manager-toolbar/data-manager-toolbar.component.html
```diff
@@ -88,6 +88,9 @@
     <ng-content select="sky-data-manager-toolbar-section" />
   </sky-toolbar>
 
+  <ng-content select="sky-filter-bar" />
+  <ng-content select="sky-list-summary" />
```
Blackbaud-CoreyArcher, line 92: Including this here just to avoid merge conflicts with another story

libs/components/lists/src/lib/modules/shared/filter-adapter/filter-adapter.service.ts
```diff
@@ -0,0 +1,27 @@
+import { Observable } from 'rxjs';
+
+import { SkyFilterAdapterData } from './filter-adapter-data';
+
+/**
+ * Abstract service that provides an interface for components to integrate with a data manager.
+ * This allows various list and filter components to work with data managers without tight coupling.
+ * @internal
+ */
+export abstract class SkyFilterAdapterService {
```
Blackbaud-SteveBrush: I noticed the JSDoc comment references “data manager,” but this service is intended to be agnostic: it shouldn’t be positioned as an “adapter” specifically for the data manager. Instead, it should stand on its own as a general-purpose service for managing filter state. To better reflect that distinction, could we use the name `SkyFilterStateService` we discussed a few weeks ago?

For example, the filter bar uses `SkyFilterStateService` to notify parent components when the filter state changes. Meanwhile, the data manager directive provides its own instance of SkyFilterStateService and acts as a bridge between the data manager’s internal state and the filter state service. In that specific context, it makes sense to refer to the provided service as an “adapter,” since it’s fulfilling that bridging role within the directive.

To help illustrate, I copied your implementation into a basic StackBlitz example: https://stackblitz.com/edit/stackblitz-starters-cc7zrb5j?file=src%2Fmodules%2Fdata-manager.ts Note how `SkyDataManagerFilterControllerDirective` provides `SkyFilterStateService` using `SkyDataManagerFilterStateAdapterService`, hence the use of “adapter” in that name.

For the types in `@skyux/lists`, we can stick with the naming convention we aligned on earlier and use “state” instead of “adapter”:
```
export interface SkyFilterState {
  appliedFilters: SkyFilterStateAppliedFilter[];
  selectedFilterIds: string[];
}

export interface SkyFilterStateChange {
  filterState: SkyFilterState;
  sourceId: string;
}

export interface SkyFilterStateFilterValue {
  value: unknown;
  displayValue: string;
}

export interface SkyFilterStateAppliedFilter {
  filterId: string;
  filterValue: SkyFilterStateFilterValue;
}

@Injectable()
export abstract class SkyFilterStateService {
  public abstract getStateChanges(sourceId: string): Observable<SkyFilterState>;
  public abstract setState(state: SkyFilterState, sourceId: string): void;
}
```
You can also view this in context here: https://stackblitz.com/edit/stackblitz-starters-cc7zrb5j?file=src%2Fmodules%2Flists.ts

libs/components/lists/src/lib/modules/shared/filter-state-service/filter-state.ts
```diff
@@ -0,0 +1,16 @@
...
+export interface SkyFilterAdapterData {
+  /**
+   * An array of filter items containing the IDs and values of the filters that have been applied.
+   */
+  appliedFilters?: SkyFilterAdapterFilterItem[];
+
+  /**
+   * An array of filter IDs that the user has selected (for components that support filter selection).
+   */
+  selectedFilterIds?: string[];
```
Blackbaud-SteveBrush, line 15: I think it will be easier for the consumer if these properties are always defined, even if they're only empty arrays. Then they won't have to do a null check, and then a length check. Possible?

libs/components/lists/src/lib/modules/shared/filter-adapter/filter-adapter-filter-item.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+export interface SkyFilterAdapterFilterItem {
+  /**
+   * A unique identifier for the filter item.
+   */
+  filterId: string;
+  /**
+   * The value of the filter item.
+   */
+  filterValue?: SkyFilterAdapterFilterValue;
```
Blackbaud-SteveBrush: I know this is probably copy/pasted from the filter bar implementation, but can you remind me why this property isn't required? What's the scenario that would cause the `filterId` to be defined, but not its `filterValue`?

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
```diff
@@ -85,11 +86,13 @@ export class SkyFilterBarComponent {
...
   });
 
+  readonly #adapterSvc = inject(SkyFilterAdapterService, { optional: true });
   readonly #confirmSvc = inject(SkyConfirmService);
   readonly #filterBarSvc = inject(SkyFilterBarService);
-  readonly #filterUpdated = toSignal(this.#filterBarSvc.filterItemUpdated);
+  readonly #filterItemUpdated = toSignal(this.#filterBarSvc.filterItemUpdated);
   readonly #modalSvc = inject(SkySelectionModalService);
   readonly #resourceSvc = inject(SkyLibResourcesService);
+  readonly #sourceId = 'skyFilterBar';
```
Blackbaud-SteveBrush, line 95: You'll need a unique ID for every instance of the filter bar (what if there are multiple filter bars on the page?). You can use the `SkyIdService` to generate a unique ID.

libs/components/data-manager/src/lib/modules/data-manager/data-manager-filters/data-manager-filter-controller.directive.ts
```diff
@@ -0,0 +1,91 @@
...
+import { SkyDataManagerState } from '../models/data-manager-state';
+
+import { SkyDataManagerFilterAdapterService } from './data-manager-filter-adapter.service';
+
+/**
+ * A directive applied to a filtering component that enables integration with a data manager.
+ */
+@Directive({
+  selector: '[skyDataManagerFilterController]',
+  standalone: true,
```
Blackbaud-SteveBrush: Directives are standalone by default. You can remove this.

libs/components/data-manager/src/lib/modules/data-manager/data-manager-filters/data-manager-filter-controller.directive.ts
```diff
@@ -0,0 +1,91 @@
...
+  ],
+})
+export class SkyDataManagerFilterControllerDirective
+  implements OnInit, OnDestroy
+{
+  #currentDataState!: SkyDataManagerState;
+  readonly #dataManagerService = inject(SkyDataManagerService);
+  readonly #adapterService = inject(SkyFilterAdapterService);
+  readonly #ngUnsubscribe = new Subject<void>();
+  readonly #sourceId = 'skyDataManagerFilterController';
```
Blackbaud-SteveBrush, line 28: Needs to be a unique ID in case there are multiple directives on the same page. You can use the `SkyIdService`.

libs/components/data-manager/src/lib/modules/data-manager/data-manager-filters/data-manager-filter-controller.directive.ts
```diff
@@ -0,0 +1,91 @@
...
+    },
+  ],
+})
+export class SkyDataManagerFilterControllerDirective
+  implements OnInit, OnDestroy
+{
+  #currentDataState!: SkyDataManagerState;
+  readonly #dataManagerService = inject(SkyDataManagerService);
+  readonly #adapterService = inject(SkyFilterAdapterService);
+  readonly #ngUnsubscribe = new Subject<void>();
```
Blackbaud-SteveBrush: It's preferred to use `takeUntilDestroyed` with a reference to the `DestroyRef`.
```
import { DestroyRef } from '@angular/core';
import { takeUntilDestroyed } from '@angular/core/rxjs-interop';

readonly #destroyRef = inject(DestroyRef);

.pipe(takeUntilDestroyed(this.#destroyRef))
```

apps/playground/src/app/components/ag-grid/data-manager/data-view-grid.component.ts
```diff
@@ -188,31 +192,42 @@ export class DataViewGridComponent implements OnInit {
...
-        if (
-          ((filters.hideOrange && item.color !== 'orange') ||
-            !filters.hideOrange) &&
-          ((filters.type !== 'any' && item.type === filters.type) ||
-            !filters.type ||
-            filters.type === 'any')
-        ) {
-          return true;
+    const filterData = this.dataState.filterData?.filters as
+      | SkyFilterAdapterData
```
Blackbaud-SteveBrush: Better way to assign this type?

libs/components/lists/src/lib/modules/shared/filter-adapter/filter-adapter-service-token.ts
```diff
@@ -0,0 +1,9 @@
+import { InjectionToken } from '@angular/core';
+
+import { SkyFilterAdapterService } from './filter-adapter.service';
+
+/**
+ * Injection token for the filter adapter service.
+ */
+export const SKY_FILTER_ADAPTER_SERVICE =
```
Blackbaud-SteveBrush: Is this used anywhere?

libs/components/lists/src/lib/modules/shared/filter-state-service/filter-state-filter-value.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+export interface SkyFilterAdapterFilterValue {
+  /**
+   * The real value for the filter.
+   */
+  value: unknown;
+  /**
+   * A human-readable string for use with values that can't be displayed to the user.
+   */
+  displayValue?: string;
```
Blackbaud-SteveBrush, line 12: Similar question: Is there a use-case for a filter to only have a raw value and not a display value?

libs/components/lists/src/lib/modules/shared/filter-adapter/filter-adapter-data-change.ts
```diff
@@ -0,0 +1,10 @@
+import { SkyFilterAdapterData } from './filter-adapter-data';
+
+/**
+ * Represents a change to the filter adapter data including the source that triggered the change.
+ * @internal
+ */
+export interface SkyFilterAdapterDataChange {
+  data: SkyFilterAdapterData;
+  source: string;
```
Blackbaud-SteveBrush: The methods have a parameter named `sourceId`. We can use that same name here, so the correlation is more obvious.

libs/components/lists/src/lib/modules/shared/filter-adapter/filter-adapter-data-change.ts
```diff
@@ -0,0 +1,10 @@
+import { SkyFilterAdapterData } from './filter-adapter-data';
+
+/**
+ * Represents a change to the filter adapter data including the source that triggered the change.
+ * @internal
+ */
+export interface SkyFilterAdapterDataChange {
+  data: SkyFilterAdapterData;
```
Blackbaud-SteveBrush: `"data"` is a bit imprecise. What about `filterState`?

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-value.ts
```diff
@@ -1,13 +1,6 @@
...
-  /**
-   * The real value for the filter.
-   */
-  value: unknown;
-  /**
-   * A human-readable string for use with values that can't be displayed to the user.
-   */
-  displayValue?: string;
-}
+export type SkyFilterBarFilterValue = SkyFilterAdapterFilterValue;
```
Blackbaud-SteveBrush: Changing a type from `interface` to `type` is technically a breaking change. 
See: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

If you need to share the same properties between the two interfaces, you can use `extends` instead.

libs/components/data-manager/src/lib/modules/data-manager/data-manager-filters/data-manager-filter-controller.directive.ts
```diff
@@ -0,0 +1,91 @@
...
+    {
+      provide: SkyFilterAdapterService,
+      useExisting: SkyDataManagerFilterAdapterService,
+    },
+  ],
+})
+export class SkyDataManagerFilterControllerDirective
+  implements OnInit, OnDestroy
+{
+  #currentDataState!: SkyDataManagerState;
```
Blackbaud-SteveBrush: Data state is created in `ngOnInit`, so I would leave this as `SkyDataManagerState | undefined` so that it's not improperly accessed elsewhere in the directive's lifecycle.

libs/components/code-examples/src/lib/modules/ag-grid/data-grid/data-manager-multiselect/view-grid.component.ts
```diff
@@ -246,20 +245,35 @@ export class ViewGridComponent implements OnInit, OnDestroy {
...
-        return (
-          (!!(filters.hideSales && item.department.name !== 'Sales') ||
-            !filters.hideSales) &&
-          ((filters.jobTitle !== 'any' &&
-            item.jobTitle?.name === filters.jobTitle) ||
-            !filters.jobTitle ||
-            filters.jobTitle === 'any')
-        );
+    const filters = this.#dataState?.filterData?.filters?.appliedFilters as
+      | SkyFilterBarFilterItem[]
```
Blackbaud-SteveBrush: Instead of making consumers assert the type of the `filters?: any` object, you could experiment with generics on the `SkyDataManagerState` class. 

```
#dataState: SkyDataManagerState<SkyDataManagerFilterState> | undefined;
```

The class would assign the `TFilterState` generic to `any` for backward compatibility.
```
export class SkyDataManagerState<TFilterState = any> {
```
You can reference this StackBlitz where I fiddled with the idea (see the `main.ts` file and the `data-manager.ts` file):
https://stackblitz.com/edit/stackblitz-starters-cc7zrb5j?file=src%2Fmodules%2Fdata-manager.ts

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-item.ts
```diff
@@ -1,15 +1,6 @@
...
-  /**
-   * A unique identifier for the filter item.
-   */
-  filterId: string;
-  /**
-   * The value of the filter item.
-   */
-  filterValue?: SkyFilterBarFilterValue;
-}
+export type SkyFilterBarFilterItem = SkyFilterAdapterFilterItem;
```
Blackbaud-SteveBrush: Same feedback as below, RE: `interface` vs `type`.

libs/components/lists/src/lib/modules/shared/filter-adapter/filter-adapter.service.ts
```diff
@@ -0,0 +1,27 @@
...
+   */
+  public abstract getFilterDataUpdates(
+    sourceId: string,
+  ): Observable<SkyFilterAdapterData>;
+
+  /**
+   * Updates the filter adapter data and broadcasts the change to subscribers.
+   * Implementations should emit a new value to `filterDataChange` with the provided data.
+   */
+  public abstract updateFilterData(
```
Blackbaud-SteveBrush: As shown in the StackBlitz above, I’d recommend using “filter state” instead of the more ambiguous “filter data” to better convey what the methods are doing. While the data manager does have its own concept of “filter data,” outside that specific context, “state” is more meaningful, especially in relation to the `SkyFilterStateService`.

Blackbaud-SteveBrush (commented):
Overall, this looks great; the implementation is solid and works as expected (no comments there). I noticed a few minor points related to coding conventions and generics, but the main feedback centers on the naming of the "adapter" classes.

I did notice in the code example, that when filter items wrap, there isn't much margin between the buttons. Is that something the design team could help with?

Screenshot 2025-10-15 at 16 42 34


Happy to talk through any of this if you'd like.

libs/components/lists/src/lib/modules/shared/filter-adapter/filter-adapter.service.ts
Blackbaud-CoreyArcher: converted filter adapter to filter state serve

libs/components/lists/src/lib/modules/shared/filter-state-service/filter-state.ts
Blackbaud-CoreyArcher: I think this is correct; these properties mimic the acceptable values of the filter bar inputs they replace. `selectedFilterIds` being undefined is meaningfully distinct from an empty array. It means that the filter selection button and modal aren't used. For `appliedFilters` the input isn't required. We wouldn't want consumers to have to supply an empty array if no filters are set.

libs/components/lists/src/lib/modules/shared/filter-adapter/filter-adapter-filter-item.ts
Blackbaud-CoreyArcher: A filter item without a value is emitted by the filter item components to the filter bar when users clear a filter value

libs/components/filter-bar/src/lib/modules/filter-bar/filter-bar.component.ts
Blackbaud-CoreyArcher: I don't think this is the case. The data manager service is provided by the consumer to a data manager and there is specific guidance for data manager to avoid using more than one data manager per service instance and I would guess that there's an implicit understanding of only using one filter bar per data manager/data manager toolbar

libs/components/data-manager/src/lib/modules/data-manager/data-manager-filters/data-manager-filter-controller.directive.ts
Blackbaud-CoreyArcher: removed

libs/components/data-manager/src/lib/modules/data-manager/data-manager-filters/data-manager-filter-controller.directive.ts
Blackbaud-CoreyArcher: See above, regarding filter bar component id.

libs/components/data-manager/src/lib/modules/data-manager/data-manager-filters/data-manager-filter-controller.directive.ts
Blackbaud-CoreyArcher: converted to DestroyRef

libs/components/lists/src/lib/modules/shared/filter-adapter/filter-adapter-service-token.ts
Blackbaud-CoreyArcher: removed

libs/components/lists/src/lib/modules/shared/filter-state-service/filter-state-filter-value.ts
Blackbaud-CoreyArcher: If the value for a filter is a string that can be displayed to the user, then displayValue is superfluous. This was a convenience factor for the consumer

libs/components/lists/src/lib/modules/shared/filter-adapter/filter-adapter-data-change.ts
Blackbaud-CoreyArcher: this interface wasn't used and I removed it. based on how we solve the remaining issues I will rename these properties as suggested if we reintroduce this

libs/components/lists/src/lib/modules/shared/filter-adapter/filter-adapter-data-change.ts
Blackbaud-CoreyArcher: see above

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-value.ts
Blackbaud-CoreyArcher: reverted to cascading interfaces

libs/components/data-manager/src/lib/modules/data-manager/data-manager-filters/data-manager-filter-controller.directive.ts
Blackbaud-CoreyArcher: initialized to empty DataManagerState

libs/components/filter-bar/src/lib/modules/filter-bar/models/filter-bar-filter-item.ts
Blackbaud-CoreyArcher: fixed

libs/components/lists/src/lib/modules/shared/filter-adapter/filter-adapter.service.ts
Blackbaud-CoreyArcher: converted filterAdapter to filterState

Blackbaud-SteveBrush (dismissed):

Blackbaud-SteveBrush (approved):

PR #4057, opened 2025-10-23
fix(components/lists): remove non-selectable repeater item from tab order and update role of repeater item with only context-menu to `grid`

Blackbaud-SandhyaRajasabeson:
[AB#3430131](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3430131)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4057/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4057/storybooks/ag-grid-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/4057/storybooks/data-manager-storybook/)
- [filter-bar-storybook](https://blackbaud.github.io/skyux-pr-preview/4057/storybooks/filter-bar-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/4057/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/4057/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/4057/storybooks/lookup-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/4057/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/4057/storybooks/phone-field-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/4057/storybooks/split-view-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4057/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4057/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4057/playground/)

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
```diff
@@ -400,17 +400,15 @@ export class SkyRepeaterComponent
...
+    // Determine a role using a hierarchy based on https://www.w3.org/WAI/ARIA/apg/
+    //   1. If there are one or more interactions in the repeater item projected content or there is a context menu, use grid.
     //   3. If there are no interactions, use list.
 
     // Default to list role.
     let autoRole: SkyRepeaterRoleType = 'list';
 
     const roleMap: Record<SkyRepeaterRoleType, SkyRepeaterItemRolesType> = {
       list: { item: 'listitem', title: undefined, content: undefined },
-      listbox: { item: 'option', title: undefined, content: undefined },
```
Blackbaud-SandhyaRajasabeson, line 413: According to the W3 docs, the role `listbox` does not provide an accessible way to handle `option` items inside of it that have interact-able  elements (like checkboxes in selectable repeaters). So there are no valid versions of skyux repeater that would be a `listbox`

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
```diff
@@ -55,17 +55,16 @@ export class SkyRepeaterItemComponent
...
-   * - Disabled items should not be focusable per [W3C](https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_disabled_controls).
-   * - One item per list/grid/listbox should be tab focusable per [W3C](https://www.w3.org/TR/wai-aria-practices-1.1/#grid).
+   * Make the first, non-disabled item tab-focusable in a selectable repeater.
+   * - Disabled items should not be focusable per [W3C](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols).
+   * - One item per grid should be tab focusable per [W3C](https://www.w3.org/WAI/ARIA/apg/patterns/grid/).
    */
   public get tabindex(): 0 | -1 {
     return this.#repeaterService.items.filter((item) => !item.disabled)[0] ===
-      this
+      this && this.selectable
```
Blackbaud-SandhyaRajasabeson: in selectable repeaters: A consumer tabbing to a repeater would have focus initially set on the entire first repeater item. They can then tab through the repeater items using arrows.

in non selectable repeaters with interactable content: the first interactable element will receive focus, arrow keys are not part of keyboard interaction for this instance.

- It is expected that a consumer he puts focus on a repeater will have to tab through all interactable elements before leaving repeater.

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
```diff
@@ -400,17 +400,15 @@ export class SkyRepeaterComponent
   }
 
   #updateRole(): void {
-    // Determine a role using a hierarchy based on https://www.w3.org/TR/wai-aria-practices-1.1/
-    //   1. If there are one or more interactions in the repeater item projected content, use grid.
-    //   2. If there are selectable repeater items and no other interactions, use listbox.
+    // Determine a role using a hierarchy based on https://www.w3.org/WAI/ARIA/apg/
+    //   1. If there are one or more interactions in the repeater item projected content or there is a context menu, use grid.
```
Blackbaud-SteveBrush: The comment reads funny now that the second item is removed (only has "1." and "3."). Numbered lists should have more than 2 items, so in this case, we should use standard bullets. 
```
- If there are one or more...
- If there are no interactions...
```

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
```diff
@@ -435,7 +433,7 @@ export class SkyRepeaterComponent
         (selector) =>
           `sky-repeater-item-title ${selector}:not([hidden]), sky-repeater-item-content ${selector}:not([hidden])`,
       )
-      .concat([`skyux-dropdown`])
+      .concat([`sky-dropdown`])
```
Blackbaud-SteveBrush, line 436: Thanks for fixing this, but I don't think adding `sky-dropdown` is even necessary since `button` is covered in the main array. We can probably safely remove the `concat` altogether.

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
Blackbaud-SteveBrush: Just confirming: you verified this with Matt (that there's no valid version of repeater that would be a listbox)?

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
```diff
@@ -55,17 +55,16 @@ export class SkyRepeaterItemComponent
...
       ? 0
       : -1;
   }
-
```
Blackbaud-SteveBrush, line 68: Can we re-add the newline? It makes the code "breathe" better with the extra line.

libs/components/lists/src/lib/modules/repeater/repeater-item.component.ts
Blackbaud-SteveBrush: This operation can be cleaned up a bit. Put the `selectable` check in the same place as the `disabled` check, and instead of using `filter(...)[0]`, use `find()`. (I don't think `.find()` existed when this component was first written.)
```
public get tabindex(): 0 | -1 {
  return this.#repeaterService.items.find(
    (item) => !item.disabled && this.selectable,
  ) === this
    ? 0
    : -1;
}
```

libs/components/lists/src/lib/modules/repeater/repeater.component.ts
Blackbaud-SandhyaRajasabeson: the main array is only for things inside `repeater-item-title` and `repeater-item-content`, this accounts for the context menu which is rendered outside of those. I added a unit test that fails when this was `skyux-dropdown`

Blackbaud-SteveBrush: Ah, I see now. Thanks for clarifying.

Blackbaud-SteveBrush (approved):

PR #4062, opened 2025-10-30
fix(components/lists): add background color styles to list summary when scrolled

Blackbaud-CoreyArcher:
[AB#3604640](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3604640)

### DIFF
--- a/libs/components/lists/src/lib/modules/list-summary/list-summary.component.scss
+++ b/libs/components/lists/src/lib/modules/list-summary/list-summary.component.scss
@@ -24,8 +24,8 @@
       var(--sky-comp-list_summary-space-inset-left)
   );
   background-color: var(
-    --sky-comp-override-list-header-background-color,
-    var(--sky-override-list-summary-background-color)
+    --sky-override-list-summary-background-color,
+    var(--sky-comp-override-list-header-background-color)
   );
   border-bottom: var(--sky-override-list-summary-border-bottom, none);
 }
--- a/libs/components/lists/src/lib/modules/list-summary/list-summary.component.scss
+++ b/libs/components/lists/src/lib/modules/list-summary/list-summary.component.scss
@@ -24,10 +24,20 @@
       var(--sky-comp-list_summary-space-inset-left)
   );
   background-color: var(
-    --sky-override-list-summary-background-color,
-    var(--sky-comp-override-list-header-background-color)
+    --sky-comp-override-list-header-background-color,
+    var(--sky-override-list-summary-background-color)
   );
   border-bottom: var(--sky-override-list-summary-border-bottom, none);
+}
+
+:host-context(.sky-viewkeeper-fixed) {
+  background-color: var(
+    --sky-override-list-summary-background-color,
+    var(
+      --sky-comp-override-list-header-background-color,
+      var(--sky-background-color-page-default, var(--sky-color-background-page))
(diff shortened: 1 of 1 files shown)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4062/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4062/storybooks/ag-grid-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/4062/storybooks/data-manager-storybook/)
- [filter-bar-storybook](https://blackbaud.github.io/skyux-pr-preview/4062/storybooks/filter-bar-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/4062/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/4062/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/4062/storybooks/lookup-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/4062/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/4062/storybooks/phone-field-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/4062/storybooks/split-view-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4062/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4062/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4062/playground/)

libs/components/lists/src/lib/modules/list-summary/list-summary.component.scss
```diff
@@ -24,12 +24,22 @@
       var(--sky-comp-list_summary-space-inset-left)
   );
   background-color: var(
-    --sky-override-list-summary-background-color,
```
Blackbaud-TrevorBurch, line 27: Do we need this switch? If so - seems like the viewkeeper portion below should also be switched. But the override being first seems right to me. (Probably has no effect in practice either way since the override would be in default but the custom property is probably never set there - but it feels like the override should be first).

Blackbaud-CoreyArcher: I only switched it because it was in that order in Toolbar, and when that change alone didn't have any effect, I tried adding the viewkeeper host context selector but never changed it back. I'm ok with whichever way you feel is best.

Blackbaud-TrevorBurch: Lets change this one back and leave the viewkeeper. I think toolbar may technically be flipped - but I would leave it alone unless we hit a bug since the override for both of these are in default that shouldn't ever get the dimmed property set anyway.

Blackbaud-CoreyArcher: reverted

Blackbaud-TrevorBurch (approved):

PR #4063, opened 2025-10-31
feat(components/data-grid): create an easy mode for data grid

johnhwhite:
[AB#3605103](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3605103) [AB#3677437](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3677437) [AB#3677441](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3677441)

[Code examples](https://blackbaud.github.io/skyux-pr-preview/4063/code-examples-playground/#/?docsGroup=data-grid-package%20%7C%20@skyux%2Fdata-grid&docs-active-tab=examples)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4063/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4063/storybooks/ag-grid-storybook/)
- [data-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4063/storybooks/data-grid-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/4063/storybooks/data-manager-storybook/)
- [filter-bar-storybook](https://blackbaud.github.io/skyux-pr-preview/4063/storybooks/filter-bar-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/4063/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/4063/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/4063/storybooks/lookup-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/4063/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/4063/storybooks/phone-field-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/4063/storybooks/split-view-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4063/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4063/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4063/playground/)

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,727 @@
...
+   * @default 0
+   */
+  public readonly pageSize = input<number, unknown>(0, {
+    transform: (val: unknown) => coerceNumberProperty(val, 0),
+  });
+
+  /**
+   * The query parameter name to use for the current page number. When set, page changes are reflected in the URL.
+   */
+  public readonly pageQueryParam = input<string>();
```
Blackbaud-SteveBrush: Can this handle multiple grids on the same page? To keep things simple, I'd recommend removing the feature that responds to URL parameters.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,727 @@
...
+   * Fires when columns change. This includes changes to the displayed columns and changes
+   * to the order of columns. The event emits an array of IDs for the displayed columns that
+   * reflects the column order.
+   */
+  public readonly selectedColumnIdsChange = output<string[]>();
+
+  /**
+   * The columns to hide by default based on the ID or field of the item.
+   */
+  public readonly hiddenColumns = input<string[], unknown>([], {
```
Blackbaud-SteveBrush: How is this input different from the `selectedColumnIds` input? Can we have just the one?

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
```diff
@@ -0,0 +1,137 @@
...
+  public readonly width = input<number, unknown>(0, {
+    transform: numberAttribute,
+  });
+
+  /**
+   * The filter ID that maps this column to a filter bar item.
+   * When set, changes to the corresponding filter will apply to this column.
+   * Defaults to the column's `field` value.
+   */
+  public readonly filterId = input<string>();
```
Blackbaud-SteveBrush: Only one filter per column? What if the row is filtered against data that doesn't have a column?

Blackbaud-SteveBrush (commented):
I'm impressed with how much functionality you were able to pack into this component and how cleanly it can all be configured by the consumer through the template.

1.  **Integration with data manager:**  
    How well does `sky-data-grid` work with the full set of features offered by data manager? Are there any data manager capabilities that wouldn’t function as expected if a consumer uses `sky-data-grid` as one of the views? It would be helpful to include another code example showing how a consumer could wire up `sky-data-grid` inside a data manager instance.

2.  **Inline delete + multiselect example:**  
    I noticed you added support for inline delete confirmation. Could you add a code example demonstrating multiselect combined with inline delete and a context menu?

3.  **Clarifying the role of "simple grid":**  
    My initial understanding of "simple grid" was that it would be minimal in both API surface and functionality--essentially just a grid without additional features. However, this PR introduces inline delete, paging, filtering logic, and more. The features work well, and I'm not opposed to them, but I’m concerned that we may be drifting into a more specialized grid that fits only certain scenarios. This may also add complexity when considering how it interacts with components like data manager.

    I think it would help to map out how this component fits into our broader "grid ecosystem." A decision tree (or even just a bulleted breakdown) would help clarify which situations call for each grid option. This will also help us when we document the various grid features available to consumers. Something like:

    *   **Choose data grid when...**  
        You have all data client-side (no server‑side paging), you need simple filters tied directly to visible columns, you only need a single view, and you don't require a column picker.

    *   **Choose data manager with AG Grid when...**  
        You need server‑side paging, data entry, custom cell renderers, complex filters not tied to visible columns, multiple views, or a column picker.

    *   **Choose data manager with data grid when...**  
        (This is where I’m unsure.) What features are supported in this combination--filtering, data state, paging, search, etc.? A brief explanation would help clarify whether this hybrid scenario is viable and when a consumer should choose it.

Other features to think through:
- Sticky settings?
- Async data handling?

johnhwhite:
The when to use data-grid question is a good one, and I think we understood the way this complicates a crowded field of options. Here's the decision tree:

```mermaid


flowchart TD
    %% SKY UX Grid Ecosystem Decision Tree

    A([Start]) --> B{Need spreadsheet‑like<br/>editable data?}
    B -- Yes --> C[Data Entry Grid<br/><small>• Only option for spreadsheet-style editing<br/>• Requires AG Grid hard mode<br/>• Not preferred for record-management</small>]
    B -- No --> D{Need full AG Grid API<br/>like events, renderers, options?}

    D -- Yes --> E[AG Grid Wrapper / Hard Mode<br/><small>• Full AG surface exposed<br/>• High upgrade/maintenance cost<br/>• Risk of divergent implementations</small>]
    E --> F{Need Data Manager<br/>sticky settings, column chooser?}
    F -- Yes --> G[Hard Mode + Data Manager]
    F -- No --> H[Hard Mode Wrapper no DM]

    D -- No --> I{New build OR migrating from<br/>old Grid / List Builder Grid View?}
    I -- Yes --> J[Data Grid new<br/><small>• SKY UX API / no AG details<br/>• Integrates with Filter Bar<br/>• Smoother migration path</small>]
    I -- No --> J

    J --> K{Need Data Manager<br/>sticky settings, column chooser?}
    K -- Yes --> L[Data Grid new + Data Manager]
    K -- No --> M[Data Grid new]

    %% Server-side filtering consideration
    subgraph Considerations
      N{{Do you require server‑side<br/>search/filtering?}}
      N -- Yes --> O[[Avoid Data Manager search;<br/>implement server-side with<br/>Data Grid new]]
      N -- No --> P[[Client-side is fine →<br/>Data Grid new ± Data Manager]]
    end

    %% Cross-links to consideration from primary "new" paths
    J -. also check .-> N
    L -. also check .-> N
    M -. also check .-> N

```

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,692 @@
...
+   * `pageChange`, `sortChange`, filter change, and search (when using search such as with a SKY UX data manager).
+   * If this input is not set, the data grid will page, sort, filter, and apply SKY UX data manager search text to the
+   * `data` provided.
+   * @default `undefined`
+   */
+  public readonly totalRowCount = input<number | undefined>(undefined);
+  readonly #useInternalFilters = computed(() => {
+    const totalRowCount = this.totalRowCount();
+    return typeof totalRowCount === 'undefined';
+  });
```
johnhwhite, line 242: This input is the proposed solution for AB#3677441 provider options.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
```diff
@@ -0,0 +1,175 @@
...
+   * @default -1
+   */
+  public readonly flexWidth = input<number, unknown>(-1, {
+    transform: numberAttribute,
+  });
+
+  /**
+   * Text to display in the column header.
+   */
+  public readonly heading = input<string>();
```
Blackbaud-SteveBrush: Can we make this `headingText` to be consistent with our other components?

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
```diff
@@ -0,0 +1,175 @@
...
+})
+export class SkyDataGridColumnComponent {
+  /**
+   * The description for the column.
+   */
+  public readonly description = input<string>();
+
+  /**
+   * The property to retrieve cell information from an entry on the grid `data` array.
+   * You must provide either the `id` or `field` property for every column,
```
Blackbaud-SteveBrush: Should the description mention "columnId" instead of "id"? I see "id" used in a few other places, too.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
```diff
@@ -0,0 +1,175 @@
...
+   */
+  public readonly hidden = input<boolean, unknown>(false, {
+    transform: booleanAttribute,
+  });
+
+  /**
+   * Whether the column can be resized by dragging the column header border.
+   * @default true
+   */
+  public readonly isResizable = input<boolean, unknown>(true, {
```
Blackbaud-SteveBrush: What do you think about a simplified, `resizable` for the name, to match the other boolean inputs?

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
```diff
@@ -0,0 +1,175 @@
...
+    transform: booleanAttribute,
+  });
+
+  /**
+   * Whether the column sorts the grid when users click the column header.
+   * @default true
+   */
+  public readonly isSortable = input<boolean, unknown>(true, {
```
Blackbaud-SteveBrush: Same here: `sortable`?

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
```diff
@@ -0,0 +1,175 @@
...
+   * which contains the value passed to the column, and the `row` variable, which contains
+   * the entire row data.
+   */
+  public readonly template = input<TemplateRef<unknown>>();
+
+  /**
+   * The data type of the column used for filtering, sorting, and rendering when a template is not provided.
+   * @default 'text'
+   */
+  public readonly type = input<'text' | 'number' | 'date' | 'boolean'>('text');
```
Blackbaud-SteveBrush: Can we be more specific and name this `dataType` in case we need another input down the road that sets a "type"? This also avoids using the reserved "type" attribute.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-filter.ts
```diff
@@ -0,0 +1,281 @@
...
+      return rowValue > filterValue;
+    case 'greaterThanOrEqual':
+      return rowValue >= filterValue;
+    default:
+      logger.warn(
+        `Unsupported number or date filter operator: ${filterOperator}`,
+      );
+      return true;
+  }
+}
```
Blackbaud-SteveBrush, line 287: Thanks for putting the filter logic in a separate file; nice and clean!

libs/components/data-grid/src/index.ts
```diff
@@ -0,0 +1,11 @@
...
+export { SkyDataGridColumnComponent } from './lib/modules/data-grid/data-grid-column.component';
+export { SkyDataGridModule } from './lib/modules/data-grid/data-grid.module';
+export { SkyDataGridFilterOperator } from './lib/modules/types/data-grid-filter-operator';
+export { SkyDataGridFilterValue } from './lib/modules/types/data-grid-filter-value';
+export {
+  SkyDataGridNumberRangeFilterValue,
+  SkyDataGridNumberRangeFilterFormGroup,
+} from './lib/modules/types/data-grid-number-range-filter-value';
+export { SkyDataGridSort } from './lib/modules/types/data-grid-sort';
+export { SkyDataGridColumnInlineHelpComponent as λ1 } from './lib/modules/data-grid/data-grid-column-inline-help.component';
```
Blackbaud-SteveBrush: Since this component is only used internal to another component, I don't think you need to export it (the build should still work).

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.html
```diff
@@ -0,0 +1,49 @@
+<sky-wait [isWaiting]="!gridReady()" />
+@if (gridOptions(); as gridOptions) {
+  @if (useDataManager) {
+    <sky-data-view [viewId]="viewId()">
```
Blackbaud-SteveBrush: This looks like a nice convenience for our consumers! I have a few questions about how this would work with multiple views in a data manager:

**Questions:**
1. How would this interact if a consumer wraps the data grid in a `sky-data-view` component?
2. Have we considered an approach where consumers explicitly opt-in via a directive (similar to how we handle filter bar with `skyDataManagerFilterController`)? By automaticly wrapping the grid with `sky-data-view`, the data grid would be the only view that doesn't need an explicit `sky-data-view` wrapper, which could be confusing when developers add additional views. Creating a directive would make the integration more discoverable and consistent with other data manager views. You could follow the existing pattern set with the `skyDataManagerFilterController`.

**Suggested approach:**

```html
<sky-data-manager>
  <sky-data-view viewId="myView">
    <sky-data-grid skyDataManagerGridViewController>
    </sky-data-grid>
  </sky-data-view>
</sky-data-manager>
```

**How it would work:**

1. Create in data-manager: `skyDataManagerGridViewController` directive (similar to `skyDataManagerFilterController`)
2. Directive behavior: Injects parent `SkyDataViewComponent` to get `viewId`, then provides a bridge service that exposes `displayedColumnIds` and `searchText`
3. Data-grid consumes: Optionally injects the bridge service to receive state updates
4. Remove from data-grid: All direct data-manager dependencies to keep clean separation

**Benefits:**

1. Keeps dependencies isolated (neither data-manager nor data-grid depend on the other)
2. Directive is component-agnostic (could work with other grid components)
3. Consistent with existing filter bar pattern
4. `viewId` stays on `sky-data-view` where it belongs

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,784 @@
...
+  /**
+   * The set of IDs for the rows to prompt for delete confirmation.
+   * The IDs match the `multiselectRowId` properties of the `data` objects.
+   */
+  public readonly rowDeleteIds = model<string[]>([]);
+
+  /**
+   * The column IDs or fields for columns to show. Should not be combined with `hiddenColumns`.
+   */
+  public readonly selectedColumnIds = input<string[], unknown>([], {
```
Blackbaud-SteveBrush: The name of this input made a bit more sense in the list builder context because it has its own column picker--the value represented the column IDs the user selected from that UI. But in `data-grid`, there isn’t a picker, so the term "selected" can be misleading, especially alongside "selected" rows.

What do you think about renaming this to `displayedColumnIds` instead? You used the phrase "displayed columns" in the JSDoc for `selectedColumnIdsChange`, and I think that terminology fits well here too. It also pairs nicely with the idea of `displayedColumnIds` vs. `hiddenColumnIds`.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,784 @@
...
+
+  readonly #columnDefs = computed<ColDef<T>[]>(() => {
+    const columns = this.columns();
+    const sort = this.sort();
+    const displayed = [
+      ...this.selectedColumnIds().filter(Boolean),
+      ...this.#dataManagerSelectedColumnIds().filter(Boolean),
+    ];
+    const hidden = this.hiddenColumnIds().filter(Boolean);
+    return columns.map((col): ColDef => {
```
Blackbaud-SteveBrush: This `map` is throwing the ESLint rule `complexity`. Being a new component, it would be ideal if it didn't introduce new lint errors.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,784 @@
...
+   * The set of IDs for the rows to select in a multiselect grid.
+   * The IDs match the `multiselectRowId` properties of the `data` objects.
+   * Rows with IDs that are not included are de-selected in the grid.
+   */
+  public readonly selectedRowIds = model<string[]>([]);
+
+  /**
+   * The sort setting for the grid.
+   */
+  public readonly sort = model<SkyDataGridSort | undefined>(undefined);
```
Blackbaud-SteveBrush: Being a verb, I initially thought the `sort` input allowed the consumer to provide a sorting function, but that's not the case. What about a name represented by your JSDoc comment: `sortSettings`?

libs/components/code-examples/src/lib/modules/data-grid/filtered/example.component.html
```diff
@@ -0,0 +1,75 @@
+<sky-filter-bar [(appliedFilters)]="appliedFilters">
```
Blackbaud-SteveBrush, line 1: Does the filter bar need any bottom margin when placed above a data grid? Maybe Matt knows what's best to do, here.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,784 @@
...
+  });
+
+  /**
+   * The total number of records. When this input is set, it is expected that `data` will be updated for each
+   * `page` change, `sort` change, `appliedFilters` change, and search (when using search such as with a SKY UX data manager).
+   * If this input is not set, the data grid will page, sort, filter, and apply SKY UX data manager search text to the
+   * `data` provided, and the total row count is assumed to be `data.length`.
+   * @default undefined
+   */
+  public readonly totalRowCount = input<number | undefined>(undefined);
```
Blackbaud-SteveBrush: Currently, the `totalRowCount` input serves as a mode switch between client-side and server-side data handling. When set, it disables internal pagination, filtering, and search, requiring consumers to manage data, loading, and `totalRowCount` inputs plus coordinate updates via effects.

I suggest we remove the `data` input and implement a `loadAsync` output that follows the existing `searchAsync` pattern from the lookup component. This would provide a clearer, more consistent API for both client-side and server-side setups:

```
public readonly loadAsync = output<SkyDataGridLoadAsyncArgs<T>>();
```

Consumer HTML:

```
<sky-data-grid 
  pageSize="10"
  (loadAsync)="loadEmployees($event)"
>
  <!-- columns -->
</sky-data-grid>
```

Consumer TypeScript:

```
loadEmployees(args: SkyDataGridLoadAsyncArgs) {
  this.http.get<{ items: Employee[], totalCount: number }>(
    '/api/employees', 
    { params: args.params }
  ).subscribe(result => {
    args.callback(result);
  });
}
```

Benefits:

- Consistent with existing SKY UX patterns (lookup's `searchAsync`)
- Grid manages `loading` and `totalCount` internally
- Single event-driven API instead of coordinating multiple inputs
- Grid automatically calls the event when page/sort/filters change

How it works:

1. Grid fires `loadAsync` when page/sort/filters change
2. Consumer makes API call
3. Consumer calls `args.callback(result)` with data
4. Grid updates internal state and displays results
5. This pattern helps developers quickly implement async data loading without needing to understand the intricacies of mode switching, making it easier to build correctly the first time.

libs/components/data-grid/documentation.json
```diff
@@ -0,0 +1,31 @@
...
+          "SkyDataGridFilterValue",
+          "SkyDataGridNumberRangeFilterValue",
+          "SkyDataGridSort"
+        ],
+        "primaryDocsId": "SkyDataGridModule"
+      },
+      "testing": {
+        "docsIds": ["SkyDataGridHarness", "SkyDataGridHarnessFilters"]
+      },
+      "codeExamples": {
```
Blackbaud-SteveBrush, line 23: I think we could tighten up the code examples just a bit, and add a few more to spotlight specific features:

- [x] Basic: includes static data and column types (this appears to be done)
- [ ] Pagination: page, pageSize, and pageQueryParam (possible to include this param?)
- [ ] Filters: filter bar, check with Matt about design guidelines
- [ ] Multiselect: context menu, handle selection changes
- [ ] Async data (there is an example named "Custom filters and delayed loading data grid" but I think it would benefit consumers to have the async feature isolated.
- [ ] Data manager: includes async data and pagination, filter bar, search, column picker, and sort
- [ ] Inline row delete

Blackbaud-SteveBrush (commented):
Overall looks good and works well. Summary of the feedback:
 
- API naming suggestions
- Create a directive to simplify data manager view integration and to decouple dependencies
- Alternative approach for the async data API

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
johnhwhite: Good call. Updated.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
johnhwhite: Yes. I'd meant to update that.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
johnhwhite: Good call. Updated.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
johnhwhite: Good call. Updated.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
johnhwhite: Good call. Updated to `dataType`.

libs/components/data-grid/src/index.ts
johnhwhite: Good call. Updated.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
johnhwhite: Broke that up to make things less complex.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
johnhwhite: Good call. I went with grid's `sortField`; let me know if you think that works.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.html
johnhwhite: I had thought about this, but data grid relies on ag-grid project, which already has a dependency on data manager. If it couldn't be avoided, it might as well be simplified, and that's how I got to this implementation.

It's possible to nest `<sky-data-grid>` within a `<sky-data-view>` and not pass `viewId`, but the consumer would be responsible for all of the connections. Since `<sky-data-view>` alone does not provide much, it seemed ok to bury it.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
johnhwhite: I like `displayedColumnIds`. Updated.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
johnhwhite: The `[data]` input comes from the grid component and I think we'd want to leave it as an option regardless for migrations. I removed the `[loading]` input and show a loading indicator whenever `[data]` is nullish, as happens with AsyncPipe or ResourceRef when waiting on an asynchronous value.

I'm not convinced `loadAsync` is necessary. The `(searchAsync)` pattern responds to keystrokes in a form context where signals were not available, whereas data grid deals with largely static data sets and can take advantage of the signal graph to handle updates.

I've updated the async example to use data manager for all state updates (previously `sort` came from data grid for no particular reason).

We can talk it through tomorrow.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
```diff
@@ -0,0 +1,177 @@
...
+   * The data type of the column used for filtering, sorting, and rendering when a template is not provided.
+   * @default 'text'
+   */
+  public readonly dataType = input<'text' | 'number' | 'date' | 'boolean'>(
+    'text',
+  );
+
+  /**
+   * The width of the column in pixels. When set to `0`, the column width is evenly distributed.
+   * @default 0
```
Blackbaud-ErikaMcVey: any reason we chose 0 over undefined?

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
```diff
@@ -0,0 +1,177 @@
...
+  selector: 'sky-data-grid-column',
+  template: '',
+})
+export class SkyDataGridColumnComponent {
+  /**
+   * The unique ID for the column. You must provide either the `columnId` or `field` property
+   * for every column, but do not provide both. Use `columnId` when the column does not map directly to a field
+   * in the data set.
+   */
+  public readonly columnId = input<string>();
```
Blackbaud-ErikaMcVey, line 30: do we do anything to actually prevent them setting both? throw an error?

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
```diff
@@ -0,0 +1,177 @@
...
+   * but do not provide both. If `columnId` does not exist on a column, then `field` is the entry
+   * for the grid `selectedColumnIds` array.
+   */
+  public readonly field = input<string>();
+
+  /**
+   * When set to a value greater than `-1`, it overrides `width` and works like
+   * [CSS flex-grow](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/flex-grow), where a column
+   * with `flexWidth="2"` is twice the width of a column with `flexWidth="1"`, and `flexWidth="0"` does not auto-expand.
+   * @default -1
```
Blackbaud-ErikaMcVey: same as width--why not just undefined?

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
```diff
@@ -0,0 +1,177 @@
...
+   */
+  public readonly flexWidth = input<number, unknown>(-1, {
+    transform: numberAttribute,
+  });
+
+  /**
+   * Text to display in the column header.
+   */
+  public readonly headingText = input<string>();
```
Blackbaud-ErikaMcVey: should we make this required for a11y purposes?

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,828 @@
...
+   * are de-selected in the grid.
+   */
+  public readonly rowHighlightedId = input<string>();
+
+  /**
+   * Whether the data grid is stacked on another data grid. When specified, the appropriate
+   * vertical spacing is automatically added to the data grid.
+   * @default false
+   */
+  public readonly stacked = input<boolean, unknown>(false, {
```
Blackbaud-ErikaMcVey, line 232: is this _only_ for if it is stacked on another data grid, or is it more general like our other stacked inputs?

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,828 @@
...
+   * View ID when using SKY UX Data Manager. When this input is set, `sky-data-grid` provides a `sky-data-view` for a
+   * SKY UX Data Manager. Requires a
+   * [`SkyDataManagerService`](https://developer.blackbaud.com/skyux/components/data-manager?docs-active-tab=development#class_sky-data-manager-service)
+   * be provided and configured.
+   */
+  public readonly viewId = input<string>();
+
+  /**
+   * The width of the grid in CSS pixels. When set to `0`, the grid will use the width of its container.
+   * @default 0
```
Blackbaud-ErikaMcVey: similar to my other width comment--why not just be unset by default?

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,828 @@
...
+   * The unique ID that matches a property on the `data` object.
+   * @default 'id'
+   */
+  public readonly multiselectRowId = input<keyof T>('id');
+
+  /**
+   * The number of items to display per page. Set to `0` to disable pagination.
+   * @default 0
+   */
+  public readonly pageSize = input<number, unknown>(0, {
```
Blackbaud-ErikaMcVey, line 209: also odd to me, why not just don't set pageSize if you don't want pagination?

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,828 @@
...
+   * The IDs match the `multiselectRowId` properties of the `data` objects.
+   */
+  public readonly rowDeleteIds = model<string[]>([]);
+
+  /**
+   * The set of IDs for the rows to select in a multiselect grid.
+   * The IDs match the `multiselectRowId` properties of the `data` objects.
+   * Rows with IDs that are not included are de-selected in the grid.
+   */
+  public readonly selectedRowIds = model<string[]>([]);
```
Blackbaud-ErikaMcVey, line 281: I noticed if you filter out selected rows and then clear the filters they're still selected; this appears to differ from the behavior of the data manager demo, but I'm not sure if that is specific to that demo or even a specific choice.

Also, the OG data manager large playground examples appears to have broken

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
johnhwhite: Fair question. It's using `transform` to handle string inputs, and I'd prefer to keep the type as `number` rather than `number | undefined`, so it's using `0` as the default. I'm removing the `@default` annotation and updating the docs to not mention `0`.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
johnhwhite: I've added a log warning. It wouldn't break anything but it would cause ambiguity.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
johnhwhite: Same reason. I'm removing the `@default` annotation and updating the docs to not mention `-1`.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
johnhwhite: Good call. Added.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
johnhwhite: I've updated the docs. It's just adding the class, so it could be anything.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
johnhwhite: Same. I'm removing the `@default` annotation and updating the docs to not mention `0`.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
johnhwhite: Same. I'm removing the `@default` annotation and updating the docs to not mention `0`.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
johnhwhite: I've updated the docs to specify that it only works with one grid on the page.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
johnhwhite: Good catch. That should be fixed now and I've added tests around it.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
Blackbaud-SteveBrush: I appreciate the added context, but I still have concerns about the API design that should be addressed in this PR.

Data grids are commonly used with server-side pagination, filtering, and sorting for large datasets--we could argue that this is a primary use case, not an edge case (the `sky-list` component has nearly 800 usages of the `dataProvider` and `data` inputs; the `grid` component has 93 usages of the static `data` input). While signals help with reactive updates, they don't eliminate the need for a clear async loading pattern. 

Having `[data]` and `[totalRowCount]` inputs creates a confusing dual-mode API where their presence/absence switches behavior. This requires consumers to understand mode switching logic rather than having a single, clear contract. The `loadAsync` pattern provides a simple API:

- **All** inputs work regardless of the type of data -- Consumer doesn't need to read the docs to know which inputs/outputs work for static vs async data
- Grid fires event with current state (page, sort, filters) -- One-stop location to respond to data state without juggling multiple signals
- Consumer fetches data and calls callback -- Consumer doesn't need to worry about subscription or loading/error-state management
- Same pattern whether you have 10 rows or 10 million -- Documentation doesn't have branching logic about what to do in specific scenarios; same API for all use cases

You're right that signals weren't available when `searchAsync` was created. However, the pattern works because it provides a clear contract for imperative actions: "component needs data, consumer provides it." That same principle applies here. While signals are great for reactive data flow, events remain the right tool for imperative actions like "load data now with these parameters."

Additional concern: Pagination is difficult to set up when `totalRowCount` is set

I tried asking Copilot how it would setup pagination when `totalRowCount` is set; its conclusion was that it's not possible. When `totalRowCount` is provided, the grid disables internal pagination and only displays the first `pageSize` rows via `data.slice(0, pageSize)`. The pagination UI appears, but clicking pages does nothing because the page number is ignored when slicing data.

To make pagination work, consumers must:

- Watch for changes to `page()`, `sortField()`, and `appliedFilters()`
- Manually fetch new data for the current page
- Update the data input with only the rows for that page

This pattern isn't documented and requires consumers to coordinate multiple inputs/outputs through effects--exactly the complexity that `loadAsync` would eliminate. The async example doesn't show pagination, so maybe I'm missing a key piece.

Let's discuss during flex. The current implementation requires consumers to coordinate multiple inputs and understand mode switching, but I think we can do better with a single, event-driven pattern that's easy for our consumers to reason about and is consistent with existing SKY UX patterns.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.html
Blackbaud-SteveBrush: My primary concern isn’t the data manager dependency itself (though that still matters), but the expectations consumers will have when using the API. New components should follow established patterns; in this case, the data manager encourages wrapping views in a `sky-data-view`, so anyone familiar with it would naturally place the data grid inside one. Consumers could integrate the data grid with the data manager manually if they wanted, but the directive we provide makes that integration far simpler.

Maybe try asking Copilot to create a directive in the same vein as the filter bar directive and see what it produces. If it adds more complexity than it's worth, we can reassess.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
Blackbaud-SteveBrush: Alternatively, if the event‑driven pattern doesn’t feel like the right fit, I could be convinced to keep the `data` input, provided we also introduce a `dataLoader` input to automatically handle async data for consumers. This way, the "mode switching" would be self-documenting and more intuitive to consumers. It would also let the data grid manage transforms internally, so all inputs behave consistently whether the data is static or async. 

I put together a simple demo to illustrate the idea:
https://stackblitz.com/edit/stackblitz-starters-kppeqtgw?file=src%2Fmain.ts.

Consumers wouldn’t need to handle any data transformations, whether the data is static or async, which helps meet the goal of keeping the grid simple for the most common use cases. 

Does this feel like a reasonable balance while still minimizing consumer boilerplate?

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.html
Blackbaud-SteveBrush: To summarize: my primary concern isn’t the data manager dependency itself (though that still matters), but the expectations consumers will have when using the API. New components should follow established patterns; in this case, the data manager encourages wrapping views in a `sky-data-view`, so anyone familiar with it would naturally place the data grid inside one. Consumers could integrate the data grid with the data manager manually if they wanted, but the directive we provide makes that integration far simpler.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
johnhwhite: The `totalRowCount` docs say "When this input is set, it is expected that `data` will be updated for each _`page` change_, `sort` change, `appliedFilters` change, and search (when using search such as with a SKY UX data manager)." Page is the only thing that does not currently run through data manager, and perhaps it should be.

I like the idea of offering a `dataLoader` or `loadAsync` in addition to `data`, and emit an event whenever criteria change about what data to show as an alternative to listening to a data manager state change.

libs/components/code-examples/src/lib/modules/data-grid/asynchronous-data/example.component.ts
```diff
@@ -0,0 +1,155 @@
...
+  protected readonly hideInactiveModal = HideInactiveFilterModalComponent;
+  protected readonly startDateFilterModal = StartDateFilterModalComponent;
+  protected readonly pageRequest = signal<SkyDataGridPageRequest | undefined>(
+    undefined,
+  );
+
+  protected readonly viewId = 'dataGridWithCustomFilters' as const;
+
+  // Computed client side in this example, but could be an HTTP resource where parameters are sent to the server for determining data to show.
+  protected readonly recordsToShow = resource({
```
Blackbaud-SteveBrush: The `resource` API is marked as experimental (even in Angular 21). Let's instead demonstrate standard `HttpClient` requests (mocked) since most of our consumers are familiar with it.

libs/components/code-examples/src/lib/modules/data-grid/data-manager/example.component.ts
```diff
@@ -0,0 +1,130 @@
...
+ * @title Data grid with data manager
+ */
+@Component({
+  selector: 'app-data-grid-data-manager-example',
+  templateUrl: './example.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  providers: [SkyDataManagerService],
+  imports: [
+    I18nPluralPipe,
+    SkyAgGridModule,
```
Blackbaud-SteveBrush: From the consumer's perspective, they wouldn't necessarily know that data grid uses AG Grid, so asking them to import it would seem strange. Can the `SkyDataGridModule` export `SkyAgGridModule`?

libs/components/data-manager/src/lib/modules/data-manager/data-manager-host/data-manager-host-controller.directive.ts
```diff
@@ -0,0 +1,237 @@
...
+import { SkyDataManagerFilterControllerDirective } from '../data-manager-filters/data-manager-filter-controller.directive';
+import { SkyDataManagerService } from '../data-manager.service';
+import { SkyDataManagerSortOption } from '../models/data-manager-sort-option';
+import { SkyDataManagerState } from '../models/data-manager-state';
+import { SkyDataViewState } from '../models/data-view-state';
+
+import { SkyDataManagerHostService } from './data-manager-host.service';
+
+/**
+ * A directive applied to a data-displaying component (like `sky-data-grid`) that enables integration with a data manager.
```
Blackbaud-SteveBrush, line 24: This description seems to indicate that the directive should be added to the `sky-data-grid`, but from the code examples, it's only being added to the `sky-data-view` components (I'm assuming so it can overload the `viewId` input).

Is that true? If so, we should update the JSDoc comment to reflect that it should only be used on the `sky-data-view` component. Additionally, we could update the selector: `sky-data-view[skyDataManagerHostController]`.

libs/components/data-manager/src/lib/modules/data-manager/data-manager-host/data-manager-host-controller.directive.ts
```diff
@@ -0,0 +1,237 @@
...
+@Directive({
+  selector: '[skyDataManagerHostController]',
+  providers: [
+    SkyDataManagerHostService,
+    {
+      provide: SkyDataHostService,
+      useExisting: SkyDataManagerHostService,
+    },
+  ],
+  hostDirectives: [SkyDataManagerFilterControllerDirective],
```
Blackbaud-SteveBrush, line 37: This seems strange to have here, since the docs say to add it to a "filter toolbar". Why is it needed here?

libs/components/data-manager/src/lib/modules/data-manager/data-manager-host/data-manager-host-controller.directive.ts
```diff
@@ -0,0 +1,237 @@
...
+        !!newDataManagerState.activeSortOption?.descending ||
+      !!oldDataViewState?.additionalData !==
+        !!newDataViewState?.additionalData ||
+      Number(oldDataViewState?.additionalData?.page) !==
+        Number(newDataViewState?.additionalData?.page) ||
+      String(oldDataManagerState.searchText) !==
+        String(newDataManagerState.searchText) ||
+      Number(oldDataManagerState.selectedIds?.length ?? 0) !==
+        Number(newDataManagerState.selectedIds?.length ?? 0) ||
+      !!oldDataManagerState.selectedIds?.some(
```
Blackbaud-SteveBrush, line 226: I fear that the complexity in this condition will be difficult to maintain long-term and will be prone to bugs. Can this logic be simplified if we add the AG Grid integration logic directly in the data grid component (instead of using the `skyAgGridDataManagerAdapter`)?

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,901 @@
...
+
+  /**
+   * The sort setting for the grid.
+   */
+  public readonly sortField = model<SkyDataGridSort | undefined>(undefined);
+
+  /**
+   * Fires when sorting or page number changes.
+   */
+  public readonly pageRequest = output<SkyDataGridPageRequest>();
```
Blackbaud-SteveBrush, line 291: We talked about this in detail, but I'm going to insist we try implementing a `loadAsync` output that handles the async "hard mode" switching. It would replace the `externalRowCount` input and the `pageRequest` and `rowCountChange` outputs, creating a singular entry point that communicates the mode's intent:

**Static mode (automatic features):**

- Consumer sets data input
- Component automatically handles filtering, pagination, sorting
- Simple scenarios, small datasets

**Async mode (consumer-managed):**

- Consumer provides loadAsync event handler
- Component emits load requests: `{ page, pageSize, sort, filters, result }`
- Consumer sets `args.result` to their Observable
- Component manages loading/error states internally

This makes the mode explicit (consumer provides handler or doesn't) and eliminates the confusing `externalRowCount` mode switch.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,901 @@
...
+  public readonly data = input<T[] | null | undefined>();
+
+  /**
+   * The number of records when using a remote data source.
+   * When `externalRowCount` is set, it is expected that `data` will be updated whenever `pageRequest` emits a new value.
+   * When both `externalRowCount` and `pageSize` are set, the number of pages is assumed to be `Math.ceil(externalRowCount / pageSize)`.
+   * If `externalRowCount` is not set, the data grid will page, sort, filter, and apply SKY UX data manager search text to the
+   * `data` provided, and the row count is assumed to be `data.length`.
+   */
+  public readonly externalRowCount = input<number | undefined>(undefined);
```
Blackbaud-SteveBrush: Still hung up on this being a "mode switch". Can we go with `loadAsync` instead?

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,901 @@
...
+
+  /**
+   * Emits a row count after filters are updated. Not used when `externalRowCount` is set.
+   */
+  public readonly rowCountChange = output<number>();
+
+  /**
+   * Fires when users cancel the deletion of a row.
+   */
+  public readonly rowDeleteCancel = output<SkyAgGridRowDeleteCancelArgs>();
```
Blackbaud-SteveBrush: The `SkyAgGridRowDeleteCancelArgs` and `SkyAgGridRowDeleteConfirmArgs` types are exposed here (and in the code example); we probably want data grid versions of those types so that AG Grid is kept as an internal detail. 

```
export type SkyDataGridRowDeleteCancelArgs = SkyAgGridRowDeleteCancelArgs;
```

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.html
```diff
@@ -0,0 +1,49 @@
+<sky-wait [isWaiting]="!gridReady()" />
+@if (gridOptions(); as gridOptions) {
+  @if (useDataManager) {
+    <sky-ag-grid-wrapper
+      skyAgGridRowDelete
+      skyAgGridDataManagerAdapter
```
Blackbaud-SteveBrush, line 6: During our Tooling Sync, we agreed that `@skyux/data-manager` should be removed as a peer dependency to ensure the data grid isn't tightly coupled to any specific data-management solution. The only reason the host directive exists is to accomplish that goal.

What would it take to remove the peer dependency and the branching `useDataManager` logic in the component? The host directive would handle the data manager integration, and we'd copy any necessary code from the `skyAgGridDataManagerAdapter` for AG Grid integrations (keeping `skyAgGridDataManagerAdapter` as-is for consumers that want to interact with AG Grid directly).

Does that approach make sense, or are there complications I'm not seeing?

Blackbaud-SteveBrush (commented):
I resolved the other conversations so we can continue with fresh context. We're closer, but I think we're still looping on the same points. To avoid more back and forth, here's what I'd like to see in the next iteration:

- Implement `loadAsync` as the mode switch for async "hard mode"
- Remove `@skyux/data-manager` as a peer dependency
- Replace the experimental `resource` implementation in the code examples

If you see any blockers or have concerns about these changes that I haven't considered, let's talk through them before the next review.

libs/components/data-manager/src/lib/modules/data-manager/data-manager-host/data-manager-host-controller.directive.ts
johnhwhite: That was added because data host wasn't dealing with the filters field in data manager. Removing it.

Blackbaud-SteveBrush:
Just as a point of record: we talked offline and agreed we need to pivot on this and create two separate components: one to handle static data w/ automatic sorting/etc. and one to handle asynchronous data loading. This will help to keep the APIs and implementations simpler for our consumers.

PR #4133, opened 2025-12-19
feat(components/lookup): support searching when minimum characters is set to 0

Blackbaud-SandhyaRajasabeson:
[AB#3605881](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3605881)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4133/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4133/storybooks/ag-grid-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/4133/storybooks/data-manager-storybook/)
- [filter-bar-storybook](https://blackbaud.github.io/skyux-pr-preview/4133/storybooks/filter-bar-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/4133/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/4133/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/4133/storybooks/lookup-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/4133/storybooks/phone-field-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4133/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4133/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4133/playground/)

apps/playground/src/app/components/lookup/lookup/lookup.component.html
```diff
@@ -44,6 +44,25 @@
         [debounceTime]="1000"
         [disabled]="disabled"
         [enableShowMore]="true"
+        [searchTextMinimumCharacters]="0"
```
Blackbaud-TrevorBurch: Seems odd that we have an input box with a label calling out setting it to 0 right below one that is also doing that, no?

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -226,12 +227,16 @@ export class SkyAutocompleteComponent implements OnDestroy, AfterViewInit {
   /**
    * The minimum number of characters that users must enter before
    * the autocomplete component searches the data source and displays search
-   * results in the dropdown list.
+   * results in the dropdown list. Can be set to `0` to search on focus.
```
Blackbaud-TrevorBurch: Should we say 'Can be set to `0` to trigger `searchAsync` on focus or return all items if using the `data` input.'

That feels a little more direct on what happens here.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.spec.ts
```diff
@@ -617,6 +617,106 @@ describe('Autocomplete component', () => {
...
+      fixture.detectChanges();
+
+      enterSearch('', fixture);
+
+      const actionsContainer = getActionsContainer();
+      expect(actionsContainer).not.toBeNull();
+      expect(actionsContainer.textContent).toContain('No matches found');
+    }));
+
+    // TODO: confirm this path with design
```
Blackbaud-TrevorBurch: ????

Blackbaud-TrevorBurch: This seems right but I would expect it to be `Show matches` with the number if it isn't no results. This test is explicitly triggering that no results state.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.spec.ts
```diff
@@ -617,6 +617,106 @@ describe('Autocomplete component', () => {
       expect(spy).not.toHaveBeenCalled();
     }));
 
+    it('should search with empty text when searchTextMinimumCharacters is 0', fakeAsync(() => {
```
Blackbaud-TrevorBurch, line 620: I see tests here for the `data` input but not for `searchAsync`. We need to make sure we are testing that path as well.

apps/playground/src/app/components/lookup/lookup/lookup.component.html
Blackbaud-SandhyaRajasabeson: WOOPS i meant to make a whole new one so it wasnt stomping on the testing this one was doing lol

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.spec.ts
Blackbaud-SandhyaRajasabeson: From conversation with design: confirmed that we want the `Show all number` version of the text.

libs/components/lookup/src/lib/modules/autocomplete/autocomplete.component.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable complexity */
```
Blackbaud-TrevorBurch: If this is needed - can we scope this to where it is needed and then do an enable comment after that area? That way we aren't just disabling it for all of autocomplete.

Blackbaud-TrevorBurch: Or if there is something we can extract to not need it at all that would be even better - but I know that isn't always practical

Blackbaud-TrevorBurch (approved):

PR #4243, opened 2026-02-26
feat(sdk/skyux-eslint)!: add prefer-input-box rule

johnhwhite:
[AB#3673551](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3673551)

## Summary by CodeRabbit

* **New Features**
  * Added five template lint rules to enforce valid input types, require SKY UX input wrappers, restrict invalid children inside input wrappers, and recommend SKY UX form-control components; enabled in recommended and comprehensive configs.

* **Documentation**
  * Added guidance, examples, exemptions, and rationale for each new rule.

* **Tests**
  * Unit tests covering valid and invalid template scenarios for all new rules.

* **Chores**
  * Added shared utilities to support the new rules.


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4243/storybook/)

Component Storybooks:

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4243/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4243/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4243/playground/)

libs/sdk/skyux-eslint/src/configs/template-recommended.ts
```diff
@@ -13,6 +13,7 @@ export default [
       'skyux-eslint-template/no-radio-group-with-nested-list': 'error',
       'skyux-eslint-template/no-unbound-id': 'error',
       'skyux-eslint-template/prefer-disabled-attr': 'error',
+      'skyux-eslint-template/prefer-input-box': 'error',
```
Blackbaud-SteveBrush, line 19: Adding a new rule to a preset is considered a breaking change. Let's update the PR title and description to communicate that.

libs/sdk/skyux-eslint/src/rules/template/prefer-input-box.ts
```diff
@@ -0,0 +1,98 @@
...
+} from '@angular-eslint/utils';
+
+import { createESLintTemplateRule } from '../utils/create-eslint-template-rule';
+
+export const RULE_NAME = 'prefer-input-box';
+export const messageId = 'preferInputBox';
+
+const TARGET_ELEMENTS = new Set(['input', 'select', 'textarea']);
+
+const EXEMPT_INPUT_TYPES = new Set([
```
Blackbaud-SteveBrush: What do you think about switching to an allow list for input types here, similar to the allow list of wrappable components below? It seems to align more closely with the rule’s intent: only certain native controls/components should trigger wrapper enforcement. An exempt list can keep growing as edge cases come up, and it’s easier for something to slip through. With an allow list, we get a safer default and new or unknown input types won’t be treated as valid targets unless we explicitly include them.

Blackbaud-SteveBrush (commented):
Looks like this PR is missing a `libs/sdk/skyux-eslint/docs/rules/template/prefer-input-box.md` file to serve as the rule's documentation.

libs/sdk/skyux-eslint/src/rules/template/prefer-input-box.ts
johnhwhite: Going by [the spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input), that list would be checkbox, color, date, datetime-local, email, file, month, number, password, radio, range, search, tel, text, time, url, and week — it's a long, strange list.

The expansion has been into text-like inputs, it doesn't change often, and I feel more comfortable listing the exceptions.

Blackbaud-SteveBrush: Shared this with the team offline, and they aligned on narrowing `prefer-input-box` to the following scope:

- Use an allow list of input types that should be wrapped in an input box.
- Report when an input with an allow-listed type is not wrapped in `sky-input-box` (or an approved form control component such as `sky-datepicker`).
- Report when an input is wrapped in `sky-input-box` but its type is not in the allow list.

We also branched into a separate question: how should we handle other input types?

- We should create a separate rule for types that should use SKY form components instead of native inputs. Example: `input[type="date"]` should recommend `sky-datepicker`.
- We also discussed a potential rule for invalid/empty type values (or confirming whether `angular-eslint` already covers this).

The second set should be tracked as a new story. For this PR, the team was unanimous on the `prefer-input-box` scope above. The remaining open point is finalizing which input types belong in the allow list, likely in partnership with Design so docs and lint guidance stay aligned.

libs/sdk/skyux-eslint/src/rules/template/no-invalid-input-types.ts
```diff
@@ -0,0 +1,85 @@
...
+      ['Element'](element: TmplAstElement): void {
+        if (element.name.toLowerCase() !== 'input') {
+          return;
+        }
+
+        const typeAttr = element.attributes.find(
+          (attr) => attr.name === 'type',
+        );
+
+        // No type attribute defaults to "text", which is valid.
```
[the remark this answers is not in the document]
johnhwhite: @coderabbitai We encourage inline comments and JSDoc comments in our codebase. 
Comments improve code readability and help team members understand complex logic. 
Please do not flag comments as issues.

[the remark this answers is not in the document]
johnhwhite: @coderabbitai Thanks for calling that out. Yes, remove the learning "Do not add any comments in the code" and I will work with the team to clarify those instructions files.

libs/sdk/skyux-eslint/src/rules/template/prefer-input-box.ts
```diff
@@ -0,0 +1,88 @@
...
+
+const INPUT_BOX_INPUT_TYPES = new Set([
+  'email',
+  'month',
+  'number',
+  'password',
+  'range',
+  'text',
+  'url',
+  'week',
```
Blackbaud-SteveBrush: The `month`, `week`, and `range` types should probably be added to the `no-invalid-input-box-children` rule instead because they include a native user interface which doesn't play well with existing input box styles. (You can put all three of these types in an input box to see that we don't have styles in place to support them.)

The datepicker doesn't support a month-only or week-only selection so we can't recommend they use datepicker for `month` or `week` types, either.

libs/sdk/skyux-eslint/src/rules/template/no-invalid-input-box-children.ts
```diff
@@ -0,0 +1,207 @@
...
+  'month',
+  'number',
+  'password',
+  'range',
+  'text',
+  'url',
+  'week',
+]);
+
+const VALID_SKY_COMPONENTS = new Set([
```
Blackbaud-SteveBrush: This constant and the `VALID_INPUT_BOX_INPUT_TYPES` are referenced in multiple files. Can we export them from a shared file to ensure future changes affect the related rules?

libs/sdk/skyux-eslint/docs/rules/template/prefer-form-control-component.md
```diff
@@ -0,0 +1,182 @@
...
+### color
+
+Use `<sky-colorpicker>` instead of `<input type="color">`.
+
+```html
+<!-- Before -->
+<input type="color" />
+
+<!-- After -->
+<sky-input-box>
```
Blackbaud-SteveBrush: The colorpicker shouldn't be wrapped in an input box, correct?

libs/sdk/skyux-eslint/src/rules/template/prefer-form-control-component.ts
```diff
@@ -0,0 +1,86 @@
...
+      preferSkyRadio:
+        'Use <sky-radio-group> and <sky-radio> instead of <input type="radio">.',
+      preferSkySearch: 'Use <sky-search> instead of <input type="search">.',
+      preferSkyPhoneField:
+        'Use <sky-phone-field> instead of <input type="tel">.',
+      preferSkyTimepicker:
+        'Use <sky-timepicker> instead of <input type="time">.',
+    },
+    schema: [],
+    type: 'suggestion',
```
Blackbaud-SteveBrush: The other "prefer" rules use `"problem"`. Can we align the type for all three?

libs/sdk/skyux-eslint/src/rules/template/prefer-input-box.ts
```diff
@@ -0,0 +1,88 @@
...
+      ['Element:exit'](): void {
+        ancestorStack.pop();
+      },
+    };
+  },
+  defaultOptions: [],
+  meta: {
+    docs: {
+      description:
+        'Require text-like form control elements to be placed inside a <sky-input-box> component.',
```
Blackbaud-SteveBrush: Instead of "text-like", can we come up with a descriptor that includes native `select` elements? Even if the description is more verbose: "Require `select`, `textarea`, and text-entry `input` elements to be placed inside..."

johnhwhite: Updated.

libs/sdk/skyux-eslint/src/rules/template/prefer-form-control-component.ts
johnhwhite: Good catch. Updated.

libs/sdk/skyux-eslint/docs/rules/template/prefer-form-control-component.md
johnhwhite: Right, this shouldn't have been there.

libs/sdk/skyux-eslint/src/rules/template/no-invalid-input-box-children.ts
johnhwhite: Consolidated.

libs/sdk/skyux-eslint/src/rules/template/prefer-input-box.ts
johnhwhite: Fair. Dropped those input types as valid input box children.

Blackbaud-SteveBrush (approved):
Mind adding a `BREAKING CHANGE` message to the top of the PR description so we can easily copy/paste it into the final changelog entry for v14?

PR #4245, opened 2026-02-26
fix(sdk/skyux-eslint)!: modify `skyux-eslint-template/prefer-label-text` ESLint rule to require `labelText` for all `sky-input-box` implementations

Blackbaud-SteveBrush:
BREAKING CHANGE
The `skyux-eslint-template/prefer-label-text` ESLint rule now always reports an error for `<sky-input-box>` components that use a `<label>` child element instead of the `labelText` input. Previously, `<sky-input-box>` instances that used the `skyId` directive on a child input element were exempt from this rule, as this pattern was used to implement "hard mode" label association. This exemption has been removed. All `<sky-input-box>` components must now use the `labelText` input for labeling. Run `ng lint --fix` to automatically migrate eligible usages.

[AB#3673971](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3673971)

## Summary by CodeRabbit

* **Documentation**
  * Expanded labeling guidance with detailed before/after examples, affected components, modern label patterns, autofix notes, and updated usage snippets.

* **Improvements**
  * Rule now covers additional input/label scenarios previously bypassed, including cases with internal identifiers.

* **Tests**
  * Broadened test coverage to include autofix conversions and explicit non-fix cases (e.g., bound labels).

### DIFF
--- a/libs/sdk/skyux-eslint/docs/rules/template/prefer-label-text.md
+++ b/libs/sdk/skyux-eslint/docs/rules/template/prefer-label-text.md
@@ -2,6 +2,8 @@
 
 Ensures form components set the `labelText` (or `headingText`) attribute, which automatically activates key usability and accessibility features.
 
+Setting `labelText` (or `headingText`) on a component is the preferred, modern approach to labeling SKY UX form components. It replaces older patterns that placed a dedicated label child element (e.g. `<sky-checkbox-label>`, `<sky-input-box><label>`) inside the component.
+
 - Type: problem
 - 🔧 Supports autofix (`--fix`)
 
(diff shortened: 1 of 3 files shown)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4245/storybook/)

Component Storybooks:

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4245/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4245/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4245/playground/)

libs/sdk/skyux-eslint/docs/rules/template/prefer-label-text.md
```diff
@@ -13,6 +15,62 @@ The rule does not have any configuration options.
...
+- **Hint text** — The `hintText` input provides persistent inline guidance below the checkbox.
+- **Automatic error messages** — Built-in validation error messages include the `labelText` value to provide meaningful context to users.
+
+### `<sky-file-attachment>`
+
+Replaces the deprecated `<sky-file-attachment-label>` child element. Setting `labelText` enables:
+
+- **Inline help** — The `helpPopoverContent` and `helpKey` inputs only activate when `labelText` is also set.
+- **Hint text** — The `hintText` input provides persistent inline guidance below the file attachment.
+
```
Blackbaud-SandhyaRajasabeson, line 41: support for automatic error handling is also added for file attachment.

libs/sdk/skyux-eslint/docs/rules/template/prefer-label-text.md
```diff
@@ -13,6 +15,62 @@ The rule does not have any configuration options.
...
+Replaces the `<label>` child element pattern. Setting `labelText` enables:
+
+- **Automatic label association** — The label is automatically linked to the component's internal input element. No `skyId` directive, `[for]` binding, or manual `id` management is required.
+- **Inline help** — The `helpPopoverContent` and `helpKey` inputs only activate when `labelText` is also set.
+- **Hint text** — The `hintText` input provides persistent inline guidance below the input.
+- **Character count** — The `characterLimit` input places a character count indicator on the input.
+- **Automatic error messages** — Built-in validation error messages include the `labelText` value to provide meaningful context to users.
+
+When autofixing, the rule also removes the `sky-form-control` CSS class from child `<input>`, `<select>`, and `<textarea>` elements, since that class is no longer needed when `labelText` is used.
```
Blackbaud-SandhyaRajasabeson, line 52: nice QOL addition with the rule!

libs/sdk/skyux-eslint/docs/rules/template/prefer-label-text.md
```diff
@@ -13,6 +15,62 @@ The rule does not have any configuration options.
...
+- **Hint text** — The `hintText` input provides persistent inline guidance below the radio button.
+
+### `<sky-toggle-switch>`
+
+Replaces the deprecated `<sky-toggle-switch-label>` child element. Setting `labelText` enables:
+
+- **Inline help** — The `helpPopoverContent` and `helpKey` inputs only activate when `labelText` is also set.
+
+<br>
+
```
Blackbaud-SandhyaRajasabeson, line 74: Why isn't `sky-radio-group` included in this list? I know `sky-checkbox-group` came later, but I believe radio group has been there for a while and hence has deprecated inputs that directs people to use `headingText` instead.

Blackbaud-SandhyaRajasabeson, line 74: Do we want to include a blurb that including label text means they dont have to set inputs that like `labelledBy` on checkbox anymore since they are deprecated in favor of `labelText`?

Blackbaud-SteveBrush: The reason it's absent is because the the rule is designed around components with a deprecated child element (e.g. `<sky-radio-label>`, `<sky-checkbox-label>`) that needs replacing. The radio group never had a label component, so it's not needed in the rule or the docs for the rule.

Blackbaud-SteveBrush: The ESLint rule docs are probably best kept scoped to what the rule enforces and what `labelText` directly enables. Since the deprecated ARIA inputs like `ariaLabel` and `ariaLabelledBy` are already covered in the migration guide and component API docs, adding a note here might start mixing concerns. The current bullet-point format reads cleanly as-is, and developers would likely encounter the deprecation warnings at runtime when using those inputs anyway.

libs/sdk/skyux-eslint/docs/rules/template/prefer-label-text.md
Blackbaud-SteveBrush: Good point. Added!

libs/sdk/skyux-eslint/docs/rules/template/prefer-label-text.md
Blackbaud-SandhyaRajasabeson: Hmm technically part of adding `labelText` is also that labelling accessibility concerns are now handled by us, would that be a better way of wording this for the part of what the rule being enforced enables for them?

Blackbaud-SteveBrush: That works. I updated the leading paragraph since the entire purpose of the `labelText` and `headingText` inputs is to provide accessibilty features, and not necessarily a component-specific thing. Does that work?

Blackbaud-SandhyaRajasabeson (approved):

PR #4288, opened 2026-03-11
fix(components/tabs)!: replace `@angular/animations` with CSS transitions

Blackbaud-SteveBrush:
BREAKING CHANGE

`SkySectionedFormComponent` and `SkyVerticalTabsetComponent` have replaced `@angular/animations` with CSS transitions. Tests that interact with sectioned forms or vertical tabs in responsive/mobile mode may need to add `provideNoopSkyAnimations()` from `@skyux/core` to their `TestBed` providers to disable SKY UX CSS transitions during tests.

```
import { provideNoopSkyAnimations } from '@skyux/core';

  providers: [provideNoopSkyAnimations()],
});
```

[AB#3913635](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3913635)
[AB#3913634](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3913634)

## Summary by CodeRabbit

* **Refactor**
  * Modernized sectioned form and vertical tabset animation handling for improved performance and rendering efficiency.
  * Updated animation state management with reactive control patterns.

* **Tests**
  * Updated test infrastructure across all tab component tests and harnesses.

* **Chores**
  * Removed @angular/animations peer dependency.


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4288/storybook/)

Component Storybooks:

- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/4288/storybooks/action-bars-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/4288/storybooks/pages-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/4288/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/4288/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/4288/storybooks/text-editor-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4288/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4288/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4288/playground/)

libs/components/tabs/src/lib/modules/sectioned-form/sectioned-form.component.spec.ts
```diff
@@ -287,16 +287,15 @@ describe('Sectioned form component', () => {
     expect(indexChangedSpy.calls.count()).toEqual(1);
   });
 
-  it('should have a visible animation state on load in mobile', () => {
```
Blackbaud-SteveBrush, line 290: The original expectation was not correct; animations never transitioned on init, even before my changes.

libs/components/tabs/src/lib/modules/vertical-tabset/vertical-tabset.component.html
```diff
@@ -24,30 +24,31 @@
     "
     [attr.role]="isMobile ? undefined : ariaRole"
   ></span>
-  @if (maintainTabContent || tabService.tabsVisible()) {
```
Blackbaud-SteveBrush, line 27: The `@if` destroyed and recreated the group container (and the `skyTransitionEndHandler` directive on it) each time tabs toggled. A freshly created directive misses the initial trigger change, so the noop animation fallback never fired, forcing us to add workaround logic in the component.

Removing the `@if` lets the directive persist so it correctly observes the `tabsVisible()` signal transition from `false → true`, handling both real and noop animations without any branching in the component.

This isn't a problem because the element was already hidden via the `sky-vertical-tabset-hidden` class (`display: none`). The `@if` was redundant; it just added DOM churn on top of the CSS-based hiding that was already in place. The `<ng-content>` inside is projected content that exists regardless, so keeping the container in the DOM has no meaningful cost.

libs/components/code-examples/src/lib/modules/tabs/wizard/basic/modal.component.spec.ts
```diff
@@ -3,6 +3,7 @@ import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
```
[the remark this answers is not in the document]
Blackbaud-SteveBrush: Still needed for other dependencies.

libs/components/tabs/src/lib/modules/vertical-tabset/vertical-tabset.component.html
```diff
@@ -24,30 +24,31 @@
...
-      <ng-content />
-    </div>
-  }
+  <div
+    #groupContainerWrapper
+    class="sky-vertical-tabset-group-container"
+    skyTransitionEndHandler
+    transitionPropertyToTrack="transform"
+    [class.sky-animation-slide-enter-left]="tabService.isMobile()"
+    [ngClass]="{ 'sky-vertical-tabset-hidden': !tabService.tabsVisible() }"
```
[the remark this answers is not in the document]
Blackbaud-SteveBrush: Trying to avoid unnecessary changes in this PR.

libs/components/tabs/src/lib/modules/vertical-tabset/vertical-tabset.component.html
johnhwhite: If we wanted to keep the `@if` would it be reasonable to use `animate.leave` here? I'm concerned changing the behavior might have consequences.

Blackbaud-SteveBrush: The issue isn't on the "leave" side, but the "enter" side. When the `@if` recreates the tab navigation, the transition handler directive is freshly constructed.  When animations are disabled, its noop fallback `effect` intentionally skips its first signal read (to avoid emitting on init), so it never detects a change. As far as I could tell, no leave animation strategy would fix that.

The `@if` was redundant with the existing `sky-vertical-tabset-hidden` class (`display: none`). With or without the `@if`, when tabs are hidden: the element is invisible and takes no layout space, screen readers skip it, and it's not focusable. The only real difference (I can see) is that the tab-button elements stay in the DOM instead of being destroyed and recreated every time a user toggles between tabs and content on mobile. We could argue that removing the churn is better, as it avoids the cost of tearing down and rebuilding that subtree on every toggle. 

Something else to note: the `@if` remains on the tab content, so I feel like the impact to consumers is minimal.

johnhwhite: Ok, makes sense.

johnhwhite (approved):

PR #4349, opened 2026-03-31
feat(components/charts): add @skyux/charts library -- DRAFT

ThomasOrtiz:
This is a draft PR to collect early feedback on the new Charts package before merge readiness.
Not intended to merge in current state.

**Feature progress**
UX Review
- Interaction Boxes
- Keyboard Navigation
- Chart Height and Bar Chart auto sizing

TODO
- Aria review/announcements: Matt's evaluating requirements
- Testing, Visual Tests, Harnesses, 
- Examples, Demo, Documentation 

## Summary by CodeRabbit

* **New Features**
  * Added new charting library with bar, line, and donut chart components
  * Charts support customizable axes, interactive legends, and clickable data points
  * Keyboard navigation and screen reader accessibility for chart interaction
  * Added data table modal to view chart data in tabular format
  * Included interactive hover indicators and focus visualization for chart elements


apps/playground/src/app/components/charts/shared/chart-demo-utils.ts
```diff
@@ -0,0 +1,90 @@
...
+    'October',
+    'November',
+    'December',
+  ] as const;
+
+  public static random(min?: number, max?: number): number {
+    min = min ?? 0;
+    max = max ?? 0;
+    this.#seed = (this.#seed * 9301 + 49297) % 233280;
+    return min + (this.#seed / 233280) * (max - min);
```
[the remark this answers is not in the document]
ThomasOrtiz: This test utility was copied from ChartJS so leaving it as-is for now

apps/playground/src/app/components/charts/line-chart-demo/line-chart-demo.component.ts
```diff
@@ -0,0 +1,339 @@
...
+        decimals: 0,
+      }).map((value, index) => {
+        return {
+          category: ChartDemoUtils.months({ count: 12 })[index],
+          label: `$${value}`,
+          value,
+        };
+      }),
+    },
+  ];
```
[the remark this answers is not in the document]
ThomasOrtiz: This is good feedback and helps make the demo better.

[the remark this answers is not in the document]
ThomasOrtiz: Applied changes in latest commit

libs/components/charts/src/lib/modules/chart-bar/chart-bar.component.ts
```diff
@@ -0,0 +1,211 @@
...
+  // #region Dependency Injection
+  readonly #chartService = inject(SkyChartService);
+  readonly #chartRegistry = inject(SkyBarChartRegistry);
+  readonly #chartConfigService = inject(SkyBarChartConfigService);
+  // #endregion
+
+  // #region Inputs
+  public readonly orientation = input<SkyBarChartOrientation>('vertical');
+  public readonly stacked = input(false, { transform: booleanAttribute });
+  public readonly dataPointsClickable = input(false, {
```
ThomasOrtiz: @Blackbaud-SteveBrush I'll want some API feedback on this change as we have a few options. 

**Context**: UX requested that when consumers do not care about click events that we disable the Active Indicator but the new `output` signal syntax makes this tougher. 

Angular References
- This is an intentional API change: GH [Add observed property to OutputEmitterRef #54837](https://github.com/angular/angular/issues/54837)
- It sounds like Angular would rather people use an Input + Output for this behavior 
- That being said they expose outputFromObservable as a workaround

---

Previously
```typescript
@Output() public readonly dataPointClicked = new EventEmitter<SkyChartClickedDataPoint<SkyBarDatum>>();

if (this.dataPointClicked.observed) {
  // pivot on if consumer cares
}
```

**New way** - what I think Angular would prefer us to do
```typescript
public readonly dataPointsClickable = input(false, { transform: booleanAttribute });
public readonly dataPointClicked = output<SkyChartClickedDataPoint<SkyBarDatum>>();

if (dataPointsClickable()) {
	// pivot on if consumer cares
}
```

**New way** - using the [outputFromObservable](https://angular.dev/api/core/rxjs-interop/outputFromObservable) escape hatch 
```typescript
import { outputFromObservable } from '@angular/core/rxjs-interop';

readonly #dataPointClickedStream = new Subject<SkyChartClickedDataPoint<SkyBarDatum>>();
readonly dataPointClicked = outputFromObservable(this.#dataPointClickedStream);

if (this.#dataPointClickedStream.observed) {
  // pivot on if consumer cares
}
```

libs/components/charts/testing/src/modules/bar-chart/fixtures/bar-chart-harness-test.component.ts
```diff
@@ -0,0 +1,12 @@
+import { Component } from '@angular/core';
+import { provideNoopAnimations } from '@angular/platform-browser/animations';
+
+// #region Test component
+@Component({
+  selector: 'sky-bar-chart-fixture',
+  templateUrl: './bar-chart-harness-test.component.html',
+  imports: [],
+  providers: [provideNoopAnimations()],
+})
```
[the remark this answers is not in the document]
ThomasOrtiz: Removed in latest commit

libs/components/charts/ng-package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "../../../node_modules/ng-packagr/ng-package.schema.json",
+  "dest": "../../../dist/libs/components/charts",
+  "lib": {
+    "entryFile": "src/index.ts",
+    "styleIncludePaths": ["../../.."]
+  },
+  "inlineStyleLanguage": "scss",
+  "allowedNonPeerDependencies": []
```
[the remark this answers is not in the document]
ThomasOrtiz: This was addressed in a commit

libs/components/charts/README.md
```diff
@@ -0,0 +1,7 @@
+# charts
+
+This library was generated with [Nx](https://nx.dev).
+
+## Running unit tests
+
+Run `nx test datetime` to execute the unit tests.
```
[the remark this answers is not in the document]
ThomasOrtiz: Addressed

libs/components/charts/testing/src/modules/chart-bar/fixtures/chart-bar-harness-test.component.html
```diff
@@ -0,0 +1 @@
+TODO
```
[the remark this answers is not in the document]
ThomasOrtiz: This is good feedback however I have no worked on harnesses or tests yet so ignoring for now

libs/components/charts/testing/src/modules/chart-bar/chart-bar-harness.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    return { boxHarness: barChartHarness, fixture, loader };
+  }
+
+  it('should work', async () => {
+    const { fixture } = await setupTest();
+
+    fixture.detectChanges();
+
+    expect(true).toBeFalse();
+  });
```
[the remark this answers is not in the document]
ThomasOrtiz: This is good feedback however I have not worked on harnesses or tests yet so ignoring this for now.

libs/components/charts/testing/src/modules/chart-bar/chart-bar-harness.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+describe('Bar chart test harness', () => {
+  async function setupTest(
+    options: {
+      dataSkyId?: string;
+    } = {},
+  ): Promise<{
+    boxHarness: SkyBarChartHarness;
+    fixture: ComponentFixture<BarChartHarnessTestComponent>;
+    loader: HarnessLoader;
+  }> {
```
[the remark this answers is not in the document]
ThomasOrtiz: This is good feedback however I have not worked on harnesses or tests yet so ignoring this for now.

libs/components/charts/testing/src/modules/chart-bar/fixtures/chart-bar-harness-test.component.html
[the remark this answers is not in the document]
ThomasOrtiz: This is good feedback however I have not worked on harnesses or tests yet so ignoring this for now.

libs/components/charts/src/lib/modules/donut-chart/donut-chart-series-datapoint.component.ts
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * The human-readable label shown in tooltips for this data point (e.g. "$10,000").
+   */
+  public readonly labelText = input.required<string>();
+
+  /**
+   * The numeric value for this data point.
+   * Accepts a single number, a floating-bar range `[min, max]`, or `null` for a gap.
+   */
+  public readonly value = input.required<SkyDonutDatum>();
```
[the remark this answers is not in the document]
ThomasOrtiz: Addressed

libs/components/charts/src/lib/modules/chart/chart.component.ts
```diff
@@ -0,0 +1,131 @@
...
+  // #endregion
+
+  protected readonly includeHeaderMargin = computed(() => {
+    const headingVisible = !!this.headingText() && !this.headingHidden();
+    const subtitleVisible = !!this.subtitleText() && !this.subtitleHidden();
+    return headingVisible || subtitleVisible;
+  });
+  protected readonly headingClass = computed(
+    () => `sky-font-heading-${this.headingLevel()}`,
+  );
```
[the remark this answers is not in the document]
ThomasOrtiz: Good catch - addressed in a recent commit

libs/components/charts/src/lib/modules/chart-data-grid-modal/chart-data-grid-modal.component.ts
```diff
@@ -0,0 +1,145 @@
...
+    ];
+
+    // Add a column for each series
+    series.forEach((seriesItem) => {
+      columns.push({
+        field: seriesItem.labelText,
+        headerName: seriesItem.labelText,
+        minWidth: 120,
+      });
+    });
```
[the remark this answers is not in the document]
ThomasOrtiz: I believe that this implementation should be okay

libs/components/charts/src/lib/modules/chart-data-grid-modal/chart-data-grid-modal.component.ts
```diff
@@ -0,0 +1,145 @@
...
+
+      // Add data from each series for this category
+      series.forEach((seriesItem) => {
+        const dataPoint = seriesItem.data[categoryIndex];
+        const fieldName = seriesItem.labelText;
+        row[fieldName] = dataPoint.labelText;
+      });
+
+      rows.push(row);
+    });
```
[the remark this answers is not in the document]
ThomasOrtiz: Good feedback - addressed

libs/components/charts/src/lib/modules/chart-data-grid-modal/chart-data-grid-modal.component.ts
```diff
@@ -0,0 +1,145 @@
...
+    // Create a row for each category
+    categories.forEach((category, categoryIndex) => {
+      const row: ChartDataGridRow = { category };
+
+      // Add data from each series for this category
+      series.forEach((seriesItem) => {
+        const dataPoint = seriesItem.data[categoryIndex];
+        const fieldName = seriesItem.labelText;
+        row[fieldName] = dataPoint.labelText;
+      });
```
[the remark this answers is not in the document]
ThomasOrtiz: Good feedback - addressed

libs/components/charts/src/lib/modules/chart-legend/chart-legend.component.scss
```diff
@@ -0,0 +1,84 @@
+.sky-chart-legend {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
```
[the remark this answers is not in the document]
ThomasOrtiz: Addressed

libs/components/charts/src/lib/modules/chart/chart.component.scss
```diff
@@ -0,0 +1,42 @@
...
+
+sky-chart-legend {
+  grid-column: 1 / -1;
+}
+
+.sky-chart-actions {
+  grid-column: 2;
+  grid-row: 1;
+  align-self: start;
+}
```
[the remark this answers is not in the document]
ThomasOrtiz: Addressed

libs/components/charts/src/lib/modules/shared/services/chart-style.service.ts
```diff
@@ -0,0 +1,535 @@
...
+   * This method creates a temporary element, applies the CSS value to it, and then reads the computed pixel value.
+   */
+  #cssCalcNumber(varName: `--${string}`, cssFallback?: string): number {
+    const calcText = this.#css(varName, cssFallback);
+    const el = document.createElement('div');
+    el.style.opacity = calcText;
+    document.body.appendChild(el);
+
+    const out = getComputedStyle(el).opacity;
+    el.remove();
```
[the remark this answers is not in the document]
ThomasOrtiz: Addressed

libs/components/charts/src/lib/modules/shared/services/chart-style.service.ts
```diff
@@ -0,0 +1,535 @@
...
+
+    return Number(out);
```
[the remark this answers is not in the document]
ThomasOrtiz: Resolved in latest commit

libs/components/charts/src/lib/modules/chart-donut/chart-donut-config.service.ts
```diff
@@ -0,0 +1,151 @@
...
+      },
+      tooltip: {
+        intersect: false,
+        callbacks: {
+          label(context) {
+            const series = options.series;
+            const category = categories[context.dataIndex];
+            const slice = series.data.find((d) => d.category === category);
+            const percent = percentOfVisibleDataset(context);
+            return `${slice?.labelText} (${percent.toFixed(2)}%)`;
```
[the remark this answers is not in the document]
ThomasOrtiz: Using the category here is intentional as 
1. it should be a unique key per dataset. 
2. the dataset index from ChartJS does not necessarily represent the same data as what the user passed in.

libs/components/charts/src/lib/modules/shared/plugins/keyboard-nav/radial-navigation-strategy.ts
```diff
@@ -0,0 +1,87 @@
...
+    ) {
+      newIndex = (newIndex + 1) % dataLength;
+    } else if (
+      key === ChartKeys.Navigation.ArrowLeft ||
+      key === ChartKeys.Navigation.ArrowUp
+    ) {
+      newIndex = (newIndex - 1 + dataLength) % dataLength;
+    }
+
+    return { datasetIndex: current.datasetIndex, index: newIndex };
```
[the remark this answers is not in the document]
ThomasOrtiz: We can safely assume `[0]` here because we do not support multi-series donut charts

libs/components/charts/src/lib/modules/shared/plugins/keyboard-nav/cartesian-navigation-strategy.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    return {
+      seriesLabel: String(dataset?.label ?? ''),
+      seriesIndex: focused.datasetIndex + 1,
+      totalSeries: chart.data.datasets.length,
+      categoryLabel: String(chart.data.labels?.[focused.index] ?? ''),
+      value: String(dataset?.data[focused.index] ?? ''),
+      index: focused.index + 1,
+      total: dataset?.data.length ?? 0,
+    };
```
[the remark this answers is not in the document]
ThomasOrtiz: This is probably good feedback but leaving this as-is for now since UX is investigating what we want to do. 
We have a task to look at aria-live announcements

libs/components/charts/src/lib/modules/shared/plugins/indicator/line-indicator-bounds.ts
```diff
@@ -0,0 +1,50 @@
...
+import { IndicatorBounds, IndicatorStyles } from './indicator-types';
+
+export function getLineIndicatorBounds(
+  activeElements: ActiveElement[],
+  styles: IndicatorStyles,
+): IndicatorBounds {
+  const points = activeElements.map((el) => getPointGeometry(el));
+
+  return getSinglePointBounds(points[0], styles);
+}
```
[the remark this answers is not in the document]
ThomasOrtiz: Added explicit guards

libs/components/charts/src/lib/modules/shared/plugins/indicator/bar-indicator-bounds.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export function getBarIndicatorBounds(
+  chartArea: ChartArea,
+  activeElements: ActiveElement[],
+  styles: IndicatorStyles,
+): IndicatorBounds {
+  const bars = activeElements.map((el) => getBarGeometry(el));
+
+  return getSingleBarBounds(bars[0], chartArea, styles);
+}
```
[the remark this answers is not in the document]
ThomasOrtiz: Added explicit guards

libs/components/charts/src/lib/modules/shared/plugins/indicator/line-indicator-bounds.ts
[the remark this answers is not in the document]
ThomasOrtiz: Added explicit guards

libs/components/charts/src/lib/modules/shared/plugins/indicator/line-indicator-bounds.ts
```diff
@@ -0,0 +1,50 @@
...
+
+function getSinglePointBounds(
+  point: LinePointGeometry,
+  styles: IndicatorStyles,
+): IndicatorBounds {
+  const padding = styles.padding + Math.PI / 2;
+  const radius = point.radius;
+  const diameter = radius * 2;
+  const diameterWithPadding = diameter + padding * 2;
```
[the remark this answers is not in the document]
ThomasOrtiz: Addressed

libs/components/charts/src/lib/modules/donut-chart/donut-chart-registry.service.ts
```diff
@@ -0,0 +1,61 @@
...
+        // Not the series we're looking for
+        if (s.id !== seriesId) {
+          return s;
+        }
+
+        // Remove any existing point in the series with the same category
+        const dataToKeep = s.data.filter((p) => p.category !== point.category);
+        const newData = [...dataToKeep, point];
+        return { ...s, data: newData };
+      }),
```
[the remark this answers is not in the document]
ThomasOrtiz: This is acceptable for now until we hear feedback

libs/components/charts/src/lib/modules/donut-chart/donut-chart-series-datapoint.component.ts
```diff
@@ -0,0 +1,67 @@
...
+  constructor() {
+    effect(() => {
+      const datapoint = this.datapoint();
+      this.#registry.upsertPoint(this.#series.id, datapoint);
+    });
+  }
+
+  public ngOnDestroy(): void {
+    this.#registry.removePoint(this.#series.id, this.category());
+  }
```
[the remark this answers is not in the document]
ThomasOrtiz: Address the concern in all 3 chart implementations by introducing a stable point id

libs/components/charts/src/lib/modules/chart-bar/chart-bar.component.ts
```diff
@@ -0,0 +1,225 @@
...
```
Blackbaud-SteveBrush: To match the wording of its companion input, can we go with `dataPointsClickEnabled`?

libs/components/charts/src/lib/modules/chart-bar/chart-bar.component.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  /**
+   * A CSS height value (e.g. `'400px'`, `'20rem'`, `'50vh'`) for the chart.
+   * When unspecified, the chart uses internal sizing logic.
+   */
+  public readonly height = input<string>();
+  // #endregion
+
+  // #region Outputs
+  public readonly dataPointClicked =
```
Blackbaud-SteveBrush: We use present tense for events (similar to native HTML attributes): `dataPointClick`.

blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4349/storybook/)

Component Storybooks:

- [a11y-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/a11y-storybook/)
- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/avatar-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/colorpicker-storybook/)
- [core-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/core-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/errors-storybook/)
- [filter-bar-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/filter-bar-storybook/)
- [flyout-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/flyout-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/forms-storybook/)
- [help-inline-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/help-inline-storybook/)
- [icon-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/icon-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/lookup-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/modals-storybook/)
- [navbar-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/navbar-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/phone-field-storybook/)
- [popovers-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/popovers-storybook/)
- [progress-indicator-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/progress-indicator-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/split-view-storybook/)
- [storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/theme-storybook/)
- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/tiles-storybook/)
- [toast-storybook](https://blackbaud.github.io/skyux-pr-preview/4349/storybooks/toast-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4349/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4349/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4349/playground/)

libs/components/charts/src/lib/modules/axis/chart-measure-axis.component.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * The preferred lower bound for the measure axis. The chart may still go below this value if the data requires it.
+   */
+  public readonly preferredMin = input<number>();
+
+  /**
+   * The preferred upper bound for the measure axis. The chart may still exceed this value if the data requires it.
+   */
+  public readonly preferredMax = input<number>();
```
Blackbaud-SteveBrush: Let's use the `numberAttribute` transform for these four number inputs.

libs/components/charts/src/lib/modules/axis/chart-measure-axis.component.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * The upper bound for the measure axis. The chart will not exceed this value.
+   */
+  public readonly max = input<number>();
+
+  /**
+   * The preferred lower bound for the measure axis. The chart may still go below this value if the data requires it.
+   */
+  public readonly preferredMin = input<number>();
```
Blackbaud-SteveBrush: Having both `min` and `preferredMin` (and `max`/`preferredMax`) is a confusing API. They're mutually exclusive (setting both is contradictory) yet nothing prevents it structurally, so we rely on a runtime warning. The naming also doesn't clearly convey the difference in behavior.

Suggestion: Replace `preferredMin`/`preferredMax` with boolean inputs `allowMinOverflow`/`allowMaxOverflow`. When set, `min`/`max` behave as soft bounds (the axis can extend beyond them if the data requires it). This makes conflicting combinations structurally impossible and reads more naturally:

```
<!-- Hard bound: axis never goes below 0 -->
<sky-chart-measure-axis [min]="0" />

<!-- Soft bound: axis starts at 0 but extends below if data requires it -->
<sky-chart-measure-axis [min]="0" allowMinOverflow />
```

libs/components/charts/src/lib/modules/axis/chart-axis-registry.service.ts
```diff
@@ -0,0 +1,47 @@
...
+  readonly categoryAxis: Signal<SkyChartCategoryAxisConfig | undefined>;
+
+  /** Signals that emit the current measure axes configuration. */
+  readonly measureAxis: Signal<SkyChartMeasureAxisConfig | undefined>;
+
+  /**
+   * Updates or inserts the category axis configuration.
+   * @param axis
+   */
+  upsertCategoryAxis(axis: SkyChartCategoryAxisConfig): void;
```
Blackbaud-SteveBrush, line 29: The registry can only include one axis? Is there any potential future where we'd need to support multiple axes in the registry?

libs/components/charts/src/lib/modules/chart/chart.component.ts
```diff
@@ -0,0 +1,157 @@
...
+  public helpKey = input<string | undefined>();
+  // #endregion
+
+  protected readonly includeHeaderMargin = computed(() => {
+    const headingVisible = !!this.headingText() && !this.headingHidden();
+    const subtitleVisible = !!this.subtitleText() && !this.subtitleHidden();
+    return headingVisible || subtitleVisible;
+  });
+  protected readonly headingClass = computed(
+    () => `sky-font-heading-${this.headingStyle()}`,
```
Blackbaud-SteveBrush, line 121: The heading classes were [changed to `sky-theme-font-heading-*` in SKY UX 14](https://developer.blackbaud.com/skyux/design/styles/typography?docs-active-tab=classes). We can either rebase this PR against the `main` branch and use the new classes, or create a separate story to remember to make the switch when we cherry pick this into the `main` branch. (I'd prefer the former.)

libs/components/charts/src/lib/modules/bar-chart/bar-chart.component.ts
```diff
@@ -0,0 +1,225 @@
...
+  /**
+   * A CSS height value (e.g. `'400px'`, `'20rem'`, `'50vh'`) for the chart.
+   * When unspecified, the chart uses internal sizing logic.
+   */
+  public readonly height = input<string>();
+  // #endregion
+
+  // #region Outputs
+  public readonly dataPointClicked =
+    output<SkyChartClickedDataPoint<SkyBarDatum>>();
```
Blackbaud-SteveBrush: To align with other components' event params, let's name this `SkyChartDataPointClickArgs`.

libs/components/charts/src/lib/modules/chart-bar/chart-bar.component.ts
Blackbaud-SteveBrush: The "data point click" inputs are essentially copy-pasted between the charts. Is there an opportunity to isolate the shared logic in a [host directive](https://angular.dev/guide/directives/directive-composition-api#adding-directives-to-a-component)? If it makes the components too complicated, you can abandon the refactor; mostly just curious.

libs/components/charts/src/lib/modules/chart-bar/chart-bar-series.component.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * The display label for this series. Shown in the chart legend and tooltips.
+   */
+  public readonly labelText = input.required<string>();
+
+  /**
+   * A unique ID for this series component instance.
+   */
+  public readonly id = nextId++;
```
Blackbaud-SteveBrush: Does this need to be public?

libs/components/charts/src/index.ts
```diff
@@ -0,0 +1,27 @@
...
+export { SkyChartCategoryAxisComponent } from './lib/modules/axis/chart-category-axis.component';
+export { SkyChartMeasureAxisComponent } from './lib/modules/axis/chart-measure-axis.component';
+
+// Bar Chart
+export { SkyBarChartComponent } from './lib/modules/bar-chart/bar-chart.component';
+export { SkyBarChartSeriesComponent } from './lib/modules/bar-chart/bar-chart-series.component';
+export { SkyBarChartSeriesDatapointComponent } from './lib/modules/bar-chart/bar-chart-series-datapoint.component';
+export {
+  SkyBarChartOrientation,
+  SkyBarDatum,
```
Blackbaud-SteveBrush: Can this be named `SkyChartBarDatum`?

libs/components/charts/src/index.ts
```diff
@@ -0,0 +1,27 @@
...
+} from './lib/modules/bar-chart/bar-chart-types';
+
+// Line Chart
+export { SkyLineChartComponent } from './lib/modules/line-chart/line-chart.component';
+export { SkyLineChartSeriesComponent } from './lib/modules/line-chart/line-chart-series.component';
+export { SkyLineChartSeriesDatapointComponent } from './lib/modules/line-chart/line-chart-series-datapoint.component';
+export { SkyLineDatum } from './lib/modules/line-chart/line-chart-types';
```
Blackbaud-SteveBrush: Can this be named `SkyChartLineDatum`?

libs/components/charts/src/index.ts
```diff
@@ -0,0 +1,27 @@
...
+
+// Donut Chart
+export { SkyDonutChartComponent } from './lib/modules/donut-chart/donut-chart.component';
+export { SkyDonutChartSeriesComponent } from './lib/modules/donut-chart/donut-chart-series.component';
+export { SkyDonutChartSeriesDatapointComponent } from './lib/modules/donut-chart/donut-chart-series-datapoint.component';
+export { SkyDonutDatum } from './lib/modules/donut-chart/donut-chart-types';
```
Blackbaud-SteveBrush: Can this be named `SkyChartDonutDatum`?

libs/components/charts/src/lib/modules/bar-chart/bar-chart-series.component.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Represents a named data series in a chart.
+ */
+@Component({
+  selector: 'sky-bar-chart-series',
+  template: '',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class SkyBarChartSeriesComponent implements OnDestroy {
```
Blackbaud-SteveBrush: The series and series datapoint components are nearly identical between the three chart types. Is there a way to make these two components sharable between all three?

```
sky-chart-datapoint
sky-chart-datapoint-series
```

libs/components/charts/src/lib/modules/bar-chart/bar-chart.component.ts
```diff
@@ -0,0 +1,225 @@
...
+  SkyBarChartOrientation,
+  SkyBarChartPoint,
+  SkyBarDatum,
+} from './bar-chart-types';
+
+/**
+ * Displays a bar chart visualization.
+ */
+@Component({
+  selector: 'sky-bar-chart',
```
Blackbaud-SteveBrush: It's preferred that all components in a library have the same selector prefix. Let's name the three chart types: `sky-chart-bar`, `sky-chart-line`, and `sky-chart-donut`.

libs/components/charts/src/lib/modules/axis/chart-axis-registry.service.ts
ThomasOrtiz: That's correct . The current design only support a single **Category Axis** and a single **Measure Axis**.

ChartJS does support defining additional axes - [example](https://www.chartjs.org/docs/latest/samples/line/multi-axis.html) - but after talking to UX they said they were not aware of any existing charts or people that need this feature. That being said it wouldn't surprise me if this is an ask one day.

To support a 2nd measure axis - which I think would be the main use case - I think we'd have to
1. Update `SkyChartAxisRegistry` to allow saving multiple `measureAxis`
2. Update MeasureAxis with a `axisId`.
3. Update the SeriesComponents to specify an `axisId` they're associated with
4. Update the Cartesian Charts (Bar/Line/ect) ChartJS config generation 

<details>
<summary>Multi Axis example</summary>

image

</details>

libs/components/charts/src/lib/modules/bar-chart/bar-chart-series.component.ts
ThomasOrtiz: Its true that in the current state they are very similar however those types are likely to diverge in the future.

---

I recently simplified the `Sky[ChartType]Datum` after UX and I decided we weren't going to support Floating Bars for MVP and that we could currently support `null` as the absence of a `Sky[ChartType]Datum` for a given category.

**Previously**
- `SkyBarDatum`: `number | [number, number] | null`
   - `number` supports simple bars
   - `[number, number]` supports [floating bars](https://www.chartjs.org/docs/latest/samples/bar/floating.html)
- `SkyLineDatum`: `number | null`
   -  Line Chart Points do not have a similar concept to floating bars
- `SkyDonutDatum`: `number | null`



**Null Handling**
`null` can have semantic meaning for **Line Charts** but today we handle null the same way across charts. 

How charts handles `null`
- Bar: bar is not rendered
- Pie / Doughnut: slice is not rendered
- Radar / Polar: value is not rendered
- **Line: by default renders a gap but this is configurable** 
   - Line Charts have a `spanGaps (default: false)` option 
   - When `spanGaps: false`: the point isn't drawn and the line is broken at the category position
   - When `spanGaps: true`: the point isn't drawn but a line is drawn between between the previous and next category.

In the future if we needed to support Line Chart's Gap/NoGap option instead of explicitly supporting `null` on the `LineDatumValue` we could instead introduce some LineChart-only option like `spanGaps: boolean`.

**Copilot Summary**
- Unique cases 
   - **Bar Charts** with Floating Bars: `[start, end]`
   - **Bubble Charts** with a Value + Radius: `{ y, r }`
   - **Scatter Charts**  with a Point `[x, y]` or `{ x, y }`
- Every other chart can be modelled with `number | null`

Image

libs/components/charts/src/lib/modules/axis/chart-measure-axis.component.ts
ThomasOrtiz: I'll have to double-check but I think what you're recommending should be fine. When modelling this I initially landed on something similar but ended up going to the current state after seeing that ChartJS/HighCharts/AgCharts break out explicit settings.

| Library | Hard Bounds | Soft Bounds | Zero Inclusion |
| --- | --- | --- | --- |
| **Chart.js** | `min` / `max` | `suggestedMin` / `suggestedMax` | `beginAtZero` |
| **Highcharts** | `min` / `max` | `softMin` / `softMax` | via `softMin: 0` |
| **ECharts** | `min` / `max` | _(none)_ | `scale` boolean |
| **AG Charts** | `min` / `max` | `preferredMin` / `preferredMax` | via `preferredMin: 0` |
| **D3** | `domain([min, max])` | `nice()` | _(manual)_ |


**How ChartJS computes scale bounds**
1. Hard **min/max** override data-derived limits (so they dominate on that side).
2. Otherwise compute from data; if stacking is enabled, that affects the computed data range.
3. Apply **suggestedMin/suggestedMax** as soft expansion using the documented Math.min/Math.max rule.
4. Apply **beginAtZero** (linear) to include 0 if needed.
5. Apply **grace** to pad the computed range.
6. Tick generation + bounds selects final visible bounds strategy ('ticks' vs 'data').
7. Callbacks like afterDataLimits can mutate after calculation (most powerful, but also most “internal”).

libs/components/charts/src/lib/modules/chart-bar/chart-bar-series.component.ts
ThomasOrtiz: This is public so that the Data Point Component - `SkyBarChartSeriesDatapointComponent` - can directly associate a Data Point to a specific Series
```
effect(() => {
  const datapoint = this.#datapoint();
  this.#registry.upsertPoint(this.#series.id, datapoint);
});
```

Some options:
1. Add `/** @internal */` annotation to the `SeriesComponent.id`.
2. Update `SeriesComponent.id` to be private and expose a `upsertPoint` / `removePoint` instead that the DataPointComponent calls with its `Id`
3. Update `SeriesComponent` to handle all of the registration (Series and DataPoint) by using `ContentChildren`
	- When I tried this originally I ran into a few issues around accessing a nested component's `input.required` values. If I remember correctly that parent components need to wait to access the Required Inputs until `AfterContentInit` is ran otherwise you hit a required input exception.

libs/components/charts/src/lib/modules/chart/chart.component.ts
ThomasOrtiz: > rebase this PR against the main **branch**

This branch is currently targeting **13.x.x**. Are you saying we should go ahead and start targeting `main` instead?

libs/components/charts/src/lib/modules/chart-bar/chart-bar.component.ts
ThomasOrtiz: Addressed in 2d9afe136ecdc85d3104093049583564882501bf

libs/components/charts/src/lib/modules/chart-bar/chart-bar.component.ts
ThomasOrtiz: Addressed in 2d9afe136ecdc85d3104093049583564882501bf

libs/components/charts/src/lib/modules/axis/chart-measure-axis.component.ts
ThomasOrtiz: Addressed in 
- 01ee3730ffe82856c0044c36bb288d350d94459d
- d42c355bb3a90d1213499bfef822f5ae9fc4a88a

libs/components/charts/src/lib/modules/axis/chart-measure-axis.component.ts
ThomasOrtiz: Went ahead with your suggestion but if we decide we like the parity with the other charting libraries it shouldn't be difficult to rollback

Addressed in 4ac0139f5e065bed8c30e4911b6c61765ab6c88a

libs/components/charts/src/index.ts
ThomasOrtiz: Addressed in 8d11b80790b2661ae4af1f9a0f041c6e37d5e7b5

libs/components/charts/src/index.ts
ThomasOrtiz: Addressed in 8d11b80790b2661ae4af1f9a0f041c6e37d5e7b5

libs/components/charts/src/index.ts
ThomasOrtiz: Addressed in 8d11b80790b2661ae4af1f9a0f041c6e37d5e7b5

libs/components/charts/src/lib/modules/bar-chart/bar-chart.component.ts
ThomasOrtiz: Addressed in 8194c178556971208746124919e049eb870fa012

libs/components/charts/src/lib/modules/bar-chart/bar-chart.component.ts
ThomasOrtiz: Addressed in 9c5c540a623246f94ab95b1c16f76122468f702e

libs/components/charts/src/lib/modules/chart-bar/chart-bar.component.ts
ThomasOrtiz: Unfortunately Host Directives don't support generics and since the `dataPointClick` type is dependent on the datum type `SkyChartDataPointClickArgs<SkyChart[Type]Datum>`.

We can use a Host Directive here but the consumer loses the `output` type's specificity around the generic: `output<SkyChartDataPointClickArgs<unknown>>()`.

libs/components/charts/src/lib/modules/chart/chart.component.ts
Blackbaud-SteveBrush: Yes; we generally do not release new features in LTS versions (in this case, v13), so it's preferred that the charts be released in v14 (the `main` branch). What do you think?

libs/components/charts/src/lib/modules/axis/chart-axis-registry.service.ts
Blackbaud-SteveBrush: The axis registry service is an internal detail, so we can add support for multiple axes in the future without a breaking change, yes?

libs/components/charts/src/lib/modules/chart-bar/chart-bar.component.ts
Blackbaud-SteveBrush: All of the "datum" types are set to a type of `number`. Do we need a unique `SkyChart[Type]Datum` type for each datum, or can they all use a shared `SkyChartDatum` type?

libs/components/charts/src/lib/modules/bar-chart/bar-chart-series.component.ts
Blackbaud-SteveBrush: > Its true that in the current state they are very similar however those types are likely to diverge in the future.

The summary after this line seems to indicate that all datum types can be represented by `number | null`. Is that true? If so, does that mean we could simplify the three types of datapoint components into a single `sky-chart-datapoint` component?

libs/components/charts/src/lib/modules/chart-bar/chart-bar-series.component.ts
Blackbaud-SteveBrush: Is there an issue with using `AfterContentInit` to register the datapoints? We like to avoid public properties on components since they make it easier for consumers to make bad choices (even if we mark the field as internal or readonly, it's simple enough to override those conventions in JavaScript). This should be avoided even more so in this case since the `SkyChartBarSeriesComponent` is being exported by the public API (and not obscured with a lambda name).

libs/components/charts/src/lib/modules/axis/chart-axis-registry.service.ts
ThomasOrtiz: Yes, I believe thats right.

libs/components/charts/src/lib/modules/chart/chart.component.ts
ThomasOrtiz: Releasing in 14.x.x sounds good to me.

libs/components/charts/src/lib/modules/chart-bar/chart-bar-series.component.ts
ThomasOrtiz: Found a way to remove the `id` from the public-api with the current structure by providing a `SKY_CHART_{TYPE}_SERIES_ID` that the series and its children can inject to use.

That keeps each component self-contained with the understanding of how to register itself without leaking internals. If you want to switch to the `contentChildren + afterContentInit` approach I believe that would be fine too.

libs/components/charts/src/lib/modules/chart-bar/chart-bar.component.ts
ThomasOrtiz: Decision aroud this is blocked pending the other discussion around data point types

libs/components/charts/src/lib/modules/bar-chart/bar-chart-series.component.ts
ThomasOrtiz: No, while the majority can be represented with a single `value: number | null` there are some unique cases that need an additional value.

When looking at the table we have these formats
- Base case: `datum: number` - supports the majority of use cases.
- Tuple of two numbers: `datum: [start: number, end: number]` to support [Floating Bars](https://www.chartjs.org/docs/latest/samples/bar/floating.html).
- An object of numbers: `datum: { value: number, radius: number}` to support [Bubble Charts](https://www.chartjs.org/docs/latest/samples/other-charts/bubble.html).

PR #4430, opened 2026-06-03
feat(components/data-grid): create an easy mode for data grid

johnhwhite:
[AB#3942528](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3942528) [AB#3965849](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3965849)

## MVP implementation:

* Async data handled via Angular Resources (stable in v22) rather than `externalRowCount` / `pageRequest` / `rowCountChange` pattern.
* Filter bar intentionally not tightly coupled; consumers wire their own filter source.
* Data manager integration not included; will revisit later.
* Row deletion flow not included; will revisit later.
* Other legacy inputs dropped: height, width, multiselectRowId, rowHighlightedId, column description, hiddenColumnIds, displayedColumnIds, and displayedColumnIdsChange output; prioritizing leaner API rather than ease of migration.

## Summary by CodeRabbit

* **New Features**
  * New SkyUX data-grid package with sorting, pagination, multiselect, row-delete flows, inline help, and public components.
  * Storybook demos and interactive playground examples added.

* **Tests**
  * End-to-end Cypress specs and visual regression snapshots; expanded unit/integration tests and a testing harness for data-grid.

* **Documentation**
  * README and documentation metadata linking examples to docs.

* **Chores**
  * SKY UX package versions updated to 14.3.7; build/config updated for the new library.


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4430/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4430/storybooks/ag-grid-storybook/)
- [data-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4430/storybooks/data-grid-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4430/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4430/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4430/playground/)

apps/code-examples/src/assets/stack-blitz/package.json
```diff
@@ -21,51 +21,51 @@
...
+    "@skyux/progress-indicator": "^14.3.7",
+    "@skyux/router": "^14.3.7",
+    "@skyux/select-field": "^14.3.7",
+    "@skyux/split-view": "^14.3.7",
+    "@skyux/tabs": "^14.3.7",
+    "@skyux/text-editor": "^14.3.7",
+    "@skyux/theme": "^14.3.7",
+    "@skyux/tiles": "^14.3.7",
+    "@skyux/toast": "^14.3.7",
+    "@skyux/validation": "^14.3.7",
```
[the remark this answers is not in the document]
johnhwhite: This cannot be added until `@skyux/data-grid` is published, which won't happen until this PR merges and ships.

libs/components/code-examples/src/lib/modules/data-grid/paging/example.component.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+import { DataGridPagingExampleComponent } from './example.component';
+
+describe('Data grid paging example', () => {
+  async function setupTest(): Promise<{
+    fixture: ComponentFixture<DataGridPagingExampleComponent>;
+    loader: HarnessLoader;
+  }> {
+    await TestBed.configureTestingModule({
+      imports: [DataGridPagingExampleComponent],
+      providers: [provideRouter([])],
```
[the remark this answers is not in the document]
johnhwhite: I clarified the guidelines. Providing `provideNoopSkyAnimations()` is not necessary for all tests.

libs/components/code-examples/src/lib/modules/data-grid/paging/example.component.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    await pagingHarness.clickNextButton();
+    await expectAsync(pagingHarness.getCurrentPage()).toBeResolvedTo(2);
+
+    await pagingHarness.clickPreviousButton();
+    await expectAsync(pagingHarness.getCurrentPage()).toBeResolvedTo(1);
+
+    await pagingHarness.clickPageButton(2);
+    await expectAsync(pagingHarness.getCurrentPage()).toBeResolvedTo(2);
+  });
```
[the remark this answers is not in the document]
johnhwhite: I clarified the guidelines. The `fixture.detectChanges()` and `fixture.whenStable()` is handled within the test harnesses and adding another round of change detection cycles is entirely unnecessary.

libs/components/code-examples/src/lib/modules/data-grid/paging/example.component.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+      gridHarness.getDisplayedColumnHeaderNames(),
+    ).toBeResolvedTo([
+      'Name',
+      'Age',
+      'Start date',
+      'End date',
+      'Department',
+      'Job title',
+    ]);
+  });
```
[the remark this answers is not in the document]
johnhwhite: I believe I have satisfied the requirements here. What else is needed? Or can this be resolved?

apps/e2e/data-grid-storybook/src/app/data-grid/data-grid.component.html
```diff
@@ -0,0 +1,72 @@
+<div class="sky-padding-even-lg">
```
Blackbaud-ErikaMcVey: can we use the new public API classes? `sky-padding-even-lg` will be going away

apps/playground/src/app/components/data-grid/basic/data-grid.component.html
```diff
@@ -0,0 +1,192 @@
...
+  class="sky-btn sky-btn-primary"
+  (click)="dataForSimpleGrid = undefined"
+>
+  No data
+</button>
+<div id="screenshot-grid">
+  <sky-data-grid height="200" width="600" [data]="dataForSimpleGrid">
+    <sky-data-grid-column field="column1" headingText="Column1" />
+    <sky-data-grid-column field="column2" headingText="Column2" />
+    @if (!hideCol3()) {
```
Blackbaud-ErikaMcVey, line 17: think there is any benefit to this being an input on the column? If we do end up providing some kind of data manager integration, the column picker might use that?

Edit: actually, looks like it already is an input on the column?

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,642 @@
...
+  public readonly width = input<number, unknown>(0, {
+    transform: (val: unknown) => coerceNumberProperty(val, 0),
+  });
+
+  /**
+   * Fires when columns change. This includes changes to the displayed columns and changes
+   * to the order of columns. The event emits an array of IDs for the displayed columns that
+   * reflects the column order.
+   */
+  public readonly displayedColumnIdsChange = output<string[]>();
```
Blackbaud-ErikaMcVey: do we need this now? How might a consumer use it? It seems like they have full control over the columns shown right now, nothing external would be changing the displayed columns

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,642 @@
...
+  /**
+   * The current page number of the grid when `pageSize` has been set.
+   */
+  public readonly page = model<number>(1);
+
+  /**
+   * The set of IDs for the rows to prompt for delete confirmation.
+   * The IDs match the `multiselectRowId` properties of the `data` objects.
+   */
+  public readonly rowDeleteIds = model<string[]>([]);
```
Blackbaud-ErikaMcVey: multiselect and inline row delete weren't on our original feature list, and they add quite a bit to the API--do you think they're worth the additional 6 inputs/outputs?

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
```diff
@@ -0,0 +1,642 @@
...
+  public readonly compact = input<boolean, unknown>(false, {
+    transform: coerceBooleanProperty,
+  });
+
+  /**
+   * The data for the grid. Each item requires an `id`, and other properties should map to a `field` of the grid columns.
+   * When `data` is `null` or `undefined`, the grid will show a loading indicator, and when `data` is an empty array,
+   * the grid will show a "no rows" message.
+   */
+  public readonly data = input<T[] | null | undefined>();
```
Blackbaud-ErikaMcVey: I actually can't tell from the AI generated notes, but I had been under the impression we did want to support asynchronous data in the MVP--was that not the case?

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
```diff
@@ -0,0 +1,169 @@
...
+   * @default 'text'
+   */
+  public readonly dataType = input<'text' | 'number' | 'date' | 'boolean'>(
+    'text',
+  );
+
+  /**
+   * The description for the column.
+   */
+  public readonly description = input<string>();
```
Blackbaud-ErikaMcVey: how is the description used currently?

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  /**
+   * The content of the help popover. When specified, a [help inline](https://developer.blackbaud.com/skyux/components/help-inline)
+   * button is added to the column header. The help inline button displays a [popover](https://developer.blackbaud.com/skyux/components/popover)
+   * when clicked using the specified content and optional title.
+   */
+  public readonly helpPopoverContent = input<string | TemplateRef<unknown>>();
+
+  /**
+   * Whether the column is initially hidden when grid `selectedColumnIds` are not provided.
```
Blackbaud-ErikaMcVey: "when `selectedColumnIds` are not provided." I don't actually see that as an input?

libs/components/data-grid/src/lib/modules/data-grid/data-grid.ts
```diff
@@ -0,0 +1,642 @@
...
+    }
+    return classes;
+  });
+
+  readonly #activatedRoute = inject(ActivatedRoute, { optional: true });
+  readonly #gridService = inject(SkyAgGridService);
+  readonly #router = inject(Router, { optional: true });
+
+  readonly #columnDefs = computed<ColDef<T>[]>(() => {
+    const columns = this.columns();
```
Blackbaud-ErikaMcVey, line 387: how does this work if the inputs on a column are changed? if someone programmatically shows/hides a column or makes it sortable or something, will the column definitions update?

Blackbaud-ErikaMcVey:
My first pass was mostly just testing out the playground/looking at the public API.

Also, something odd happens when sorting the paged grids, the top row is no longer visible? Interestingly, if you go to the next page then go back, you can see the top row again
image

apps/e2e/data-grid-storybook/src/app/data-grid/data-grid.component.html
johnhwhite: Good catch. Fixed.

apps/playground/src/app/components/data-grid/basic/data-grid.component.html
johnhwhite: The input is on the column. This is demonstrating what happens with the template conditional.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
johnhwhite: Good catch. I've removed it.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
johnhwhite: Fair. I've removed row-delete as that API is more complex. I'm leaving multiselect for now.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.component.ts
johnhwhite: Good catch. I've added support for Angular's new `Resource` type (stable in 22).

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
johnhwhite: Good catch. It was brought over from the grids api but it is redundant.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.component.ts
johnhwhite: Good catch. That input was dropped. Updated these docs.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.ts
johnhwhite: Yes, it's updating `GridOptions` whenever a condition changes via an `effect`.

apps/playground/src/app/components/data-grid/basic/data-grid.component.html
Blackbaud-ErikaMcVey: ok--so if you make the input conditional, does it update the grid options accordingly? like would the below work?
```
<sky-data-grid-column field="column2" headingText="Column2" [hidden]="hideCol3()" />
```

Edit: apparently I asked that already a few weeks ago and you answered below 😂

libs/components/data-grid/src/lib/modules/types/data-grid-sort.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Applies a sort to a `SkyDataGrid` and reflects updates when the grid's sort changes.
```
Blackbaud-SteveBrush, line 2: [api] Needs `@preview`.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.ts
```diff
@@ -0,0 +1,681 @@
...
+   */
+  public readonly pageSize = input<number, unknown>(0, {
+    transform: (value: unknown) => coerceNumberProperty(value, 0),
+  });
+
+  /**
+   * The query parameter name that stores the current page number.
+   * When set, the grid syncs page changes to the URL for deep linking, and there should only be one grid on the page.
+   */
+  public readonly pageQueryParam = input<string | undefined>();
```
Blackbaud-SteveBrush, line 220: [bug] A consumer who sets `pageQueryParam` _and_ drives `[(page)]` programmatically gets a grid that moves pages correctly but the URL param does not update. To test: create a button that sets the page number programmatically using `page`; the URL param does not update.

libs/components/data-grid/src/lib/modules/types/data-grid-sort.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+export interface SkyDataGridSort<
+  T extends Record<'id', string> = Record<'id', string> &
```
Blackbaud-SteveBrush: [api] The generic `T` on `SkyDataGrid`/`SkyDataGridSort` currently only buys one thing for consumers: `keyof T` autocomplete/validation on `sortField.fieldSelector`. It doesn't extend to column field (a plain string) or the cell template context, so the safety is inconsistent. Meanwhile it imposes real friction for the consumer: it forces handlers to be parameterized (`SkyDataGridSort<Constituent>` instead of `SkyDataGridSort`), and because `fieldSelector: keyof T` is an invariant position, mismatched instantiations produce confusing "missing properties" errors. Recommend dropping the generic and typing `fieldSelector` as `string`; given the columns and template are already stringly-typed, this will make implementing handlers easier for our consumers.

If you want to keep the generic, we need to figure out a way to make it optional. The following usage throws an error in the console when running `ng build`:

```typescript
// ❌ Plain `SkyDataGridSort` defaults its generic, so it doesn't match
//    the `SkyDataGridSort<Constituent>` the grid emits:
//    "Type 'Record<"id", string> & Record<string, string>' is missing the
//     following properties from type 'Constituent': first_name, last_name, ..."
protected onSortChange(sort: SkyDataGridSort | undefined): void { /* ... */ }

// ✅ Consumer must repeat the row type to line up with the inferred `T`:
protected onSortChange(sort: SkyDataGridSort<Constituent> | undefined): void { /* ... */ }
```

libs/components/data-grid/src/lib/modules/data-grid/data-grid.ts
```diff
@@ -0,0 +1,681 @@
...
+   * How the grid fits to its parent. The valid options are `width`,
+   * which fits the grid to the parent's full width, and `scroll`, which allows the grid
+   * to exceed the parent's width. If the grid does not have enough columns to fill
+   * the parent's width, it always stretches to the parent's full width.
+   * @default 'width'
+   */
+  public readonly fit = input<'width' | 'scroll'>('width');
+
+  /**
+   * Whether data is being loaded. When this is true, the grid shows a waiting overlay and is not interactive.
```
Blackbaud-SteveBrush: [docs] A "loading" state can be active even when `loading=false` by setting `data` to null. Probably fine, but we need to update the docs to inform the consumer.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  /**
+   * The width of the column in pixels. When no width is set, the column width is evenly distributed.
+   */
+  public readonly width = input<number, unknown>(0, {
+    transform: (value) => numberAttribute(value, 0),
+  });
+
+  /**
+   * Whether text in this column should wrap to multiple lines. For best performance, large grids should set a `height`
```
Blackbaud-SteveBrush: [docs] The `wrapText` doc on `sky-data-grid-column` tells consumers to "set a `height`" to keep rows virtualized, but there is no `height`/`maxHeight` input — `domLayout` is hardcoded to `'autoHeight'` and the only sizing input is `minHeight`, so the grid always draws every row. Either add a `height`/`maxHeight` input that enables virtualization, or remove the "set a height" guidance and state that the grid always auto-sizes to its rows (so paging is the only way to bound row count).

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.ts
```diff
@@ -0,0 +1,166 @@
...
+   * use `field`; the column's ID defaults to the `field` value.
+   */
+  public readonly field = input<string>();
+
+  /**
+   * When set, `flexWidth` overrides `width` and works like
+   * [CSS flex-grow](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow), where a column
+   * with `flexWidth="2"` is twice the width of a column with `flexWidth="1"`, and `flexWidth="0"` does not auto-expand.
+   */
+  public readonly flexWidth = input<number, unknown>(-1, {
```
Blackbaud-SteveBrush: [api] Replace the magic-number sentinel on `flexWidth` (`-1` = "not set") with `input<number | undefined>()`, where `undefined` cleanly means "not set." The `-1` is especially opaque since `flexWidth="0"` is already a documented, meaningful value.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.ts
```diff
@@ -0,0 +1,166 @@
...
+   * of the `sky-data-grid-column` component. The template has access to the `value` variable,
+   * which contains the value passed to the column, and the `row` variable, which contains
+   * the entire row data.
+   */
+  public readonly template = input<TemplateRef<unknown>>();
+
+  /**
+   * The width of the column in pixels. When no width is set, the column width is evenly distributed.
+   */
+  public readonly width = input<number, unknown>(0, {
```
Blackbaud-SteveBrush: [api] Replace the magic-number sentinel on `width` (`0` = "auto") with `input<number | undefined>()`, where `undefined` cleanly means "not set." A zero value isn't clear that this means "auto" width.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.ts
```diff
@@ -0,0 +1,681 @@
...
+  public readonly data = input<T[] | null | undefined>();
+
+  /**
+   * How the grid fits to its parent. The valid options are `width`,
+   * which fits the grid to the parent's full width, and `scroll`, which allows the grid
+   * to exceed the parent's width. If the grid does not have enough columns to fill
+   * the parent's width, it always stretches to the parent's full width.
+   * @default 'width'
+   */
+  public readonly fit = input<'width' | 'scroll'>('width');
```
Blackbaud-SteveBrush: [api] Rename the `fit` input to `columnFit` and its values from `'width' | 'scroll'` to `'container' | 'content'`. The current values mix two different concepts — 'width' describes what the grid fits to, while 'scroll' describes a side effect of the behavior — so fit="scroll" doesn't read cleanly and a reader has to consult the docs to understand it. 'container' and 'content' are parallel nouns that both answer "what does the grid fit to?", which makes the call site self-documenting (`columnFit="container"` constrains the grid to its parent's width; `columnFit="content"` sizes to the content and allows horizontal overflow).

libs/components/data-grid/src/lib/modules/types/data-grid-sort.ts
```diff
@@ -0,0 +1,17 @@
...
+    Record<string, string>,
+> {
+  /**
+   * Whether to apply the sort in descending order.
+   */
+  descending?: boolean;
```
Blackbaud-SteveBrush: [api] Rename the `SkyDataGridSort` properties to `{ direction: 'asc' | 'desc'; field: keyof T; }`. `field` matches the column's `field` input, and `direction: 'asc' | 'desc'` replaces the optional `descending` boolean — which ambiguously treats both `false` and `undefined` as ascending — with two explicit, self-documenting values.

libs/components/data-grid/src/lib/modules/types/data-grid-sort.ts
johnhwhite: Good catch. Fixed.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.ts
johnhwhite: Good catch. Fixed and added a test.

libs/components/data-grid/src/lib/modules/types/data-grid-sort.ts
johnhwhite: I've dropped the generic from `SkyDataGridSort`.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.ts
johnhwhite: Updated the docs.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.ts
johnhwhite: Good catch. I've removed the piece about height and performance.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.ts
johnhwhite: I've removed the magic-number sentinels.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.ts
johnhwhite: Updated.

libs/components/data-grid/src/lib/modules/types/data-grid-sort.ts
johnhwhite: Updated.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.ts
```diff
@@ -0,0 +1,799 @@
...
+  ],
+  templateUrl: './data-grid.html',
+  styleUrl: './data-grid.css',
+  host: {
+    '[class.sky-margin-stacked-lg]': 'stacked()',
+  },
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class SkyDataGrid<
+  T extends Record<'id', string> = Record<'id', string> &
```
Blackbaud-SteveBrush: Now that we've removed the generic from `SkyDataGridSort`, the generic type parameter `T` on `SkyDataGrid` isn't earning its complexity. It does appear on the public surface via the `data` input, but it's inert: `T` is only ever inferred from `data` and never consumed by any other public member. No consumer supplies a type argument (nothing in the fixtures, code examples, specs, or harness writes `SkyDataGrid<...>`), and because no other input/output (nor the `SkyDataGridColumn` children) is typed in terms of `T`, the inferred type has nowhere to propagate. Everything else that references `T` is internal, and the only thing that actually depends on the row's shape is `row.id` (which needs just the `{ id: string }` constraint, not a full generic). So the generic adds a complex signature to the public API without buying any call-site type safety the constraint alone wouldn't.

I recommend dropping the generic parameter and instead enforcing the `id` requirement through the `data` input type, which is where it now matters. Export a named interface and use it as the input type:

```
export interface SkyDataGridRowData {
  id: string;
}

public readonly data = input.required<SkyDataGridRowData[] | null | undefined>();
```
Because binding is structural, this requires every item to have a string `id` while still allowing any extra fields, and a shape missing `id` fails template type checking right at the `[data]` binding; no generic argument needed. It also gives consumers a named contract they can implement.

Blackbaud-SteveBrush (approved):

blackbaud-sky-build-user:
Cherry-pick successful! :cherries: :tada: See https://github.com/blackbaud/skyux/pull/4486

PR #4437, opened 2026-06-08
chore(dev-infra): improve AI agent instructions, skills, and MCP config

Blackbaud-SteveBrush:
[AB#3963505](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3963505)
[AB#3963503](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3963503)

## Summary by CodeRabbit

* **Documentation**
  * Clarified SCSS override workflow and tightened compat-mixin instructions; narrowed where Angular/unit-testing/visual-testing guidance applies
  * Added commit-message rules (Conventional Commits) and updated agent/public-export conventions
  * Added multiple SKY UX workflow guides and skills for scaffolding components/libraries, harnesses, unit tests, visual tests, code examples, and related checklists

* **Tests**
  * Extended e2e schematic tests to cover ESLint/Storybook generation behavior

* **Chores**
  * Added Nx MCP server entry for development tooling


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4437/storybook/)

Component Storybooks:

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4437/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4437/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4437/playground/)

.github/skills/add-skyux-component/SKILL.md
```diff
@@ -0,0 +1,103 @@
...
+## Conventions at a Glance
+
+- Library source lives at
+  `libs/components/<library>/src/lib/modules/<module>/`.
+- Component files follow the pattern:
+  - `<name>.component.ts`, `.html`, `.scss`, `.spec.ts`
+  - a `NgModule` (`<name>.module.ts`) that imports the standalone component
+    and `exports` the public-facing one.
+- Components are **standalone**; do NOT set `standalone: true` explicitly in
+  the decorator. Use `ChangeDetectionStrategy.OnPush`, `input()`/`output()`,
```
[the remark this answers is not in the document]
Blackbaud-SteveBrush: Skipping — the skill is already correct: SKY UX uses standalone components (Angular 21's default, so `standalone: true` is intentionally omitted) distributed via a thin `NgModule` wrapper, and the `standalone: false` files cited are legacy components mid-migration, not the convention for new code.

.github/skills/add-skyux-code-example/SKILL.md
```diff
@@ -0,0 +1,114 @@
...
+  if a sibling example needs them. Use `example.*`, never `demo.*`.
+- **Component class name**: `<Library><Component><ExampleName>ExampleComponent`
+  (e.g. `FormsInputBoxBasicExampleComponent`,
+  `ListsRepeaterInlineFormExampleComponent`). When a library has a single
+  example, the segments may collapse (e.g. `avatar/avatar/` →
+  `AvatarExampleComponent`); follow the sibling you are mirroring.
+- The component is **standalone** (no `standalone: true` in the decorator),
+  uses `ChangeDetectionStrategy` defaults of a sibling, imports the SKY UX
+  module(s) it demonstrates, and carries a `/** @title ... */` JSDoc above the
+  class describing the example.
```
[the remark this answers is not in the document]
Blackbaud-SteveBrush: @coderabbitai Components are standalone by default, so `standalone: true` should not be used. Does that change your feedback?

AGENTS.md
```diff
@@ -46,14 +47,20 @@ before working on matching files. Consult these when your change touches their
 scope:
 
 - [angular.instructions.md](.github/instructions/angular.instructions.md)
-  (`applyTo: **`) — Angular best practices for maintainable, performant, and
-  accessible code. Applies to all Angular work.
+  (`applyTo: **/*.{ts,html}`) — Angular best practices for maintainable,
+  performant, and accessible code. Applies to all Angular work.
+- [component-unit-testing.instructions.md](.github/instructions/component-unit-testing.instructions.md)
```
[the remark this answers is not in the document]
Blackbaud-SteveBrush: Done.

.github/instructions/add-scss-override.instructions.md
```diff
@@ -1,19 +1,17 @@
 ---
-applyTo: '**/*.scss'
+applyTo: 'libs/components/**/*.scss'
```
Blackbaud-ErikaMcVey: yeah, I think this can be deleted at this point--it was useful when tokenizing components, but now that that is done, I don't think we use it much

.github/instructions/component-unit-testing.instructions.md
```diff
@@ -0,0 +1,167 @@
...
+tests, so computed styles and accessibility are assertable.
+
+This guidance is for **library source specs**. It does not cover:
+
+- **Code-example specs** (`libs/components/code-examples/**`) — follow
+  [code-examples-unit-testing.instructions.md](./code-examples-unit-testing.instructions.md).
+- **Harness specs** (`libs/components/**/testing/src/**`) — follow
+  [skyux-copilot-harnesses.instructions.md](./skyux-copilot-harnesses.instructions.md).
+
+Before writing, **open a sibling spec in the same library** to match
```
Blackbaud-ErikaMcVey: How confident are we that ALL of our specs currently follow these best practices and are good enough to be a blueprint? I wonder if it would be better to create a really clean sample to provide for these instructions. 

Are we saying copy a spec in the library because of the providers bit? Even then, I doubt every spec file in a library needs the exact same providers--might be better to provide it with a baseline, and then give instructions on how to determine what other providers are needed.

Edit: I see we do have a small example below, along with common providers. My concern with this is that once you tell copilot to look at that other file, it will be in it's context and it may reference it for more than just providers and end up copying anti-patterns. Getting copilot to _not_ reference existing code has already been a challenge for me _without_ ever explicitly telling it to copy something.

.github/instructions/component-unit-testing.instructions.md
```diff
@@ -0,0 +1,167 @@
...
+Write Karma + Jasmine unit tests for the components and directives that ship
+from a `@skyux/*` library (the source under
+`libs/components/<library>/src/lib/**`). SKY UX styles are loaded during these
+tests, so computed styles and accessibility are assertable.
+
+This guidance is for **library source specs**. It does not cover:
+
+- **Code-example specs** (`libs/components/code-examples/**`) — follow
+  [code-examples-unit-testing.instructions.md](./code-examples-unit-testing.instructions.md).
+- **Harness specs** (`libs/components/**/testing/src/**`) — follow
```
Blackbaud-ErikaMcVey, line 19: don't we have a few jest tests in manifest and packages? should they be added to this ignore list, or is the different file extension enough that it will ignore it? Not sure if the `applyTo` is interpreted as gospel or a suggestion 😂

.github/instructions/component-unit-testing.instructions.md
```diff
@@ -0,0 +1,167 @@
...
+  new `async` specs (and before loading a harness).
+- Use `fakeAsync` + `tick()`/`flush()` to control timer-based async such as
+  animations and debounces. Because this repo runs on Karma + Jasmine with
+  zone.js, keep using these; the "`fakeAsync` is discouraged" note on
+  angular.dev applies only to zoneless Vitest projects.
+
+## Assertions & matchers
+
+Import `expect`/`expectAsync` from `@skyux-sdk/testing` (this registers the SKY
+custom matchers); standard Jasmine matchers remain available.
```
Blackbaud-ErikaMcVey, line 122: How does it get context on the standard Jasmine matchers? Just by looking at other spec files, or the node module? I wonder if it would be more "efficient" if we gave it a list, or at least clear instructions on where to look so we don't have to maintain a list.

.github/instructions/component-unit-testing.instructions.md
```diff
@@ -0,0 +1,167 @@
...
+- `expect(el).toHaveCssClass('sky-...')` / `.not.toHaveCssClass(...)`
+- `expect(el).toBeVisible()` / `expect(el).toExist()`
+- `expect(el).toHaveText('...')` / `toHaveResourceText(...)` /
+  `toHaveLibResourceText(...)` for i18n strings
+- `await expectAsync(fixture.nativeElement).toBeAccessible()` — see below
+
+## Accessibility
+
+Accessibility coverage is expected. Add a `describe('a11y', ...)` block that
+asserts the rendered output is accessible in each meaningful state:
```
Blackbaud-ErikaMcVey: would it be worth expanding on "each meaningful state" a bit, either in the example below or just by elaborating that that means with all possible input combinations and code paths?

.github/instructions/scss-override-mixins.instructions.md
```diff
@@ -1,14 +1,15 @@
 ---
-applyTo: '**/*.scss'
+applyTo: 'libs/components/**/*.scss'
```
Blackbaud-ErikaMcVey: can be deleted as well

.github/instructions/component-harnesses.instructions.md
```diff
@@ -7,125 +7,273 @@ description: 'SKY UX Copilot instructions for generating harnesses for component
 
 ## Overview
 
-Create test harnesses that provide a stable API for interacting with SKY UX components in automated tests. Harnesses should expose the component's public functionality through user-centric methods.
+Create test harnesses that provide a stable, user-centric API for interacting
```
Blackbaud-ErikaMcVey, line 10: does this need an applyTo or description? Just so it doesn't think it should create a test harness for something outside of components?

.github/instructions/skyux-copilot-harnesses.instructions.md
```diff
@@ -7,125 +7,273 @@ description: 'SKY UX Copilot instructions for generating harnesses for component
...
 ## Overview
 
-Create test harnesses that provide a stable API for interacting with SKY UX components in automated tests. Harnesses should expose the component's public functionality through user-centric methods.
+Create test harnesses that provide a stable, user-centric API for interacting
+with SKY UX components in automated tests. A harness exposes a component's
+public behavior through async methods so consumers' tests do not depend on
+internal DOM structure.
 
-## Basic Structure
+Before writing a new harness, **open a sibling harness in the same library (or a
```
Blackbaud-ErikaMcVey: similar concern as above...I know that there were already some patterns we wanted to go back and change in test harnesses. If we don't want to create a whole dummy example, maybe just direct it to one that we think is really good already?

.github/skills/add-component-harness/SKILL.md
```diff
@@ -0,0 +1,93 @@
+---
```
Blackbaud-ErikaMcVey, line 1: I'm surprised by the split between the skill and the instructions file. Does that provide some desired functionality? Do we know if it ingests all instructions files for context for every prompt you give it, or if only ingests them if it thinks they're relevant based on the description?

It also just seems like it could be 2 places we have to maintain/keep info in sync

.github/skills/add-skyux-code-example/SKILL.md
```diff
@@ -0,0 +1,114 @@
+---
```
Blackbaud-ErikaMcVey, line 1: and then these next three are only skills, not a skill and instructions. Why the difference in approach?

.github/instructions/component-unit-testing.instructions.md
Blackbaud-SteveBrush: It's a good observation, but we're fine for now:
The `manifest` project uses the `.test.ts` extension, and its files sit directly under `src/`, not `src/lib/`.
The `packages` project does use `.spec.ts`, but every spec lives under `src/schematics/**`, never `src/lib/**`.

.github/instructions/component-unit-testing.instructions.md
Blackbaud-SteveBrush: The standard jasmine matchers are exposed on the global `jasmine` type, imported via the tsconfigs. I think we're okay on this point.

.github/instructions/component-unit-testing.instructions.md
Blackbaud-SteveBrush: Updated the wording to tell the agent a "meaningful state" is one that "changes the rendered DOM in a way a keyboard or screen-reader user would perceive — not every input permutation" to guard against an infinite number of input combinations.

.github/instructions/component-unit-testing.instructions.md
Blackbaud-SteveBrush: Good call-out! I baked-in the canonical example instead of asking the agent to reference potentially outdated patterns from external sources.

.github/instructions/skyux-copilot-harnesses.instructions.md
Blackbaud-SteveBrush: Yep; good call.

.github/skills/add-component-harness/SKILL.md
Blackbaud-SteveBrush: Good question. Instruction files apply automatically whenever the agent edits a file matching their applyTo glob (always-on guardrails per file type); skills are invoked only when the agent is asked to perform a specific task and orchestrate the workflow. To avoid drift, skills reference the instruction files for code-level conventions rather than duplicating them.

.github/skills/add-skyux-code-example/SKILL.md
Blackbaud-SteveBrush: The `add-skyux-component` and `add-skyux-library` skills are pure scaffolding orchestration. They don't produce a single coherent file type. The `add-skyux-code-example` skill partially has a companion (code-example-unit-testing) but it's only for unit tests; based on this feedback, I broadened the instructions file (renamed to `component-code-examples.instrunctions.md`) so it works like the other skills/instructions pairs.

Blackbaud-ErikaMcVey (approved):

PR #4541, opened 2026-07-24
feat(components/packages): add TypeScript 6 migration schematic

johnhwhite:
Adds an update-15 migration that prepares consumer workspaces for
TypeScript 6 by converting implicit-baseUrl imports to relative
imports, converting legacy `module Foo {}` namespaces, removing
`baseUrl` (while rebasing `paths`), and cleaning up deprecated and
default-matching tsconfig compiler options.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01PoLAoPsvX7BD1yenuTywyX

[AB#4062933](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/4062933)

## Summary by CodeRabbit

* **New Features**
  * Added an automated migration to prepare Angular workspaces for TypeScript 6.
  * Updates imports and legacy namespace syntax for TypeScript 6 compatibility.
  * Cleans up deprecated and default TypeScript configuration options.
  * Rebases path mappings and warns about potentially incompatible settings.

* **Tests**
  * Added comprehensive coverage for source transformations, configuration inheritance, warnings, and repeatable migration results.

### DIFF
--- a/libs/components/packages/src/schematics/migrations/update-15/address-typescript-6-changes/address-typescript-6-changes.ts
+++ b/libs/components/packages/src/schematics/migrations/update-15/address-typescript-6-changes/address-typescript-6-changes.ts
@@ -1,0 +1,20 @@
+import { Rule, SchematicContext, Tree } from '@angular-devkit/schematics';
+
+import { convertSourceFiles } from './lib/convert-source-files';
+import { buildTsconfigModel } from './lib/tsconfig-model';
+import { updateTsconfigFiles } from './lib/update-tsconfig-files';
+
+/**
+ * Prepares a workspace for TypeScript 6: converts implicit-`baseUrl` imports
+ * to relative imports and legacy `module Foo {}` namespaces before removing
+ * `baseUrl` (rebasing `paths`), and cleans up deprecated/default tsconfig
+ * compiler options.
+ */
+export default function addressTypescript6Changes(): Rule {
+  return (tree: Tree, context: SchematicContext): void => {
+    const model = buildTsconfigModel(tree);
+
+    convertSourceFiles(tree, model, context);
+    updateTsconfigFiles(tree, model, context);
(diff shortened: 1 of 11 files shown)


libs/components/packages/src/schematics/migrations/update-15/address-typescript-6-changes/address-typescript-6-changes.ts
```diff
@@ -0,0 +1,20 @@
...
+ * to relative imports and legacy `module Foo {}` namespaces before removing
+ * `baseUrl` (rebasing `paths`), and cleans up deprecated/default tsconfig
+ * compiler options.
+ */
+export default function addressTypescript6Changes(): Rule {
+  return (tree: Tree, context: SchematicContext): void => {
+    const model = buildTsconfigModel(tree);
+
+    convertSourceFiles(tree, model, context);
+    updateTsconfigFiles(tree, model, context);
```
Blackbaud-SteveBrush, line 18: I'd prefer we move this schematic to our internal monorepo. The schematic contains no SKY UX–specific logic; it rewrites imports and tsconfig options, which is a workspace-level concern that even Angular declines to migrate automatically, so I'd lean toward our component library not updating the tsconfigs, either. 

Running it in public consumers' workspaces risks conflicts with their own TS 6 tooling and produces breakage we can't reproduce or support; internally, we do control the workspace shape so we can make assumptions about which fixes to make. 

It also belongs next to the existing `fix-tsconfigs` schematic, which owns tsconfig defaults; co-locating them prevents the two from drifting into conflicting changes (it might be worth taking another look at the `fix-tsconfigs` schematic to make sure the defaults match what TypeScript 6 and Angular expect).

PR #4610, opened 2026-08-11
feat(components/i18n)!: make getString/getStrings update on locale changes

johnhwhite:
Breaking change: getString/getStrings observables no longer complete

What changed: SkyAppResourcesService getString()/getStrings() and SkyLibResourcesService getString()/getStrings() now use combineLatest instead of forkJoin internally, so they emit a new value whenever the active locale changes instead of emitting once and completing.

Impact on your code:
- Async pipe / ongoing subscriptions: benefit automatically — resource strings now update live when the locale changes, no action needed.
- firstValueFrom() / take(1) / .pipe(first()): still work correctly (they only care about the first value), but be aware the underlying observable itself never completes on its own now.
- Code that relies on completion — e.g. forkJoin([resourcesService.getString(...), other$]), toPromise() (deprecated but if still used, it will now hang forever waiting for completion), or manual subscribe({ complete: () => ... }) handlers — will break or hang, since the observable no longer completes. Replace those usages with take(1) or firstValueFrom if you only need a single value.
- Memory/subscription leaks: since the observable stays open, any manual .subscribe() without an async pipe or explicit takeUntilDestroyed()/unsubscribe() will now leak, whereas previously it self-cleaned on completion.

[AB#4082284](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/4082284)

## Summary by CodeRabbit

* **New Features**
  * Added compatibility services for existing internationalization integrations.
  * Added migration support for applications using updated internationalization services.
  * Improved fallback selection for language variants and English (US).

* **Bug Fixes**
  * Localized messages and resource dictionaries now refresh when the selected locale changes.
  * Empty resource requests return immediately with an empty result.
  * Improved localized modal, text expansion, and alert updates.
  * Prevented duplicate confirmation prompts during locale changes.


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4610/storybook/)

Component Storybooks:

- [a11y-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/a11y-storybook/)
- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/avatar-storybook/)
- [charts-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/charts-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/colorpicker-storybook/)
- [core-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/core-storybook/)
- [data-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/data-grid-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/errors-storybook/)
- [filter-bar-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/filter-bar-storybook/)
- [flyout-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/flyout-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/forms-storybook/)
- [help-inline-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/help-inline-storybook/)
- [icon-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/icon-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/lookup-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/modals-storybook/)
- [navbar-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/navbar-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/phone-field-storybook/)
- [popovers-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/popovers-storybook/)
- [progress-indicator-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/progress-indicator-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/split-view-storybook/)
- [storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/theme-storybook/)
- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/tiles-storybook/)
- [toast-storybook](https://blackbaud.github.io/skyux-pr-preview/4610/storybooks/toast-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4610/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4610/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4610/playground/)

libs/components/packages/src/schematics/migrations/update-15/legacy-i18n-services/legacy-i18n-services.schematic.ts
```diff
@@ -0,0 +1,71 @@
...
+
+      const recorder = tree.beginUpdate(filePath);
+
+      swapImportedClass(recorder, filePath, parseSourceFile(tree, filePath), [
+        {
+          classNames: CLASS_NAMES,
+          moduleName: MODULE_NAME,
+          filter: (node): boolean => !isStaticAddResourcesReference(node),
+        },
+      ]);
```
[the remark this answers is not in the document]
johnhwhite: @coderabbitai libs/components/packages/src/schematics/utility/typescript/swap-imported-class.ts has been updated to address this issue.

libs/components/layout/src/lib/modules/text-expand/text-expand.component.ts
```diff
@@ -208,25 +212,33 @@ export class SkyTextExpandComponent implements AfterContentInit {
   }
 
   public ngAfterContentInit(): void {
-    observableForkJoin([
+    combineLatest([
```
Blackbaud-SteveBrush, line 215: Good, here. The repeater item component still uses `observableForkJoin`.

libs/components/i18n/src/lib/modules/i18n/resources.service.ts
```diff
@@ -107,7 +107,7 @@ export class SkyAppResourcesService {
       }
     }
 
-    return forkJoin(resources$) as Observable<{ [K in keyof T]: string }>;
+    return combineLatest(resources$) as Observable<{ [K in keyof T]: string }>;
```
Blackbaud-SteveBrush: On line ~94, `getStrings` is calling `getString` for every key; each key opens its own locale subscription and one locale change produces any number of emits with partial dictionaries. What you're doing in libraries seems to work, but we might need to migrate the same single-subscription rewrite, here, too.

libs/components/i18n/src/lib/modules/i18n/get-string-for-locale.ts
```diff
@@ -9,22 +9,22 @@ export function getStringForLocale(
...
-  }
+  const normalizeLocale = (locale: string): string =>
+    locale.toUpperCase().replace('_', '-');
 
-  // Attempt to locate default resources.
-  values = getResourcesForLocale(defaultLocale);
+  const localeFallback = [
+    ...new Set([
+      normalizeLocale(preferredLocale),
+      normalizeLocale(preferredLocale.substring(0, 2)),
```
Blackbaud-SteveBrush: Adding support for optional country codes is a separate feature and introduces new behavior. It's also untested in this branch; can we drop it from this PR and create a new one so the behavior change makes it into the changelog and has appropriate tests?

libs/components/i18n/src/lib/modules/i18n/lib-resources.service.spec.ts
```diff
@@ -302,5 +331,63 @@ describe('Library resources service', () => {
         done();
       });
     });
+
+    it('re-emits when locale changes for getString', (done) => {
```
Blackbaud-SteveBrush, line 335: Can we add a test showing that the `getString` method never completes automatically? Would demonstrate one of the new behaviors of the service method.

libs/components/indicators/src/lib/modules/alert/alert.component.ts
```diff
@@ -29,7 +29,6 @@ const ALERT_TYPE_DEFAULT = 'warning';
   ],
   templateUrl: './alert.component.html',
   hostDirectives: [SkyThemeComponentClassDirective],
-  changeDetection: ChangeDetectionStrategy.Eager,
```
Blackbaud-SteveBrush, line 32: `ChangeDetectionStrategy.Eager` was removed from alert, label, and text-expand. Is dropping it required for the long-lived subscriptions to render correctly? If so, let's update the reason in the breaking changes description so our consumers will know how to address any issues that come up when migrating.

libs/components/i18n/src/lib/modules/i18n/resources-legacy.service.ts
```diff
@@ -0,0 +1,98 @@
...
+  /* eslint-disable @angular-eslint/prefer-inject -- constructor injection is required to maintain the public API for consumers who may instantiate this service directly (e.g. `new SkyAppResourcesService(...)`) */
+  constructor(
+    http: HttpClient,
+    @Optional()
+    @Inject(forwardRef(() => SkyAppAssetsService))
+    assets: SkyAppAssetsService,
+    @Optional() localeProvider: SkyAppLocaleProvider,
+    @Optional() resourceNameProvider: SkyAppResourceNameProvider,
+  ) {
+    this.#resourcesService = new SkyAppResourcesService(
```
Blackbaud-SteveBrush: Because we're introducing a new service with reactive functionality, I don't think we can do the same thing we did with the modal legacy services (legacy extends the logic of the new). By extending the logic of the new service, there's a high risk that changes will negatively affect existing legacy consumers. There are a few in this PR:

- `getStrings({})`: the empty-dictionary guard is still in the reactive service, so it emits `{}`. The old `forkJoin({})` completed _without_ emitting. `take(1)` doesn't undo that. Neither legacy spec covers the empty case.
- `take(1)` emits the first combination; `forkJoin` emitted the last value of each source. These differ whenever a locale provider or `SkyAppResourceNameProvider` emits more than once before completing. 

The biggest risk is that every future edit to `SkyLibResourcesService` becomes a potential silent edit to `SkyLibResourcesLegacyService` (and the app equivalents), and nothing in the type system or the test suite will tell us.

I think our best bet is to keep the legacy services's logic as-is (just throw "Legacy" on the service name and add some `@deprecated` comments) so that we're free to modify the new services as we see fit, without risk to legacy consumers.

libs/components/layout/src/lib/modules/text-expand/text-expand.component.ts
johnhwhite: Updated the repeater item component.

libs/components/i18n/src/lib/modules/i18n/resources.service.ts
johnhwhite: Updated to work similarly to libraries.

libs/components/i18n/src/lib/modules/i18n/get-string-for-locale.ts
johnhwhite: Split out to AB#4092222

libs/components/indicators/src/lib/modules/alert/alert.component.ts
johnhwhite: In updating the resource calls, I also added the change detection necessary to remove `ChangeDetectionStrategy.Eager`, but I've put that back so we can confirm it separately.

libs/components/i18n/src/lib/modules/i18n/resources-legacy.service.ts
johnhwhite: The legacy services are now copies of the old implementations and not tied to the new ones.

libs/components/i18n/src/lib/modules/i18n/lib-resources.service.spec.ts
johnhwhite: The `should re-emit when locale changes` test demonstrates that, right?

libs/components/i18n/src/lib/modules/i18n/lib-resources-legacy.service.ts
```diff
@@ -0,0 +1,167 @@
...
+    this.#providers = providers;
+    this.#resourceNameProvider = resourceNameProvider;
+  }
+
+  /**
+   * Adds locale resources to be used by library components.
+   *
+   * @deprecated This method is deprecated. Use `SkyLibResourcesService.addResources()` instead.
+   */
+  public static addResources(
```
Blackbaud-SteveBrush, line 59: It's good that we're duplicating the retrieval logic for the resources, but we can't have duplicate static resources for both the legacy and new services. We only call `SkyLibResourcesService.addResources(RESOURCES)` with the new service, so if a consumer imports `SkyLibResourcesLegacyService.getString()` it'll return the key, not the resolved value.

Potential fix: the legacy service's `addResources` and `getLibStringForLocale` methods should target the `SkyLibResourcesService`'s static store; we can then remove the legacy store altogether.

libs/components/i18n/src/lib/modules/i18n/resources.service.ts
```diff
@@ -66,10 +67,13 @@ export class SkyAppResourcesService {
...
-  public getString(name: string, ...args: any[]): Observable<string> {
-    const localeObs: Observable<SkyAppLocaleInfo> =
-      this.#localeProvider.getLocaleInfo();
-    return this.#getStringForLocaleInfoObservable(localeObs, name, ...args);
+  public getString(name: string, ...args: unknown[]): Observable<string> {
+    return this.#localeProvider.getLocaleInfo().pipe(
+      switchMap((localeInfo) =>
+        this.getStringForLocale(localeInfo, name, ...args),
+      ),
+      catchError(() => observableOf(name)),
```
Blackbaud-SteveBrush, line 85: In an error, `observableOf` completes the stream and future locale changes aren't piped to the subscriber, killing reactivity.

libs/components/i18n/src/lib/modules/i18n/resources.service.ts
```diff
@@ -94,20 +98,38 @@ export class SkyAppResourcesService {
...
-        resources$[objKey] = this.getString(key, ...templateItems);
-      }
+    if (entries.length === 0) {
+      return EMPTY;
     }
 
-    return forkJoin(resources$) as Observable<{ [K in keyof T]: string }>;
+    const resourcesObs = this.#localeProvider
+      .getLocaleInfo()
+      .pipe(switchMap((locale) => this.#getLocaleResourceObservable(locale)));
```
Blackbaud-SteveBrush: Previously, both `getString` and `getStrings` handled errors via `#getStringForLocaleInfoObservable`, which included a comment explaining it exists to stop impure pipes from re-firing requests indefinitely. Errors now propagate to `getStrings` consumers but are swallowed for `getString`. Can we re-implement error handling in a similar way that it was done before?

johnhwhite: Added the error handling back.

libs/components/i18n/src/lib/modules/i18n/lib-resources-legacy.service.ts
johnhwhite: Added the shared registry we'd discussed offline.

libs/components/i18n/src/lib/modules/i18n/resources.service.ts
johnhwhite: If an error fires at this point, I'm not sure what else we could do. To recover here we'd need to add [`retry()`](https://rxjs.dev/api/index/function/retry). Since `getStringForLocale` handles its own errors, this `catchError` really only covers changing locale. I'm inclined to either leave it or remove it and let the error bubble.

libs/components/i18n/src/lib/modules/i18n/resources.service.ts
```diff
@@ -94,20 +98,38 @@ export class SkyAppResourcesService {
...
+
+    return combineLatest([mappedNames$, resourcesObs]).pipe(
+      map(
+        ([mappedNames, resources]): { [K in keyof T]: string } =>
+          Object.fromEntries(
+            entries.map(({ objKey, args }, i) => [
+              objKey,
+              this.#getResourceString(
+                resources,
+                objKey,
```
Blackbaud-SteveBrush: Shouldn't we pass `name` (instead of `objKey`)? In the legacy service, if a resource isn't found, it returns the resource key, but this will return the dictionary key (e.g., `myResourceKey` vs `my_resource_key`). The unit test fails to throw on this difference because the dictionary key and the resource key happen to match: `{ hi: 'hi' }`.

johnhwhite: Good call. Updated.

Blackbaud-SteveBrush (approved):

PR #4615, opened 2026-08-12
feat: add locale-aware AG Grid text

johnhwhite:
Adds Spanish and French AG Grid locale text, and makes `SkyDataGrid` react to runtime locale changes.

Tracked by [AB#4082332](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/4082332).

## What changed

**`@skyux/ag-grid`** — `SkyAgGridService` now selects AG Grid's stock translations based on the language subtag of the locale reported by `SkyAppLocaleProvider`, layered underneath the SKY UX resource strings so SKY's intentional rewording still wins:

```ts
const agGridLocaleTextByLanguage: Record<string, LocaleText> = {
  es: AG_GRID_LOCALE_ES,
  fr: AG_GRID_LOCALE_FR,
};
```

Matching is on the language subtag only, so `es`, `es-MX`, and `ES-mx` all resolve to Spanish. There is no English file — AG Grid's built-in text is already English, so English locales supply no translations at all and the ~36 KB of English strings stay out of consumers' bundles.

The locale files are copied verbatim from AG Grid Community (MIT). `libs/components/ag-grid/src/lib/modules/shared/README.md` records the provenance and how to add a locale. Spellchecking is disabled for those files specifically — vendored translations produced ~2,500 unknown-word warnings that buried real ones.

**`@skyux/data-grid`** — `SkyDataGrid` destroys and recreates its grid when the locale changes. AG Grid reads `localeText` only at grid creation, so this is the only way to pick up a new locale. The grid is blanked for a tick and the wait indicator covers the gap; data, sort, and selection are all restored afterward. Brief flicker is expected and unavoidable.

## Bug fixed along the way

`SkyDataGrid` registers a curated AG Grid module list rather than `AllCommunityModule`, and `LocaleModule` was missing from it. AG Grid was logging `error #200 — Unable to use localeText as LocaleModule is not registered` and ignoring the option entirely.

This is pre-existing and predates the locale work: `noRowsToShow`/`noMatchingRows` never reached `SkyDataGrid` either, so its empty grid has been showing AG Grid's `"No Rows To Show"` instead of SKY UX's `"No data available"`.

The regression test asserts the **rendered** overlay text rather than the `localeText` option, because the option is set correctly either way — AG Grid just ignores it. Verified with a negative control: commenting out `LocaleModule` fails exactly that one test.

## Breaking change

Consumers using `SkyAgGridService` grid options must add their own locale handling. AG Grid reads `localeText` only when a grid is created, so an application whose locale can change at runtime has to call `getGridOptions()` again and destroy and recreate the grid for the new locale text to take effect. `SkyDataGrid` does this for you; direct `SkyAgGridService` consumers do not get it automatically.

## Testing

- `@skyux/ag-grid` — 446 tests, 100% coverage. Locale branching for en/es/fr, subtag matching across region and casing, unsupported locales, an unresolved locale provider, switching locales across successive `getGridOptions()` calls, and both SKY-resource and consumer override precedence. A separate spec checks key parity between the translations so a dropped key can't silently fall back to English.
- `@skyux/data-grid` — 70 tests, 100% coverage. Initial locale, destroy/recreate with new text, DOM removal during the gap, SKY resource strings surviving, data/sort/selection restoration, no-recreate on an unchanged re-emit, and no-recreate for a locale that resolves before the grid is built.
- Lint clean (0 errors) and both libraries build.

## Follow-up worth tracking separately

The two keys SKY UX overrides (`noRowsToShow`, `noMatchingRows`) still render in English for es/fr. `resources_fr_CA.json` lacks both keys and there is no `resources_es_ES.json`, so the untranslated English fallback overwrites AG Grid's perfectly good `'No hay filas para mostrar'` / `'Aucune ligne à afficher'`. That's a translation-pipeline gap rather than a code bug, so no translations were hand-written here.

## Summary by CodeRabbit

* **New Features**
  * Added Spanish and French localization support for AG Grid and data grids.
  * Grid interfaces now update when the application language changes, including regional and case variations.
  * Localized filters, menus, charts, accessibility labels, validation messages, and empty states are supported.
  * SKY UX and consumer-provided translations take precedence over stock translations.
* **Bug Fixes**
  * Preserved grid data, sorting, and selection when changing languages.
* **Documentation**
  * Added guidance for available locales and introducing additional translations.


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4615/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4615/storybooks/ag-grid-storybook/)
- [data-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4615/storybooks/data-grid-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4615/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4615/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4615/playground/)

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid.service.ts
```diff
@@ -141,6 +149,16 @@ const skyAgGridLocaleTextDefaults = {
   noMatchingRows: 'No matching rows',
 } satisfies LocaleText;
 
+/**
+ * AG Grid's stock translations, keyed by the language subtag of the app's
+ * locale. Locales without an entry supply no translations, leaving AG Grid to
+ * use its own built-in English text.
+ */
+const agGridLocaleTextByLanguage: Record<string, LocaleText> = {
```
Blackbaud-SteveBrush, line 157: We talked offline about setting this up so that locale strings are loaded on-demand at runtime, instead of loading all translations into the default bundle at build time. This registry forces all consumers to pay for every locale; English-only applications will load all strings for all locales. Could we modify the registry to use loaders for each resource group, and load them when `SkyAppLocaleProvider` emits?
```
const localeTextLoaders: Record<string, () => Promise<LocaleText>> = {
  es: () => import('./ag-grid-locale-es-ES').then((m) => m.AG_GRID_LOCALE_ES),
  fr: () => import('./ag-grid-locale-fr-FR').then((m) => m.AG_GRID_LOCALE_FR),
};
```
That way, these dynamically imported chunks exist in the npm package, but only one chunk is loaded when needed.

Blackbaud-SteveBrush (commented):
Are there any breaking changes in this PR? I like that we're keeping the change in v15, but we probably don't need the `feat!` exclamation mark.

libs/components/ag-grid/src/lib/modules/ag-grid/ag-grid.service.ts
johnhwhite: I don't think this is feasible because it's called from a synchronous context. We don't do this for other library resources.

johnhwhite: I'd thought breaking because of the need for consumers to respond to locale changes; `<sky-ag-grid-wrapper>` cannot tear down and rebuild the `<ag-grid-angular>` instance because it belongs to the parent view. It's additive though, not breaking, so I've removed the `!`.

Blackbaud-SteveBrush: I asked Claude to implement the loader async idea and it worked, but it added a ton of complexity. Seeing the implementation convinced me it's not worth doing for only 1-2 translations. If we end up supporting 4+ we may need to come back to this and reevaluate. Let's move forward with what you have.

Blackbaud-SteveBrush (approved):

PR #4664, opened 2026-08-20
feat: add `@skyux-sdk/vitest` package

Blackbaud-SteveBrush:
[AB#3987231](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3987231)

## Summary by CodeRabbit

* **New Features**
  * Added a Vitest testing SDK with accessibility, visibility, DOM, styling, text, and localized-resource matchers.
  * Added configurable accessibility analysis with clear violation reporting.
  * Added helpers for application and library resource testing, including template matching.
  * Added shared testing setup and resource-provider utilities.
* **Bug Fixes**
  * Improved matcher validation and diagnostics for missing elements and mismatched styles or text.
  * Improved configurable visibility and accessibility checks.
* **Documentation**
  * Added setup, build, and test instructions for the Vitest SDK.


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4664/storybook/)

Component Storybooks:

- [a11y-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/a11y-storybook/)
- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/action-bars-storybook/)
- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/angular-tree-component-storybook/)
- [avatar-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/avatar-storybook/)
- [charts-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/charts-storybook/)
- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/colorpicker-storybook/)
- [core-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/core-storybook/)
- [data-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/data-grid-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/data-manager-storybook/)
- [datetime-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/datetime-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/errors-storybook/)
- [filter-bar-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/filter-bar-storybook/)
- [flyout-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/flyout-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/forms-storybook/)
- [help-inline-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/help-inline-storybook/)
- [icon-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/icon-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/indicators-storybook/)
- [inline-form-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/inline-form-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/lookup-storybook/)
- [modals-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/modals-storybook/)
- [navbar-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/navbar-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/phone-field-storybook/)
- [popovers-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/popovers-storybook/)
- [progress-indicator-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/progress-indicator-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/split-view-storybook/)
- [storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/theme-storybook/)
- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/tiles-storybook/)
- [toast-storybook](https://blackbaud.github.io/skyux-pr-preview/4664/storybooks/toast-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4664/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4664/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4664/playground/)

libs/sdk/vitest/src/lib/matchers/element-is-visible.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export function elementIsVisible(
+  el: Element | null | undefined,
+  options?: ElementIsVisibleOptions,
+): ExpectationResult {
+  const settings = { ...DEFAULTS, ...options };
+
+  let pass = !!el;
+
+  if (el) {
```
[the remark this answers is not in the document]
Blackbaud-SteveBrush: This would cause API/behavioral drift between this matcher and its equivalent in `@skyux-sdk/testing`. We want migrations to vitest to be smooth; we can address this issue in a future PR.

libs/sdk/vitest/src/lib/utility/i18n-utils.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  while (currentToken !== undefined && matches) {
+    const tokenPosition = sample.indexOf(currentToken, lastPosition);
+
+    matches = tokenPosition >= lastPosition;
+    lastPosition = tokenPosition + currentToken.length;
+    currentToken = templateTokens.pop();
+  }
+
+  return matches;
```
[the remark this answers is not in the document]
Blackbaud-SteveBrush: Skipping. Loose matching is the named, tested contract. matchers.spec.ts:776 pins it directly. Anchoring the match would invert a behavior that a published test asserts by name. This isn't a latent bug — it's the specification.

libs/sdk/vitest/package.json
```diff
@@ -0,0 +1,57 @@
...
+  "exports": {
+    ".": {
+      "types": "./src/index.d.ts",
+      "default": "./src/index.js"
+    },
+    "./globals": {
+      "types": "./globals.d.ts"
+    },
+    "./matchers-setup.mjs": {
+      "types": "./matchers-setup.d.ts",
```
johnhwhite: **exports types path is `.d.ts` but tsc emits `.d.mts`**

`exports[\"./matchers-setup.mjs\"].types` points at `./matchers-setup.d.ts`, which tsc never emits for a `.mts` input — it emits `.d.mts`.

Verified with a real `npx nx build vitest`: `dist/libs/sdk/vitest/` contains `matchers-setup.d.mts` but no `matchers-setup.d.ts`, and the shipped dist `package.json` carries the broken path verbatim. A consumer on `moduleResolution` node16/nodenext/bundler importing `@skyux-sdk/vitest/matchers-setup.mjs` fails the `types` condition, falls through to a bare `.mjs` with no declaration, and gets TS7016 under strict mode.

Fix: `./matchers-setup.d.mts`.

libs/sdk/vitest/src/lib/utility/a11y-analyzer.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    // Disable autocomplete-valid
+    // Chrome browsers ignore autocomplete="off", which forces us to use non-standard values
+    // to disable the browser's native autofill.
+    // https://bugs.chromium.org/p/chromium/issues/detail?id=468153#c164
+    defaults.rules['autocomplete-valid'] = { enabled: false };
+
+    return new Promise((resolve, reject) => {
+      const callback: axe.RunCallback = (error, results) => {
+        if (error?.message) {
```
johnhwhite, line 114: **error guard misses string errors, then dereferences undefined `results`**

The `if (error?.message)` guard misses axe errors reported as a bare string, then dereferences an undefined `results`.

axe-core 4.12 (`axe.js:29885`) calls `callback(err.message)` — a string, with no `results` argument. So `error?.message` is `undefined`, the guard falls through, and `results.violations` throws on `undefined`. Trigger: two overlapping `toBeAccessible()` assertions (a missing `await`, or `Promise.all`) makes axe raise "Axe is already running"; the matcher then reports `pass:false` with `Cannot read properties of undefined (reading 'violations')`, swallowing the real cause. Via `errorRunRules` the callback fires twice and an empty-message error throws asynchronously outside the promise executor, so the promise never settles and the test times out.

Six specs in `a11y-analyzer.spec.ts` pass `new Error()` to mean "no error", encoding this buggy guard as the contract. Correct shape: `if (error)` plus a results presence check (axe passes `null` on success, `axe.js:29866`).

libs/sdk/vitest/src/testing/provide-resources.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export type GetStringFn = (
+  resourceKey: string,
+  ...resourceArgs: unknown[]
+) => Observable<string>;
+
+export function provideAppResources(
+  getString: GetStringFn,
+): SkyAppResourcesService {
+  TestBed.configureTestingModule({
```
johnhwhite, line 13: **`provideAppResources` and `provideLibResources` can't both be used in one test**

Each helper ends with `TestBed.inject(...)`, which instantiates the test module; a following `configureTestingModule` then hits `assertNotInstantiated` (`@angular/core/fesm2022/testing.mjs:1424`). Calling `provideAppResources(fn); provideLibResources(fn);` in one test or `beforeEach` throws "Cannot configure the test module when the test module has already been instantiated." A spec covering `toHaveResourceText` together with `toHaveLibResourceText` cannot be set up this way. The integration spec only ever calls one per suite, so it doesn't catch this.

tsconfig.base.json
```diff
@@ -40,6 +40,7 @@
         "libs/sdk/testing/vitest/src/index.ts"
```
johnhwhite: **dead path mapping still points at deleted `testing/vitest` source**

`"@skyux-sdk/testing/vitest/matchers-setup.js"` still points at the deleted `libs/sdk/testing/vitest/src/index.ts`. That directory no longer exists on this branch (only the new `@skyux-sdk/vitest` mapping was added below). TypeScript doesn't eagerly validate `paths` targets, so this is silent rather than fatal — but anything resolving through it gets TS2307, and it's misleading dead config that should have been removed with the move.

Compounding effect: this stale mapping is what lets the new integration spec's `typeof import('@skyux-sdk/vitest/matchers-setup.mjs')` probe resolve at all in-repo (via the catch-all `"*"` mapping falling through to the `.mts` source) — which is why that probe can't catch the `.d.ts`/`.d.mts` mismatch flagged on `libs/sdk/vitest/package.json`.

libs/sdk/vitest/project.json
```diff
@@ -0,0 +1,43 @@
...
+        "main": "libs/sdk/vitest/src/index.ts",
+        "assets": ["libs/sdk/vitest/*.md", "libs/sdk/vitest/globals.d.ts"]
+      }
+    },
+    "lint": {
+      "executor": "@nx/eslint:lint",
+      "options": {
+        "lintFilePatterns": [
+          "{projectRoot}/src/**/*.ts",
+          "{projectRoot}/matchers-setup.ts"
```
johnhwhite: **lint pattern says `.ts` but the entry point is `.mts`, so lint silently skips it**

`lintFilePatterns` lists `{projectRoot}/matchers-setup.ts` but the file is `matchers-setup.mts`, so the published entry point is never linted.

`npx eslint libs/sdk/vitest/matchers-setup.ts` → "No files matching the pattern were found." Nx passes `--no-error-on-unmatched-pattern`, so `nx lint vitest` reports "All files pass linting" while silently skipping `matchers-setup.mts` and the root-level `globals.d.ts` entirely. Evidence lint coverage was expected: `globals.d.ts:7` carries an `eslint-disable-next-line` for rules that never actually run on it. Every sibling project (`libs/sdk/tools`, `skyux-stylelint`, `stylelint-config-skyux`) uses a whole-project glob `{projectRoot}/**/*.ts` instead.

Note even after fixing the extension, `eslint-base.config.js:32` and `eslint-overrides.config.js:5` glob only `**/*.ts`, so `.mts` still gets no repo rules applied.

johnhwhite (commented):
Automated review — 8 findings verified with a real `nx build vitest` (no source modified). Six are posted as inline comments below. Two more couldn't be anchored inline because the files aren't touched by this diff:

**`libs/sdk/testing/private/src/a11y/a11y-analyzer.ts` (~line 130) — rules-merge fix landed only in the new fork, not here**
The new `@skyux-sdk/vitest` copy of this file uses `rules: { ...defaults.rules, ...config?.rules }`, but this still-live jasmine analyzer keeps `{ ...defaults, ...config }`. Any jasmine test calling `toBeAccessible` with custom rules replaces the whole `rules` object and silently discards the `autocomplete-valid: { enabled: false }` default set 20 lines above — whose comment explains it must stay off because Chrome ignores `autocomplete="off"`. Those tests spuriously fail on `autocomplete-valid`. Two matchers with the same public name now behave differently depending on the runner, with no shared source and no test asserting they agree.

**`libs/sdk/testing/README.md` (~line 19) — a published entry point was removed with no migration, and the old README still documents it**
This branch deletes `exports["./vitest/matchers-setup.js"]` from `libs/sdk/testing/package.json`. That subpath shipped in 14.10.0 through 14.14.0. Consumers who followed this README get `ERR_PACKAGE_PATH_NOT_EXPORTED` at test startup after upgrading. There's no `ng-update` migration in `libs/components/packages/migrations.json` to rewrite `setupFiles` to `@skyux-sdk/vitest/matchers-setup.mjs` (note also `.mjs`, not `.js`), and `libs/sdk/testing-schematics` has no migrations directory at all. Either add a migration or an explicit breaking-change note, and move the Vitest setup section of this README to the new package.

---
Also worth a mention (not part of the 8, and not blocking): `exports["./globals"]` in `libs/sdk/vitest/package.json` declares only `types` with no `default`, so a runtime `import '@skyux-sdk/vitest/globals'` throws `ERR_PACKAGE_PATH_NOT_EXPORTED`.

libs/sdk/vitest/src/lib/utility/a11y-analyzer.ts
Blackbaud-SteveBrush: Existing issue; filed a bug to address both matchers: https://dev.azure.com/blackbaud/Products/_workitems/edit/4097065

libs/sdk/vitest/src/testing/provide-resources.ts
Blackbaud-SteveBrush: These are internal-only test utilities and won't be used by our consumers. If you think there's another underlying issue, we can file a bug to address it in both the jasmine and vitest matchers since the logic is identical.

Blackbaud-SteveBrush:
> **`libs/sdk/testing/private/src/a11y/a11y-analyzer.ts` (~line 130) — rules-merge fix landed only in the new fork, not here** The new `@skyux-sdk/vitest` copy of this file uses `rules: { ...defaults.rules, ...config?.rules }`, but this still-live jasmine analyzer keeps `{ ...defaults, ...config }`. Any jasmine test calling `toBeAccessible` with custom rules replaces the whole `rules` object and silently discards the `autocomplete-valid: { enabled: false }` default set 20 lines above — whose comment explains it must stay off because Chrome ignores `autocomplete="off"`. Those tests spuriously fail on `autocomplete-valid`. Two matchers with the same public name now behave differently depending on the runner, with no shared source and no test asserting they agree.

Created a bug for the jasmine matchers: https://dev.azure.com/blackbaud/Products/_workitems/edit/4097056

libs/sdk/vitest/src/testing/provide-resources.ts
johnhwhite: Ok, right, they're test fixtures but they're not using that convention.

johnhwhite (approved):

PR #4674, opened 2026-08-21
feat!: hide fluid grid's outer margin by default and add inset input

Blackbaud-TrevorBurch:
## Summary

Implements [AB#3647919](https://dev.azure.com/blackbaud/Products/_workitems/edit/3647919): most consumers of `sky-fluid-grid` set `disableMargin="true"`, so the outer margin is now hidden by default instead of shown.

- Added a new `inset` input (default `false`) that shows the fluid grid's outer left/right margin when set to `true`.
- Deprecated `disableMargin`. It still works when explicitly set (logs a `SkyLogService` deprecation warning) and takes precedence over `inset` so existing behavior is preserved until it's removed in a future major version.
- Added an `ng update` migration schematic (`fluid-grid-inset`, registered under SKY UX 15) that rewrites `disableMargin` usage on `sky-fluid-grid` elements:
  - `[disableMargin]="true"` / `disableMargin="true"` / bare `disableMargin` → removed (hiding the margin is now the default).
  - `[disableMargin]="false"` / `disableMargin="false"` → `[inset]="true"` / `inset="true"`.
  - A bound, non-literal expression → inverted and renamed, e.g. `[disableMargin]="expr"` → `[inset]="!(expr)"`.
  - Elements that don't set `disableMargin` at all are left untouched; the schematic logs a warning telling consumers to manually verify those spots, since their margin will now be hidden by default.
- Updated internal demos, playground pages, and code examples to match (removed now-redundant `disableMargin="true"`, switched the "Fluid grid with basic setup" reference example to demonstrate `inset`).

## Migration guide (breaking change)

Starting in this version, `<sky-fluid-grid>` hides its outer margin unless `inset` is set to `true`. Run `ng update` to migrate automatically, or by hand:

| Before | After |
| --- | --- |
| `[disableMargin]="true"` | *(remove — now the default)* |
| `[disableMargin]="false"` | `[inset]="true"` |
| `[disableMargin]="expr"` | `[inset]="!(expr)"` |
| *(not set)* | Verify intent — margin is now hidden by default; add `[inset]="true"` to keep it |

## Test plan

- [x] `nx test packages` — new `fluid-grid-inset` schematic spec covers static/bound `true`/`false`, bare attribute, dynamic expressions, multiple elements, inline templates, and no-op cases (100% coverage on the new file).
- [x] Updated `SkyFluidGridComponent` unit tests for the new default, `inset` input, `disableMargin` precedence, and the deprecation warning.
- [x] Updated the `SkyFluidGridHarness` and code-example specs for the new default.
- [x] `nx format:write`, `eslint` on all changed files.
- [ ] CI Karma suites for `layout`, `layout-testing`, and `code-examples` (headless Chrome runs were too slow to complete locally in this environment).

## Summary by CodeRabbit

* **New Features**
  * Added the `inset` option to control fluid-grid spacing.
  * Added an automatic migration for legacy `disableMargin` configurations.
* **Bug Fixes**
  * Preserved existing spacing behavior while providing deprecation guidance.
  * Updated examples, layouts, and demonstrations to use the new spacing configuration.
* **Tests**
  * Expanded coverage for default margins, inset behavior, legacy compatibility, deprecation warnings, and migration scenarios.


libs/components/packages/src/schematics/migrations/update-15/fluid-grid-inset/fluid-grid-inset.schematic.ts
```diff
@@ -0,0 +1,209 @@
...
+    const value = staticAttr.value.trim().toLowerCase();
+    if (value === 'false') {
+      replaceAttribute(node, STATIC_ATTRIBUTE, 'inset="true"', offset, recorder);
+    } else {
+      // An empty value (bare attribute) or "true" both disable the margin,
+      // which is now the default, so the attribute can be removed.
+      removeAttribute(node, STATIC_ATTRIBUTE, content, offset, recorder);
+    }
+    return true;
+  }
```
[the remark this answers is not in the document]
Blackbaud-TrevorBurch: Fixed: removed the early return so an element with both the static and bound attribute now has both migrated (added a test for this case).

libs/components/packages/src/schematics/migrations/update-15/fluid-grid-inset/fluid-grid-inset.schematic.ts
```diff
@@ -0,0 +1,209 @@
...
+              template.start,
+              recorder,
+            );
+          });
+        }
+
+        tree.commitUpdate(recorder);
+      }
+    });
+  });
```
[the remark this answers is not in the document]
Blackbaud-TrevorBurch: Fixed: files visited through overlapping project roots are now deduplicated via a `visitedFiles` set, so the warning count isn't inflated (added a test for this case).

libs/components/packages/src/schematics/migrations/update-15/fluid-grid-inset/fluid-grid-inset.schematic.ts
```diff
@@ -0,0 +1,229 @@
...
+  node: ElementWithLocation,
+  content: string,
+  offset: number,
+  recorder: UpdateRecorder,
+): boolean {
+  const staticAttr = node.attrs.find((attr) => attr.name === STATIC_ATTRIBUTE);
+  const boundAttr = node.attrs.find((attr) => attr.name === BOUND_ATTRIBUTE);
+
+  if (!staticAttr && !boundAttr) {
+    return false;
```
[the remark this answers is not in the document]
Blackbaud-TrevorBurch: Fixed: an already-migrated element (static or bound `inset`) is now recognized before the legacy-attribute check, so it no longer increments `notConfiguredCount` or triggers the warning. Added tests for both the static and bound forms.

blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4674/storybook/)

Component Storybooks:

- [ag-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/ag-grid-storybook/)
- [angular-tree-component-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/angular-tree-component-storybook/)
- [data-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/data-grid-storybook/)
- [data-manager-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/data-manager-storybook/)
- [errors-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/errors-storybook/)
- [filter-bar-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/filter-bar-storybook/)
- [forms-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/forms-storybook/)
- [indicators-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/indicators-storybook/)
- [layout-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/layout-storybook/)
- [lists-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/lists-storybook/)
- [lookup-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/lookup-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/pages-storybook/)
- [phone-field-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/phone-field-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/text-editor-storybook/)
- [theme-storybook](https://blackbaud.github.io/skyux-pr-preview/4674/storybooks/theme-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4674/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4674/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4674/playground/)

apps/e2e/forms-storybook/src/app/checkbox/checkbox.component.html
```diff
@@ -192,7 +192,7 @@
     </sky-checkbox-group>
   </form>
 </sky-fluid-grid>
-<sky-fluid-grid id="icon-checkboxes">
+<sky-fluid-grid id="icon-checkboxes" inset="true">
```
Blackbaud-SteveBrush: All usage of `inset="true"` and `[inset]="true"` in this branch should be simplified to `inset`.

libs/components/code-examples/src/lib/modules/layout/fluid-grid/example.component.html
```diff
@@ -1,7 +1,7 @@
 <div class="highlight-columns">
   <sky-fluid-grid
     data-sky-id="fluid-grid"
-    [disableMargin]="disableMargin"
+    [inset]="inset"
```
Blackbaud-SteveBrush: When adding margins to their fluid grid, the majority case for our consumers will be to use the `inset` input unbound. It would be a nicer experience for our consumers to have two separate code examples for fluid grid: a basic grid without margins (default), and a second "Fluid grid with inset margins" code example that shows it simply:
```
<sky-fluid-grid inset>
   ...
```
The tests don't need to toggle the `inset` value; each code example spec can call the `hasMargin()` harness method, returning true or false depending on the example.

libs/components/layout/src/lib/modules/fluid-grid/fluid-grid.component.ts
```diff
@@ -14,12 +36,40 @@ import { SkyFluidGridGutterSizeType } from './types/fluid-grid-gutter-size-type'
...
+
+  public get disableMargin(): boolean | undefined {
+    return this.#_disableMargin;
+  }
+
+  /**
+   * Whether to add padding inside the fluid grid container so its content is
+   * inset from the container's outer left and right edges. When `false`,
+   * the fluid grid's content extends to the edges of its container.
+   * @default false
```
Blackbaud-SteveBrush, line 59: I know we talked about this briefly during Flex, but now that I've seen the implementation I think `disableMargin` should take precedence over `inset`, rather than the other way around:
1. `inset` is new API, so we should write it the "ideal" way and follow Angular best practices; which argues for a plain `booleanAttribute` transform. Letting `inset` win requires distinguishing "never set" from "explicitly false", which turns a binary input into a tri-state (true/false/undefined) and leaks that complexity into the input's public type. It also makes the `@default false` doc comment inaccurate, since the declared initial value is `undefined`.
2. Favoring `disableMargin` only matters for apps mid-migration. It's unlikely a consumer sets both inputs deliberately, so the primary reason to define a "winner" is safety during the transition. In that case, I would argue that deferring to the deprecated input is safer. An app that still sets `disableMargin` keeps its existing behavior, and the new default applies only where nothing was set.
3. It greatly simplifies the code. `insetTransform` and its 8-line explanatory comment go away, along with the `undefined` branching in `noMargin`.

libs/components/layout/src/lib/modules/fluid-grid/fluid-grid.component.ts
```diff
@@ -35,5 +85,18 @@ export class SkyFluidGridComponent {
     return this.#_gutterSize;
   }
 
+  /**
+   * Whether the fluid grid's outer left and right margin should be hidden.
+   * `inset`, when explicitly set, takes precedence over the deprecated
+   * `disableMargin` input. If only `disableMargin` is set, it's honored so
+   * existing behavior is preserved until it's removed.
+   */
+  protected get noMargin(): boolean {
```
Blackbaud-SteveBrush: Using a signal-backed `disableMargin` value would allow us to use an actual `computed`, here:

```typescript
readonly #disableMargin = signal<boolean | undefined>(undefined);

/**
 * Disables the outer left and right margin of the fluid grid container.
 * @deprecated Use `inset` instead. Note that the values are inverted:
 * setting `disableMargin` to `true` is equivalent to setting `inset` to `false`.
 */
@Input()
public set disableMargin(value: boolean | undefined) {
  this.#disableMargin.set(value);

  if (value !== undefined) {
    this.#logSvc.deprecated('SkyFluidGridComponent.disableMargin', { /* ... */ });
  }
}

public get disableMargin(): boolean | undefined {
  return this.#disableMargin();
}

/**
 * Whether to add padding inside the fluid grid container so its content is
 * inset from the container's outer left and right edges.
 * @default false
 */
public readonly inset = input(false, { transform: booleanAttribute });

protected readonly noMargin = computed(
  () => this.#disableMargin() ?? !this.inset(),
);
```

libs/components/packages/src/schematics/migrations/update-15/fluid-grid-inset/fluid-grid-inset.schematic.ts
```diff
@@ -0,0 +1,258 @@
...
+        tree.commitUpdate(recorder);
+      }
+    });
+  });
+
+  if (notConfiguredCount > 0) {
+    context.logger.warn(
+      `Found ${notConfiguredCount} <sky-fluid-grid> element(s) that do not set 'disableMargin'. ` +
+        "Starting in SKY UX 15, the fluid grid's outer left and right margin is hidden by default. " +
+        'Review these elements and add the \'inset\' input (e.g. [inset]="true") to any that should keep their margin.',
```
Blackbaud-SteveBrush: The example the log gives isn't ideal; what about something akin to `(e.g. <sky-fluid-grid inset>)`?

apps/e2e/forms-storybook/src/app/checkbox/checkbox.component.html
Blackbaud-TrevorBurch: Done

libs/components/code-examples/src/lib/modules/layout/fluid-grid/example.component.html
Blackbaud-TrevorBurch: Done

libs/components/layout/src/lib/modules/fluid-grid/fluid-grid.component.ts
Blackbaud-TrevorBurch: Done

libs/components/layout/src/lib/modules/fluid-grid/fluid-grid.component.ts
Blackbaud-TrevorBurch: Done

libs/components/packages/src/schematics/migrations/update-15/fluid-grid-inset/fluid-grid-inset.schematic.ts
Blackbaud-TrevorBurch: Done

apps/e2e/forms-storybook/src/app/checkbox/checkbox.component.html
Blackbaud-SteveBrush: I'm still seeing `inset="true"` in several places.

Blackbaud-TrevorBurch: Fixed.

Blackbaud-SteveBrush (approved):

PR #4694, opened 2026-08-25
feat(components/tiles)!: tile dashboard remains multi-column at the sm breakpoint

Blackbaud-ErikaMcVey:
## What changed

The tile dashboard collapsed to its single-column layout at both `xs` and `sm`. It now collapses only at `xs` (≤767px), so dashboards render multi-column from 768px up.

Three predicates gated on `sm`, all now keyed on `xs` alone:

- `tile-dashboard.service.ts` — the `#mode` computed. This is the one that actually moves tiles between the single- and multi-column layouts; drag/drop targets, keyboard navigation, and config persistence all read from it.
- `tile-dashboard.component.ts` — the layout class. Both branches ended up on the same condition, so the two `if`/`else` blocks collapse into one expression.
- `tile-dashboard.component.ts` — the "tile moved" assistive text, which reports column *N* of *M*. Without this it would announce "column 1 of 1" at `sm` while showing two columns.

**No SCSS changes were needed.** The styles key off `sky-tile-dashboard-xs`/`-gt-xs` and the `sky-host-responsive-container-sm-min` mixin (≥768px), both of which already treated `sm` as the larger-screen case, so the multi-column paddings and tile border radii were already correct there.

Also updated the doc comment on `SkyTileDashboardConfigLayout.singleColumn` from "small screens" to "extra-small screens".

## Breaking change

Consumers relying on a single-column layout between 768px and 991px will now see tiles laid out per `SkyTileDashboardConfigLayout.multiColumn` instead of `singleColumn`.

Note this applies to **any** column count, not just two. The multi-column layout renders whatever `multiColumn` contains at `flex-basis: 100/N%`, so a three-column dashboard now shows three columns at ~250px each in that range. Collapsing 3→2 isn't possible without a new layout tier in `SkyTileDashboardConfigLayout`, which currently models only `singleColumn` and `multiColumn`.

## Tests

- **Service spec** — two tests that used `sm` as a stand-in for single-column mode now use `xs`; the screen-size-change test asserts `sm` renders 3/1/0 tiles across the two multi columns and the single column.
- **Harness spec** — `isMultiColumn()` previously asserted only `xs` → false and `lg` → true, both unaffected by this change. Added `sm` → true.
- **Component spec** — two new tests for the moved-tile announcement at `sm` and `xs`. This branch had no assertions anywhere in the repo before; branch coverage was 100% only because drag/drop tests execute the line without checking its output. Required adding `provideSkyMediaQueryTesting()` to a spec that had no breakpoint control.
- **Visual** — the tiles spec now iterates `E2eVariations.RESPONSIVE_WIDTHS` (`[375, 800, 1000, 1280]`) with a real `cy.viewport()` per width, matching the pattern the `pages-storybook-e2e` tile dashboard specs already use. It previously covered only a default desktop viewport and `iphone-x`, so the entire changed range had no visual coverage.

The widths need a real `cy.viewport()` rather than a multi-width `widths` array: column mode is decided in JavaScript and the service physically relocates tiles between DOM containers, so Percy's re-render of a captured DOM at a different width would not switch layouts.

## Percy baselines

Two sources of expected diffs, both needing approval:

- The tiles snapshots are renamed and re-scoped (`desktop`/`mobile` → `375px`/`800px`/`1000px`/`1280px`), so this adds coverage at two breakpoints that were not previously captured.
- `blocks-tile-dashboard-page` and `tabs-tile-dashboard-page` in `pages-storybook-e2e` already iterate `RESPONSIVE_WIDTHS`, so their **800px baselines will diff.**

## Verification

`nx test tiles` 67/67 and `nx test tiles-testing` 19/19 pass, 100% statements and branches on both. Lint clean (warnings pre-existing). Cypress/Percy not run locally — needs a Percy token and a Storybook build.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated responsive tile dashboards to use a single-column layout only on extra-small screens.
  * Small-screen layouts now correctly use multiple columns.
  * Updated accessibility messaging for moved tiles based on the active layout.

* **Tests**
  * Added breakpoint coverage for small-screen and tablet layouts across themes.
  * Expanded responsive behavior and accessibility test coverage.
  * Added visual checks for configured responsive widths.

### DIFF
--- a/apps/e2e/tiles-storybook-e2e/src/e2e/tile-dashboard.component.cy.ts
+++ b/apps/e2e/tiles-storybook-e2e/src/e2e/tile-dashboard.component.cy.ts
@@ -3,41 +3,30 @@ import { E2eVariations } from '@skyux-sdk/e2e-schematics';
 describe('tiles-storybook', () => {
   E2eVariations.forEachTheme((theme) => {
     describe(`in ${theme} theme`, () => {
-      beforeEach(() =>
-        cy.visit(
-          `/iframe.html?globals=theme:${theme}&id=tiledashboardcomponent-tiledashboard--tile-dashboard`,
-        ),
-      );
+      E2eVariations.RESPONSIVE_WIDTHS.forEach((width) => {
+        describe(`at ${width}px`, () => {
+          beforeEach(() => {
+            cy.viewport(width, 960);
+            cy.visit(
+              `/iframe.html?globals=theme:${theme}&id=tiledashboardcomponent-tiledashboard--tile-dashboard`,
+            );
+          });
 
-      it('should render the component', () => {
-        cy.skyReady('app-tile-dashboard', ['.ready']).screenshot(
-          `tile-dashboard-${theme}-desktop`,
-          {
-            overwrite: true,
-          },
-        );
-        cy.get('app-tile-dashboard').percySnapshot(
-          `tile dashboard ${theme} desktop`,
-          {
-            widths: E2eVariations.DISPLAY_WIDTHS,
-          },
-        );
-      });
-
-      it('should render the component in mobile', () => {
-        cy.viewport('iphone-x', 'portrait');
-        cy.skyReady('app-tile-dashboard', ['.ready']).screenshot(
-          `tile-dashboard-${theme}-mobile`,
-          {
-            overwrite: true,
-          },
-        );
-        cy.get('app-tile-dashboard').percySnapshot(
-          `tile dashboard ${theme} mobile`,
-          {
-            widths: E2eVariations.MOBILE_WIDTHS,
-          },
-        );
+          it('should render the component', () => {
+            cy.skyReady('app-tile-dashboard', ['.ready']).screenshot(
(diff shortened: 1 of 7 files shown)


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4694/storybook/)

Component Storybooks:

- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/4694/storybooks/pages-storybook/)
- [tiles-storybook](https://blackbaud.github.io/skyux-pr-preview/4694/storybooks/tiles-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4694/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4694/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4694/playground/)

Blackbaud-TrevorBurch (approved):

PR #4753, opened 2026-09-04
fix(components/data-grid): avoid NG0950 when columns are created in a loop

johnhwhite:
[AB#4109298](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/4109298)

`headingText` is a required input on `SkyDataGridColumn`. Angular
populates signal content queries before it applies the queried
components' input bindings, so reading it off a column created inside
an `@for` block can throw NG0950 for a single change-detection pass
(angular/angular#59067). `#columnDefs` now catches only that specific
error per column and omits the column for that pass; the computed still
tracks the read as a dependency, so it self-heals once the binding
lands. Any other error is rethrown.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data grid stability when columns are created asynchronously.
  * Deferred column processing until column inputs are fully initialized.
  * Prevented uninitialized columns from causing temporary rendering errors.
  * Ensured columns added after the initial render are included once ready.
  * Improved handling of column inputs during initial rendering and subsequent updates.
  * Improved error handling during asynchronous column initialization.


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4753/storybook/)

Component Storybooks:

- [data-grid-storybook](https://blackbaud.github.io/skyux-pr-preview/4753/storybooks/data-grid-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4753/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4753/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4753/playground/)

libs/components/data-grid/src/lib/modules/data-grid/data-grid.ts
```diff
@@ -84,6 +85,7 @@ ModuleRegistry.registerModules([
   ColumnAutoSizeModule,
   EventApiModule,
   GridStateModule,
+  LocaleModule,
```
Blackbaud-SteveBrush, line 88: Is this needed for the error handling? If unrelated, let's create another PR so the reasoning appears in the changelog.

libs/components/data-grid/src/lib/modules/data-grid/data-grid.ts
```diff
@@ -97,6 +99,27 @@ function arraySorted(arr: string[]): string[] {
...
+    return false;
+  }
+
+  const code = (err as { code?: unknown }).code;
+
+  if (typeof code === 'number') {
+    return Math.abs(code) === 950;
+  }
+
+  return /^NG0950(?::\s|$)/.test(err.message);
```
Blackbaud-SteveBrush: Parsing the error code from the error message is a bit brittle. From [this Angular issue](https://github.com/angular/angular/issues/59067#issuecomment-2880931140), it looks like they recommend setting an "initialized" signal in the child component's `ngOnInit` hook. Could we implement something similar here?

**data-grid-column.ts**
```typescript
readonly #initialized = signal(false);

/**
 * Whether Angular has applied this column's input bindings. The grid's
 * content query updates before that happens, so the grid must not read
 * `headingText` until this flips.
 * @internal
 */
public readonly initialized = this.#initialized.asReadonly();

public ngOnInit(): void {
  this.#initialized.set(true);
}
```

**data-grid.ts**
```typescript
return columns
  .filter((col) => col.initialized())
  .map((col) => this.#createColDef(col));
```

libs/components/data-grid/src/lib/modules/data-grid/data-grid.ts
johnhwhite: Split out to #4772

libs/components/data-grid/src/lib/modules/data-grid/data-grid.ts
johnhwhite: Good call. Switched.

libs/components/data-grid/src/lib/modules/data-grid/data-grid-column.ts
```diff
@@ -146,11 +147,13 @@ export class SkyDataGridColumn {
...
   });
 
+  readonly #initialized = signal(false);
+  protected readonly initialized = this.#initialized.asReadonly();
   protected readonly templateChild = contentChild(TemplateRef);
 
   constructor() {
     const logger = inject(SkyLogService);
-    afterRenderEffect(() => {
+    effect(() => {
```
Blackbaud-SteveBrush: Just a follow-up question: what's the reasoning for replacing `afterRenderEffect` with `effect`? Could we leave it as-is, and simply move the `#initialized` assignment in a `ngOnInit` method? Seems to be what Angular recommends as a workaround: https://github.com/angular/angular/issues/59067#issuecomment-2880931140
```
public ngOnInit(): void {
  this.#initialized.set(true);
}
```

johnhwhite: I'd switched to `effect` because it fires sooner and unblocks columns from rendering.

I've switched to `ngOnInit()`. Tests remain the same.

Blackbaud-SteveBrush (approved):

blackbaud-sky-build-user:
Cherry-pick [failed](https://github.com/blackbaud/skyux/actions/runs/34380577072)! :x: Please resolve conflicts and create a new PR.

PR #4758, opened 2026-09-05
fix(sdk/skyux-stylelint): no-deprecated-sky-scss-variables fix should handle scss namespacing

johnhwhite:
[AB#4113536](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/4113536)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of deprecated SCSS variables in comments, quoted strings, namespaced references, and color-related functions.
  - Supports aliased color modules and accurately identifies color-channel usage.
  - Preserves source locations and keeps diagnostics aligned with automatic fixes.
  - Automatically fixes eligible variables while identifying unsupported occurrences for manual correction.
  - Handles overlapping variable names and SCSS comment syntax more reliably.

- **Documentation**
  - Clarified which deprecated usages can be auto-fixed and which require manual updates.


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4758/storybook/)

Component Storybooks:

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4758/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4758/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4758/playground/)

libs/sdk/skyux-stylelint/src/rules/no-deprecated-sky-scss-variables.ts
```diff
@@ -33,45 +182,98 @@ const ruleBase: RuleBase = (options) => {
     }
 
     root.walkDecls((decl) => {
-      const { value } = decl;
-      let match: RegExpExecArray | null;
+      // `decl.raws.value.raw` preserves the exact source text (including inline
+      // comments) that `decl.toString()` renders, whereas `decl.value` may have
+      // comments stripped. Matching/index computation must use the raw text so
+      // reported offsets line up with the actual rendered declaration.
+      const value = decl.raws.value?.raw ?? decl.value;
```
Blackbaud-SteveBrush: The warning positions are incorrect for inline `//` comment tags (due to the normalization adding a trailing `*/`). The `--fix` output also rewrites inline comments using block syntax (not ideal). 

Could we pass in `raws.value.scss` (the raw text) if it's defined so we can find the ranges for `//` comments accurately?

libs/sdk/skyux-stylelint/src/rules/no-deprecated-sky-scss-variables.ts
```diff
@@ -10,7 +10,43 @@ export const ruleName = withNamespace(ruleId);
...
+// up replacements in `deprecatedScssVarMap`; the full match (including the namespace)
+// is what gets replaced when fixing.
+const SCSS_VAR_PATTERN = /(?:[\w-]+\.)?(\$sky-[a-z0-9-]+)/g;
+
+// Sass/CSS functions whose arguments are color channels/components rather than whole
+// colors. Replacing a `$sky-*` variable with `var(--sky-theme-*)` inside one of these
+// produces syntactically valid SCSS that compiles to invalid CSS (e.g.
+// `rgba(var(--sky-theme-color-text-action), 0.25)`), so such occurrences are reported
+// but never auto-fixed.
+const COLOR_FUNCTIONS = new Set([
```
Blackbaud-SteveBrush: The `color.` prefixes are listed, but aliased `@use` imports are incorrectly "fixed" into invalid CSS. For example, `@use 'sass:color' as c;` would complete the "fix" migration and introduce broken code. 

There are [other `sass:color` functions](https://sass-lang.com/documentation/modules/color/#red) that aren't listed; do we care about those (e.g. `red`, `blue`, etc.)?

libs/sdk/skyux-stylelint/src/rules/no-deprecated-sky-scss-variables.ts
```diff
@@ -33,45 +182,98 @@ const ruleBase: RuleBase = (options) => {
...
+
+      for (const m of matches) {
+        const variable = m[1];
+        const start = m.index;
+        const end = start + m[0].length;
+        const replacement = deprecatedScssVarMap.get(variable);
+        const insideColorFunction = colorFunctionRanges.some(
+          ([rangeStart, rangeEnd]) => start >= rangeStart && start < rangeEnd,
+        );
+        const isFixable =
```
Blackbaud-SteveBrush, line 309: We should update the docs to account for cases that are no longer fixable.

libs/sdk/skyux-stylelint/src/rules/no-deprecated-sky-scss-variables.ts
```diff
@@ -33,45 +182,98 @@ const ruleBase: RuleBase = (options) => {
...
+      // `decl.raws.value.raw` preserves the exact source text (including inline
+      // comments) that `decl.toString()` renders, whereas `decl.value` may have
+      // comments stripped. Matching/index computation must use the raw text so
+      // reported offsets line up with the actual rendered declaration.
+      const value = decl.raws.value?.raw ?? decl.value;
+      // Offset of `value` within `decl.toString()`, used so each warning/fix targets
+      // only its matched variable instead of the whole declaration. Without this, all
+      // warnings on a declaration share the same (whole-node) position, which breaks
+      // down once some occurrences are fixed and others are not.
+      const valueStart = decl.prop.length + (decl.raws.between?.length ?? 0);
```
Blackbaud-SteveBrush: If `raws.between` is always populated by the parser, couldn't we set the default to `decl.raws.between ?? ': '` (the real default) and delete the ~40-line `should fall back to decl.prop.length when decl.raws.between is undefined` test? Feels like the edge case isn't worth the complexity on this one.

libs/sdk/skyux-stylelint/src/rules/no-deprecated-sky-scss-variables.ts
johnhwhite: Good point. Added better comment handling.

libs/sdk/skyux-stylelint/src/rules/no-deprecated-sky-scss-variables.ts
johnhwhite: Good catch. Broadened this to cover any alias and updated the function list to be complete.

libs/sdk/skyux-stylelint/src/rules/no-deprecated-sky-scss-variables.ts
johnhwhite: Good catch. Updated the docs.

libs/sdk/skyux-stylelint/src/rules/no-deprecated-sky-scss-variables.ts
johnhwhite: Fair enough. Dropped the handling for this edge case.

Blackbaud-SteveBrush (approved):

PR #4786, opened 2026-09-14
feat(components/tabs): vertical tabs support the same layout options as tabs

Blackbaud-TrevorBurch:
Resolves [AB#3648132](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/3648132)

`sky-vertical-tab` now offers the same layout options as `sky-tab` — `none` | `blocks` | `fit` | `list`, defaulting to `none` — so teams get the correct padding for their content based on the type of layout used inside the tab.

## What changed

- **New public type.** `SkyVerticalTabLayoutType` is exported from `@skyux/tabs` and registered in `documentation.json`, matching how `SkyTabLayoutType` is exposed.
- **New `layout` input** on `sky-vertical-tab`, documented to match `SkyTabComponent.layout` word for word.
- **Vertical tabset content spacing now comes from the shared `--sky-layout-host-*` tokens** instead of the component-specific `--sky-comp-tab-vertical-content-space-inset-*` tokens.
- **Removed the `--sky-override-*` wrappers** around the content margin, so the layout token is read directly. This matches how horizontal tabs are written (`.sky-tab` in `tab.component.scss` has no override wrapper) and is what makes the layout options take effect in the default theme as well as the modern theme.
- **New compat stylesheet entry** in the `update-15` `add-compat-stylesheets` schematic, restoring the previous spacing and scrolling for apps that do not set `layout`.

## Breaking change and the compat stylesheet

`.sky-layout-host-none` declares no content spacing, so a vertical tabset with no `layout` input now renders with **zero margin** and **`overflow-y: visible`** where it previously had padding and scrolling. The `skyux15-compat.css` stylesheet generated by `ng update` restores both via three custom properties:

| Property | Restores |
| --- | --- |
| `--sky-compat-vertical-tabset-content-spacing-xs` | content margin at the xs breakpoint |
| `--sky-compat-vertical-tabset-content-spacing-sm` | content margin at sm and above |
| `--sky-compat-vertical-tabset-content-overflow-y` | the former unconditional `overflow-y: auto` |

Because the overrides were removed rather than relocated, **default-theme apps are affected too**, not just modern-theme apps. The generated CSS emits a `:root` block with the default-theme values plus a `.sky-theme-modern` block using the real design tokens — the two genuinely differ, since `right` and `bottom` are non-zero in the modern theme.

Every compat default equals the new behavior, so apps that never install the stylesheet are unaffected by the compat mechanism itself.

## Not done in this PR

The story's fourth bullet — adding a compat stylesheet section to the migration guide — is **out of scope**. No migration guide lives in this repo; the generated stylesheet's inline instructions are the only migration guidance produced here.

## Testing

- `nx test tabs` — 194/194 on Karma, covering the layout class on the host and content wrapper for all four values, the `none` default when the input is omitted or `undefined`, updates on active-tab change and on input change, `SkyLayoutHostService.setHostLayoutForChild`, behavior with no layout host, and `activeTabLayout` emission and completion.
- `nx test packages` — 249/249 on Jest, including new cases asserting both compat blocks are emitted for `@skyux/tabs` and that nothing is emitted when `@skyux/tabs` is not installed.
- `nx run-many -t lint build -p tabs packages` — clean.

> [!IMPORTANT]
> **`vertical-tabs-fit-layout.component.cy.ts` has never been executed.** Cypress could not be run locally in this environment. Its selectors were verified by reading the component templates and it passes lint and typecheck, but nothing proves the spec passes, and the check that the Percy snapshot is actually sensitive to `layout="fit"` was never run. **Please confirm both in CI.**
>
> Percy snapshots for the existing vertical tabs stories are also unvalidated locally and are expected to shift in both themes.

## Follow-ups (not blocking)

- The global `.sky-layout-host-fit` also leaks `display: flex; flex-direction: column` onto the content wrapper. Benign for `fit`, but it is the same class of leak as the viewport-inset issue fixed here.
- The `.sky-theme-modern` compat block uses bare `var(--sky-comp-*)` with no fallback, consistent with existing convention in this file.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added configurable vertical-tab content layouts: none, blocks, fit, and list.
  * Vertical tabsets now adjust content spacing, scrolling, sizing, and overflow based on the selected layout.
  * Added fit-layout examples for horizontal and vertical tabs, including themed demonstrations.
  * Added compatibility styling to help preserve legacy vertical-tabset behavior during migration.

* **Documentation**
  * Documented vertical-tab layout options and configuration guidance.


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4786/storybook/)

Component Storybooks:

- [action-bars-storybook](https://blackbaud.github.io/skyux-pr-preview/4786/storybooks/action-bars-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/4786/storybooks/pages-storybook/)
- [split-view-storybook](https://blackbaud.github.io/skyux-pr-preview/4786/storybooks/split-view-storybook/)
- [tabs-storybook](https://blackbaud.github.io/skyux-pr-preview/4786/storybooks/tabs-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/4786/storybooks/text-editor-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4786/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4786/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4786/playground/)

apps/e2e/tabs-storybook-e2e/src/e2e/vertical-tabs-fit-layout.component.cy.ts
```diff
@@ -0,0 +1,36 @@
...
+
+        // The active tab's content is moved out of its `sky-vertical-tab`
+        // element and into the tabset's content pane, so the assertion must be
+        // rooted at the content pane rather than at the tab.
+        cy.get('.sky-vertical-tabset-content .placeholder-danger')
+          .should('exist')
+          .should('be.visible');
+
+        cy.window().screenshot(`vertical-tabs-fit-layout-${theme}`);
+        cy.window().percySnapshot(`vertical-tabs-fit-layout-${theme}`, {
```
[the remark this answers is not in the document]
Blackbaud-TrevorBurch: Skipping this one. The finding is a style preference, not a defect, and acting on it would break the consistency you asked for.

Verification

cy.window().screenshot(…) / cy.window().percySnapshot(…) is the established convention in this repo, not something my test introduced:

It's in 11 e2e spec files across colorpicker, forms, modals, pages, and tabs.
The horizontal tabs-fit-layout test — the one you asked me to mirror — uses the identical pattern.
That pattern exists on main today in the same file under its pre-rename name (fit-layout.component.cy.ts:18-19), so it has been running in CI and producing Percy snapshots for some time. It works.

The finding gives no failure mode, and I can't construct one: @percy/cypress registers percySnapshot as a parent command, so the piped subject is ignored either way, and .screenshot() off a non-element subject falls back to the viewport capture — which is what these snapshots want.

Why changing it would be a net negative

Applying it only to my file would make the vertical fit test diverge from the horizontal fit test sitting beside it, which directly contradicts the "match what tabs-fit-layout is already doing" requirement. Applying it across all 11 files would be a repo-wide refactor well outside this PR, and would move Percy baselines for five unrelated component areas.

If the team wants cy.screenshot() as the house style, that's a worthwhile standalone cleanup — with all 11 files changed together and the baselines reviewed in one pass. I'd flag it as such rather than start it here.

apps/e2e/tabs-storybook/src/app/tabs-fit-layout/tabs-fit-layout.component.scss
```diff
@@ -0,0 +1,29 @@
+:host {
+  display: block;
+}
+
+.placeholder {
+  // Fills the tab with an in-flow element rather than `position: absolute`
```
Blackbaud-ErikaMcVey: this comment is...rather confusing? was absolute positioning how it was tested before or something?

libs/components/packages/src/schematics/migrations/update-15/add-compat-stylesheets/add-compat-stylesheet.schematic.ts
```diff
@@ -43,6 +43,45 @@ Pointer events are no longer disabled on elements with the "sky-btn-disabled" cl
...
+      name: '@skyux/tabs',
+      components: [
+        {
+          name: 'vertical-tabset',
+          styles: [
+            {
+              css: `
+/* The --sky-comp-* tokens used below are declared on the element that carries
+   the ".sky-theme-modern" class, and var() is substituted on the element that
+   declares the custom property. A ":root" block therefore cannot read them, so
```
Blackbaud-ErikaMcVey: do we actually want to include all of this detail in all consumers' compat files? Maybe just add a `/* default theme */` comment above the `:root` if you think that isn't clear?

libs/components/tabs/src/lib/modules/vertical-tabset/vertical-tabset.component.scss
```diff
@@ -44,19 +40,60 @@
...
+  margin: var(--sky-layout-host-content-spacing, 0);
+  flex-grow: var(--sky-layout-host-content-flex-grow, 0);
+  overflow: var(--sky-layout-host-content-overflow, visible);
+  position: var(--sky-layout-host-content-position, static);
+
+  &.sky-layout-host {
+    &-fit {
+      // The global `.sky-layout-host-fit` rule positions a layout host against
+      // the viewport with `position: absolute` plus `--sky-viewport-*` inset
+      // offsets. This element is not a layout host: it only borrows the layout
```
Blackbaud-ErikaMcVey: I'm not sure I'm totally following this...how is it not a layout host? why would we put the layout host class on it if it is not in fact a layout host? I understand we don't want it shifted down 50px for the omnibar, since it isn't directly on the page, but I don't see how that makes it not a layout host?

It looks like the `tab` component handles that by setting `.sky-layout-host-fit` position to `static`. Would that work here or are there differences I'm not seeing?

libs/components/tabs/src/lib/modules/vertical-tabset/vertical-tabset.component.scss
```diff
@@ -105,12 +142,10 @@
...
-      --sky-override-vertical-tabset-content-margin-sm-lg,
-      var(--sky-comp-tab-vertical-content-space-inset-sm-top)
-        var(--sky-comp-tab-vertical-content-space-inset-sm-right)
-        var(--sky-comp-tab-vertical-content-space-inset-sm-bottom)
-        var(--sky-comp-tab-vertical-content-space-inset-sm-left)
-    );
+    margin: var(--sky-layout-host-content-spacing, 0);
+
+    &.sky-layout-host-none {
+      margin: var(--sky-compat-vertical-tabset-content-spacing-sm, 0);
```
Blackbaud-ErikaMcVey: should this fallback to `--sky-layout-host-content-spacing` instead 0? For the other breakpoint as well? I realize it is set to 0 for fit, but I think it communicates it more clearly which will make cleanup easier later

apps/e2e/tabs-storybook-e2e/src/e2e/tabs-fit-layout.component.cy.ts
```diff
@@ -5,18 +5,18 @@ describe('tabs fit layout', () => {
     describe(`in ${theme} theme`, () => {
       beforeEach(() =>
         cy.visit(
-          `/iframe.html?globals=theme:${theme}&id=fit-layoutcomponent--fit-layout`,
+          `/iframe.html?globals=theme:${theme}&id=tabs-fit-layoutcomponent--tabs-fit-layout`,
```
Blackbaud-ErikaMcVey: so these actually _aren't_ fit edge to edge in default anymore, and the vertical tabs fit layout aren't either. Looking at the default overrides it looks like that is actually the intended behavior, and that is why that CSS for the visual test changed? So it is actually an accurate test? Just wanted to confirm, since it threw me off that default didn't actually "fit"

libs/components/tabs/src/lib/modules/vertical-tabset/vertical-tab.component.ts
```diff
@@ -90,6 +93,28 @@ export class SkyVerticalTabComponent implements OnInit, OnDestroy {
...
   public errorIndicator: boolean | undefined = false;
 
+  /**
+   * The tab layout that applies spacing to the tab container element. Use the layout
+   * that corresponds with the top-level component type used within the tab, or use `fit` to
+   * constrain the tab contents to the available viewport.
+   * Use `none` for custom content that does not adhere to predefined spacing or constraints.
+   * @default "none"
+   */
+  @Input()
```
Blackbaud-ErikaMcVey: should we add any code examples with the layout input set?

PR #4794, opened 2026-09-16
fix(components/colorpicker): reset an originally-empty field to empty and pristine

johnhwhite:
[AB#4109461](https://dev.azure.com/blackbaud/f565481a-7bc9-4083-95d5-4f953da6d499/_workitems/edit/4109461)

Previously, resetting a colorpicker that started out empty set the
value to white (#FFFFFF) and marked the control dirty, instead of
restoring it to its true empty state.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

## Summary by CodeRabbit

- **Bug Fixes**
  - Resetting a color picker with no initial value now properly clears the input and bound form control.
  - Resetting an originally empty color picker restores its pristine state without emitting a color-change event.
  - Resetting a color picker with an initial value now restores that value consistently.
  - Clearing a value now resets the picker display, styling, native input, and form state together.
  - Clearing a bound form value now synchronizes the picker’s displayed and applied color.
  - Required-field validation remains visible after resetting an empty color picker.


blackbaud-sky-build-user:
[Storybook preview](https://blackbaud.github.io/skyux-pr-preview/4794/storybook/)

Component Storybooks:

- [colorpicker-storybook](https://blackbaud.github.io/skyux-pr-preview/4794/storybooks/colorpicker-storybook/)
- [pages-storybook](https://blackbaud.github.io/skyux-pr-preview/4794/storybooks/pages-storybook/)
- [text-editor-storybook](https://blackbaud.github.io/skyux-pr-preview/4794/storybooks/text-editor-storybook/)

Apps:

- [code-examples-playground](https://blackbaud.github.io/skyux-pr-preview/4794/code-examples-playground/)
- [integration](https://blackbaud.github.io/skyux-pr-preview/4794/integration/)
- [playground](https://blackbaud.github.io/skyux-pr-preview/4794/playground/)

libs/components/colorpicker/src/lib/modules/colorpicker/colorpicker-input.directive.ts
```diff
@@ -355,6 +340,53 @@ export class SkyColorpickerInputDirective
     this.#renderer.setProperty(element, 'value', output);
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  #applyValue(value: any, captureInitial: boolean): void {
+    if (
+      this.skyColorpickerInput &&
+      value &&
+      value !== this.skyColorpickerInput.lastAppliedColor
```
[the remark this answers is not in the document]
johnhwhite: @coderabbitai skipping this because:

- **Tracking initial-value capture with a separate boolean.** NgModel always issues a synchronous `writeValue(null)` before its real value lands, so "control started empty" is indistinguishable from normal NgModel initialization. Implementing it broke both existing `should populate correct information if model is given` specs (`initialColor` became `undefined` instead of `#00f`).
- **Clearing `#hsva` to `undefined` as the finding literally specified.** Every slider setter (`hue`, `red`, `alphaAxis`, `saturationAndLightness`) guards on `#hsva`, so a `undefined` `#hsva` would leave the reopened picker's sliders dead. Resetting to the default color fixes the same defect and keeps the picker usable.

libs/components/colorpicker/src/lib/modules/colorpicker/colorpicker.component.ts
```diff
@@ -646,12 +646,17 @@ export class SkyColorpickerComponent
...
+          this.updatePickerValues(this.initialColor);
+          this.backgroundColorForDisplay = this.initialColor;
+          this.selectedColorChanged.emit(this.selectedColor);
+          // TODO: This code assumed non-null pre-strict mode. Reevaluate in the future?
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          this.selectedColorApplied.emit({ color: this.selectedColor! });
+        } else {
+          this.#colorpickerInputSvc.clearValue.next();
+        }
+        this.ngControl?.control?.markAsPristine();
```
Blackbaud-TrevorBurch, line 659: I understand wanting to reset the value. But marking this back as pristine seems odd to me. An example of why.

1. A color field is initially undefined - but a required field.
2. The user sets a value
3. The user "resets" the field

In this situation - the required error would not show because we marked the field as pristine even though the user interacted with it.

When using the reset button - that is a change to the value - even if it is returning back to what it was initially. It feels like we should maintain the button marking the form as dirty.

johnhwhite: Good catch to double-check, but I don't think this causes the failure mode you're describing. `sky-form-errors` gates the `required` error on `touched` alone (not `dirty`), and `onTriggerButtonClick` marks the control touched whenever the picker is opened — reset doesn't clear `touched`, only `pristine`. So for a required field: set a value → reset → the required error is still shown, because `touched` is retained even though `pristine` is now `true`.

This also matches `AbstractControl.reset()` semantics in Angular core, which is more aggressive than what we're doing here — it clears both `dirty` and `touched`. We're intentionally keeping `touched` so validation feedback isn't lost.

I added a spec for exactly this scenario (required, originally-empty field → apply → reset) asserting the required error is still rendered: see the two new "should still show the required error after resetting an originally-empty, required colorpicker" tests.

Blackbaud-TrevorBurch (approved):

blackbaud-sky-build-user:
Cherry-pick successful! :cherries: :tada: See https://github.com/blackbaud/skyux/pull/4797

